PR #20, opened 2018-02-13
Update the dagger coffee example to use kt_jvm_library

cgruber:
Update the dagger coffee example to use kt_jvm_library for the code and java_binary for the final executable jar and wrapper.

### DIFF
--- a/examples/dagger/BUILD
+++ b/examples/dagger/BUILD
@@ -13,7 +13,7 @@
 # limitations under the License.
 package(default_visibility = ["//visibility:private"])
 
-load("//kotlin:kotlin.bzl", "kt_jvm_binary")
+load("//kotlin:kotlin.bzl", "kt_jvm_library")
 
 java_plugin(
     name = "dagger_plugin",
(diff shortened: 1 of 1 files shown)


cgruber:
Small note - because the dagger example is the super-simple entry-level example, we probably want it to use the default practice one would use, which I think would be a java_binary.   I'll add some alternatives like mixed-source coffee app, coffee app tests, etc. in subsequent commits.

hsyed:
I agree on the simplicity aspect, but

1) This examples is also compiled and validated by the tests. 
2) These example should also be the go to for validating the intellij plugin.

We are showing the API pattern here where the `java_library` Is collecting a bunch of jars and plugins for reuse and we are testing that `plugin_export` are being interpreted correctly by the kt_ rules.

I recommend updating [skydoc](https://github.com/bazelbuild/rules_kotlin/blob/57497866f3157cb8fd1a44943ca26846fdcada1e/kotlin/kotlin.bzl#L55-L89) instead

hsyed:
nvm my comments I need to get better at reading git diffs.

cgruber:
Hmm.  I probably could have updated the skydocs also. I'll roll a separate one.

PR #24, opened 2018-02-15
initial toolchain implementation, enables coroutines and also reorgan…

hsyed:
…ize doc wireup


hsyed:
@cgruber I merged this in as I want to implement a few more things, but please give the initial design a review here.

PR #40, opened 2018-03-02
[wip] dep analysis for strict deps and unused deps

hsyed:
So this PR has a POC of strict dep checking working. The sequence of actions run within the builder is as follows:

1. Generate jar (java + kotlin + annotation processing for both Java and Kotlin)
2. Poor mans jdeps file generation via the jdk jdeps tool and text parsing. 
3. Compare the jdeps analysis (same format --i.e., `Deps.Dependencies`)  against the `--direct_dependencies` and `--indirect_dependencies` passed via the command line.
4. If any indirect dependencies are used generate a simple error containing `buildozer` output and delete the generated jar. 

The dependency analysis / dependency selection is not in line with how the native rules currently do It https://github.com/bazelbuild/bazel/issues/4584. The analysis is attempting to select deps further away from a jars `File.owner` -- at least one hop but at the moment it could be much closer to the target. So for example a `//third_party:guice` label could be chosen for both `javax-inject` and `guice`. Perhaps the heuristic should prefer in-project labels exporting only one jar if available. 

A compiler plugin for Kotlin makes more sense so that we get more error context -- one short term work around is to run a compile phase after detecting indirect dependency usage with a restricted classpath and concatenate those errors with the buildozer suggestions. This doesn't cover the mixed mode scenario, currently java code is compiled by javac -- and writing a compiler plugin for javac makes no sense. Could the JavaBuilder  be leveraged here without going back out to skylark  ?


hsyed:
@iirina I'm wondering if you could give this POC a review or suggest someone else. It would be useful if I could reuse any bits in Bazel core -- I am essentially rewriting the java builder with a Kotlin compile phase.

kotlin/toolchains.bzl
```diff
@@ -111,6 +119,11 @@ def define_kt_toolchain(name, language_version=None, api_version=None, jvm_targe
         api_version = api_version,
         jvm_target = jvm_target,
         coroutines = coroutines,
+        jvm_strict_deps = select({
```
hsyed: @brendandouglas FYI this is how I am currently going to relay the strict dep flag to the builder. Is this the right place for it ?  My initial thoughts were to go through a common hidden attribute on the rules but https://github.com/bazelbuild/bazel/issues/287 prevents this. I can make it hidden here ? 

p.s., This branch is quite chaotic atm so don't bother reading any of the other code.

PR #43, opened 2018-03-04
Add support to kotlin 1.2.30 to the toolchains.

cgruber:


hsyed:
Cool, the [default version](https://github.com/cgruber/rules_kotlin/blob/1a97a8d42ac6481af36b977f80b0a7909782a75d/kotlin/kotlin_compiler_repositories.bzl#L150) should also be bumped, the tests will run against the new compiler then.

cgruber:
Bumped the default version. Let's see how the tests pan out.

PR #47, opened 2018-03-05
Add instructions on configuring ijwb to properly know about kotlin

cgruber:
Add instructions on configuring ijwb to properly know about kotlin, and mention the experimental plugins in case the current release isn't yet available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,24 @@ kt_jvm_library(
 )
 ```
 
+## IDE
+
+Intellij support via the intellij with blaze plugin needs to be fairly recent to support the
+features of the kotlin rules.  If you are having issues, try installing an [experimental 
+plugin release][ijwb] which support a more robust set of integrations with bazel and kotlin.
+
+Also, enable kotlin support in your .bazelproject file (found in the menu as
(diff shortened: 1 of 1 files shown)


hsyed:
@brendandouglas what is the ETA on a release of the plugin with Kotlin support. This PR creates a link to a build of the plugin that we use internally.

brendandouglas:
These kind of instructions really belong on the long-neglected docs site (ij.bazel.build).

There's another release scheduled for upload this Friday, which will be at least as recent as what's checked in here.

hsyed:
Ok so if it's going in on Friday there's no point linking to my releases.

This is all that is the documentation that is needed Brendan. Could you update them ? 

```
additional_languages:
  kotlin
```

The Kotlin section for configuring the language version is pointless with my next bit of work (picking up kotlin config from the rules) so no need to document it -- it picks up the default from the Kotlin plugin currently.

cgruber:
I looked at ij.bazel.build and couldn't see where this should really fit there. :/  

But I"ll close this PR, as it does seem that part of it is irrelevant, and the other part should be part of the docs for the tools in question.

PR #61, opened 2018-05-07
test refactor and replace moshi with protobuf, start model proto file

hsyed:
* remove moshi and use proto json support instead.
* add guava, gson, protobuf, protobuf-java, protobuf-java-util to the repositories using common repository conventions.
* refactor tests a bit.
* replace plugin descriptor with proto variant.

The new proto jars are still inlined, and even though there is a dependency on com_google_protobuf protoc should not need to be built


hsyed:
@hchauvin I've opened up and started adding external deps. The deps you need should be added following this pattern. If you use `java_import_external` a jar alias should be bound and `omit_`  properties should be should be threaded through to macros in `kotlin_compile_repositories`. 

I am going to add Guice to the builder so it can be made more testable, @hchauvin what do you think ?  Dagger is not going to work for the builder --- easily.

hsyed:
the commit in here is bundled with #62

PR #70, opened 2018-05-23
Expose toolchain dict directly in common attributes

brendandouglas:
Renames attributes for consistency with the internal rules, and removes toolchain ide info in favor of the platform_common.ToolchainInfo dict.

The renames are entirely optional -- happy to revert them. I'm assuming they're not yet referenced anywhere, and will only be used by the IntelliJ plugin in the short term.

If this change is OK, the next step would be exposing the standard library targets directly in the toolchain dict.

### DIFF
--- a/kotlin/BUILD
+++ b/kotlin/BUILD
@@ -14,7 +14,6 @@
 package(default_visibility = ["//visibility:public"])
 
 load("//kotlin:toolchains.bzl", "define_kt_toolchain")
-load("//kotlin/internal:bootstrap.bzl", "kt_toolchain_ide_info")
 
 toolchain_type(
     name = "kt_toolchain_type",
@@ -22,7 +21,5 @@ toolchain_type(
 )
 
 define_kt_toolchain(name = "default_toolchain")
-
-kt_toolchain_ide_info(name="kt_toolchain_ide_info")
 
 exports_files(["toolchains.bzl", "kotlin.bzl"], visibility=["//docs:__subpackages__"])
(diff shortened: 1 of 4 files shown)


hsyed:
CL ? What does that mean ?

Yes lets get a generic toolchain in. ‘kt_toolchain’ ?

hsyed:
CL ? What does that mean ?

Yes lets get a generic toolchain in. ‘kt_toolchain’ ?

brendandouglas:
Hah, sorry, I meant a separate pull request, not CL.

hsyed:
Fyi there is a proto model being used in the builder https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/builder/proto/kotlin_model.proto. It is stitched together in the builder though from command line flags.. I created to have a sensible immutable model in the builder and to be able to launch integration tests directly against the builder.

It might contains a lot of the details ito enable fastpath testing for kotlin in intellij ?

hsyed:
@brendandouglas Could we standardise on this proto message for toolchains ? It's in the [Proto Model](https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/builder/proto/kotlin_model.proto#L35-L52). We can add the stdlibs to it and have the rules write out a proto text info file instead of Json ? 

```
message KotlinToolchainInfo {
    message Common {
        // mandatory
        string language_version = 1;
        //mandatory
        string api_version = 2;
        // oneof "enable", "warn" or "error"
        string coroutines = 3;
    }

    message Jvm {
        // oneof "1.6", or "1.8" 
        string jvm_target = 4; // (need to set this 1)
    }

    Common common = 1;
    Jvm jvm = 2;
}
```

brendandouglas:
I've just tried to move closer to the internal rules here -- I'm not sure what the longer term plan is for the kotlin toolchain internally (I'm not the best person to discuss this with though).

cgruber:
Obsolete.

PR #75, opened 2018-06-09
Manually call the Kotlin compiler during bootstrap

dflemstr:
The usual `kotlinc` wrapper script fails to correctly detect `KOTLIN_HOME` unless we run with `--spawn_strategy=standalone`. With this change, it becomes possible to run the build even with a remote Bazel executor, or within a sandbox if desired.


hsyed:
@dflemstr thanks for tidying up the bootstrap file !

PR #82, opened 2018-06-27
Update builder to support building from a JDK9 host_javabase

hsyed:
@cushon I'll need some help on getting started. 

Initial point -- I couldn't compile HEAD with 0.15.0 so I used 0.14.1. 

The two bootstrap macros manage to build builder Kotlin libraries -- they have a genrule depending on `@local_jdk//:jdk`. 
 
```
./scripts/bazel_from_source build --verbose_failures //kotlin/builder:compiler_lib
./scripts/bazel_from_source build --verbose_failures //kotlin/builder:builder_lib
```

Moving on to the `java_binary` wrapping the builder libraries above. I get failures relating to missing JDK jars -- each run gives a different missing jar.

```
➜  rules_kotlin git:(fix_jdk9) ./scripts/bazel_from_source build --verbose_failures //kotlin/builder:builder     
INFO: Analysed target //kotlin/builder:builder (0 packages loaded).
INFO: Found 1 target...
ERROR: missing input file '@local_jdk//:jre/lib/rt.jar'
ERROR: /Volumes/work/src/github.com/hsyed/rules_kotlin/kotlin/builder/BUILD:70:1: //kotlin/builder:builder: missing input file '@local_jdk//:jre/lib/rt.jar'
 ```
I'm on OS X, I don't have a JAVA_HOME set and my shell choses JDK9 by default. But As far as I understand it the @local_jdk should be the one bundled in the bazel I just built ?


hsyed:
@cushon I'm looking at it now. Updated the script.

cushon:
For cases where you're using the JDK as a tool that runs during the build (e.g. when invoking `kotlinc`), `cfg="host"` is correct. The latest version of Bazel defaults to using a JDK 9 host_javabase, so it may be revealing cases where running `kotlinc` on JDK 9 breaks some of its assumptions, but those are bugs that should be fixed in kotlinc. Using `cfg="target"` might temporarily work around those issues, but is not a principled fix.

Depending on the JDK with `cfg="target"` is usually only necessary for rules that don't use the JDK during the build, and instead of produce an executable that runs the JDK (e.g. the implementation of `kotlin_binary` and `java_binary`).

hsyed:
Ok so the Kotlin builder process is JDK9 and the kapt susbsystem checks the `java.version` System property and does something "JDK9" specific. The `tools.jar` that is loaded into the classloader for Kotlin compilation is from JDK 8.

The builder picks out the tools jar directly from "external/local_jdk/lib/tools.jar"

The Kotlin compiler has a "--jdk-home" flag but I don't think it honors this. I will create an issue on youtrack

@cushon in the meantime I'd like to try some thing. I see an `embedded_jdk` directory in the runfiles of the builder -- but no tools.jar. Do you know how I can load the classes from the `com.sun.tools.*` into the kotlin compiler classloader in jdk9 ?

hsyed:
@cushon 

I've loaded the jdk.compiler module into the classloader via the `java.home` system property. I'm also loading `tools.jar` via the system property if it's jdk8 as well. As far as I can tell this is the correct solution. 

The dagger annotation processor which is used in the example is now complaining about missing javax.annotation.Generated. Adding the ""java.xml.ws.annotation" does not fix this.

hsyed:
@cushon all done i'll merge and close.

PR #94, opened 2018-07-10
Remove jar tool usage and add jar normalisation.

hsyed:
This PR  removes usage of the jar tool and ensure the jar created by the builder and the resource-merged jar are  normalised.

### Review Notes

I've gotten the primary jar generated by the builder and the resource-merged jar normalised in the sense that they hash the same locally.  But when used in our primary mono repo the action count building(not running)  a test_suite target isn't going down as I was expecting it to after changing a core file (101 actions executed 72).  

I'm working on porting to the new `JavaInfo` if correcting things with the existing skylark logic is too much work I can just finish that port.

my assumptions are:

* If the jars are properly normalised then a change in dep a in `a -> b -> c`  should not cause c to require a rebuild (the action should be dropped from the list). 
* When working with a file Bazel only looks at a hash of a files content for invalidation (not the file timestamps).
* The jar does not need to be propagated via `JavaInfo#compile_jar` for the invalidation logic to work.

What else could I be doing wrong ? I'm sure it's something obvious. I have added RFC's below.

### DIFF
--- a/kotlin/builder/src/io/bazel/kotlin/builder/utils/jars/JarHelper.kt
+++ b/kotlin/builder/src/io/bazel/kotlin/builder/utils/jars/JarHelper.kt
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Copied from bazel core and there is some code in other branches which will use the some of the unused elements. Fix
+// this later on.
 @file:Suppress("unused","MemberVisibilityCanBePrivate")
 package io.bazel.kotlin.builder.utils.jars
 
@@ -35,7 +37,6 @@ import java.util.zip.CRC32
  * Adjusting the timestamp for .class files is necessary since otherwise javac will recompile java
  * files if both the java file and its .class file are present.
  */
-
 open class JarHelper internal constructor (
     // The path to the Jar we want to create
     protected val jarPath: Path,
@@ -200,17 +201,17 @@ open class JarHelper internal constructor (
      * correspondence.
      *
      * @param path the path used to retrieve the timestamp in case normalize is disabled.
-     * @param bytes if this is null then the entry is a directory.
+     * @param data if this is empty array then the entry is a directory.
      */
-    protected fun JarOutputStream.copyEntry(name: String, path: Path? = null, bytes: ByteArray = EMPTY_BYTEARRAY) {
+    protected fun JarOutputStream.copyEntry(name: String, path: Path? = null, data: ByteArray = EMPTY_BYTEARRAY) {
         val outEntry = JarEntry(name)
         outEntry.time = when {
             normalize -> normalizedTimestamp(name)
             else -> Files.getLastModifiedTime(checkNotNull(path)).toMillis()
         }
-        outEntry.size = bytes.size.toLong()
+        outEntry.size = data.size.toLong()
 
-        if (bytes.isEmpty()) {
+        if (data.isEmpty()) {
             outEntry.method = JarEntry.STORED
             outEntry.crc = 0
             putNextEntry(outEntry)
@@ -218,13 +219,13 @@ open class JarHelper internal constructor (
             outEntry.method = storageMethod
             if (storageMethod == JarEntry.STORED) {
                 val crc = CRC32()
-                crc.update(bytes)
+                crc.update(data)
                 outEntry.crc = crc.value
                 putNextEntry(outEntry)
-                write(bytes)
+                write(data)
             } else {
                 putNextEntry(outEntry)
-                write(bytes)
+                write(data)
             }
         }
         closeEntry()
@@ -245,8 +246,6 @@ open class JarHelper internal constructor (
         // They must all be kept in sync.
         val TARGET_LABEL = Attributes.Name("Target-Label")
         val INJECTING_RULE_KIND = Attributes.Name("Injecting-Rule-Kind")
-
-        const val DOS_EPOCH_IN_JAVA_TIME = 315561600000L
 
         // ZIP timestamps have a resolution of 2 seconds.
         // see http://www.info-zip.org/FAQ.html#limits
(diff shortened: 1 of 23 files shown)


kotlin/builder/src/io/bazel/kotlin/builder/utils/jars/JarHelper.kt
```diff
@@ -0,0 +1,255 @@
...
+        closeEntry()
+    }
+
+    companion object {
+        const val MANIFEST_DIR = "META-INF/"
+        const val MANIFEST_NAME = JarFile.MANIFEST_NAME
+        const val SERVICES_DIR = "META-INF/services/"
+        internal val EMPTY_BYTEARRAY = ByteArray(0)
+        /** Normalize timestamps.  */
+        val DEFAULT_TIMESTAMP = LocalDateTime.of(1980, 1, 1, 0, 0, 0)
```
hsyed: set to 1980 to align with the normalisation format of singlejar.

kotlin/internal/compile.bzl
```diff
@@ -221,7 +221,7 @@ def _compile_action(ctx, rule_kind, module_name, friend_paths=depset(), src_jars
 
     # setup the merge action if needed.
     if len(output_merge_list) > 0:
-        utils.actions.fold_jars(ctx, output_jar, output_merge_list)
+        utils.actions.fold_jars(ctx, rule_kind, output_jar, output_merge_list)
```
hsyed: Is this available in the ctx somewhere ? if not should I create an issue ?

kotlin/internal/utils.bzl
```diff
@@ -187,11 +187,17 @@ def _maybe_make_srcsjar_action(ctx, srcs):
 
 
 # PACKAGE JARS #################################################################################################################################################
-def _fold_jars_action(ctx, output_jar, input_jars):
-    args=["--output", output_jar.path]
+def _fold_jars_action(ctx, rule_kind, output_jar, input_jars):
+    args=[
+        "--normalize",
+        "--compression",
+        "--deploy_manifest_lines",
```
hsyed: manually stamp this into the jar -- perhaps we need a better `stamp_jar` sidekick that merges and stamps ? .

tests/integrationtests/jvm/JvmBasicFunctionalTests.kt
```diff
@@ -41,6 +41,27 @@ class JvmBasicFunctionalTests : AssertionTestCase("tests/integrationtests/jvm/ba
         }
     }
 
+    @Test
+    fun testJarNormalization() {
+        jarTestCase(
+            name = "test_module_name_lib.jar",
+            description = "Builder jars should be normalized with the same timestamps as singlejar and including stamp data"
+        ) {
+            validateFileSha256("513d14b29eb1b95b97bf7d34e2126a716c7d1012b259b5021c16b99ca82feeb5")
```
hsyed: This should be a valid way of validating the jar right ? across operating systems ?

kotlin/internal/utils.bzl
hsyed: Another question about me manually stamping here and in the builder (using the JavaCreator). Will the "Target-Label" be picked up if I do it this way ? Skylark hasn't been made aware of it explicitly -- i.e., will Skylark read the manifest.

kotlin/builder/src/io/bazel/kotlin/builder/utils/jars/JarHelper.kt
cushon: There's nothing special about 2010-1-1. 1980-1-1 is fine in theory, but in practice using the earliest representable timestamp creates some opportunities for bugs (e.g. bazelbuild/bazel#4614).

kotlin/builder/src/io/bazel/kotlin/builder/utils/jars/JarHelper.kt
```diff
@@ -0,0 +1,255 @@
...
+        // These attributes are used by JavaBuilder, Turbine, and ijar.
+        // They must all be kept in sync.
+        val TARGET_LABEL = Attributes.Name("Target-Label")
+        val INJECTING_RULE_KIND = Attributes.Name("Injecting-Rule-Kind")
+
+        const val DOS_EPOCH_IN_JAVA_TIME = 315561600000L
+
+        // ZIP timestamps have a resolution of 2 seconds.
+        // see http://www.info-zip.org/FAQ.html#limits
+        const val MINIMUM_TIMESTAMP_INCREMENT = 2000L
```
cushon: Is this unused?

kotlin/internal/utils.bzl
cushon: > Is there a way inheriting the manifest from the primary jar with singlejar ?

Not currently. For the Java rules singlejar doesn't process the interface jars that are used at compile-time, which is the only time we need `Target-Label`.

> Skylark hasn't been made aware of it explicitly -- i.e., is Bazel meant to actively read the manifest for the stamp data ?.

Tools like JavaBuilder read the manifest data when they are executed. Bazel itself doesn't read it during analysis or anything.

kotlin/builder/src/io/bazel/kotlin/builder/utils/jars/JarHelper.kt
hsyed: it is used, I've seen the increments by examining the zip files.. 

p.s., I copy and "kotlin-converted " `JarCreator` and `JarHelper` from bazel core. I haven't changed them other then adding one method to JarHelper.

cushon: Sorry, I meant to comment on `DOS_EPOCH_IN_JAVA_TIME`. I think it's unused here and in the original `JarHelper`.

hsyed: aah ok , yes it's unused I'll remove it.

hsyed:
@cushon It seems I got it right, I did some manual testing with 
```
kt_jvm_library(name = "a", srcs =["a_b_c/A.kt"], resource_jars = [":resourcejar"])
kt_jvm_library(name = "b", srcs =["a_b_c/B.kt"], deps=[":a"])
kt_jvm_library(name = "c", srcs =["a_b_c/C.kt"], deps=[":b"])
```
and C doesn't rebuilder anymore.

Now I have to figure out whe the CI servers are hashing things differently.

hsyed:
Thanks for the review !

Issue #89, opened 2018-07-04
Design document for a `kt_library` rule.


hsyed:
I think before tackling JS we need a `kt_library` and have it fully working with the JVM rules. 

Considering just the JVM aspect. A `kt_library` at the very least is a list of src files and it should be "somehow" be used in `kt_jvm_library`. Extending this to multiplatform a `kt_library` could be multiplatform. In which case to create a `kt_jvm_library` it would need srcs containing the JVM `actual` instances. 

```py
kt_library(
  name = "foo_api",
  srcs = glob(["api/*.kt"])
)

kt_jvm_library(
  name = "foo_jvm",
  expected_by = [ ":api" ] # should this be list or single ? 
  srcs = [ "jvm/*.kt" ]
)
```

Some questions: 

1. Is a `kt_library` always multiplatform -- can it be used just to group sources together --i.e., should it also be allowed in the srcs list of `kt_jvm_library`, `kt_js_library` ?.
2. If a multiplatform library isn't always multiplatform do we need to explicetly mark it multiplatform or is this knowledge only required when compiling it --i.e., by virtue of being featured in the `expected_by` attribute ?
3. What do the providers look like ?

thewoolleyman:
Hi @hsyed ,

Just checking in.  I made an attempt to just hack this to work by creating a `kt_js_library` rule, and just define a parallel execution/provider path for js files.

However, with my limited knowledge of Bazel I got stuck on what to do with the `KotlinBuilder` class and toolchain, which is currently JVM-specific.

Would love to see this working, but I don't know enough about Bazel yet to proceed any further without significant guidance/direction.

So, putting this on the back burner for now and going back to Gradle to build my kotlin/js 😔

Thanks,
-- Chad

thewoolleyman:
@hsyed 

One main thing that I realized is that in the JS ecosystem, there has to be some core concepts of managing dependencies via npm/yarn, and output bundles.

Other rules libraries implement this in various ways, for example https://github.com/bazelbuild/rules_nodejs or https://github.com/bazelbuild/rules_typescript.

Some of these also implement other features we may or may not want, e.g. building for dev mode, and running local servers (e.g. express).

My main question is: for the core stuff, dependencies and bundles, would we need to reimplement a lot of it, or would it be possible to "compose" or reuse with some of these other sets of rules like the nodejs ones.

[...]

PR #114, opened 2018-08-12
basic js support

hsyed:
@alexeagle 

Notes for node:

* `kt_js_library` is currently wrapped in a  macro and it locks the `module_name` and `module_root` to be the same as the label name. If the `module_name` and `module_root` could be picked up from a provider (could be a legacy provider -- i.e., `kt_js` on the provider struct) then a better approach can be used.
* Entry point requires the workspace name. I would have created a `kt_js_binary` macro to wrap the compile and execute in one step. But the workspace name is not available in a macro. Is there some way around this ? 

@brendandouglas

Notes for IJ plugin:

Some experimentation is needed. Simply attaching the JS jars (stdlib and otherwise) to the library table as a "Kotlin JS artifact" should do the trick. Please have a look and suggest any provider changes that you think might be required. I'd be happy to do the actual work if I could get my work upstreamed in reasonable time.


examples/node/express/Routes.kt
```diff
@@ -0,0 +1,16 @@
+package express
+
+import express.auth.isAuthenticated
+
+//import express.auth.isAuthenticated
```
hsyed: fix

kotlin/builder/proto/kotlin_model.proto
```diff
@@ -147,3 +147,25 @@ message JvmCompilationTask {
...
+    message Outputs {
+        // The primary output
+        string js = 1;
+        string jar = 3;
+        string srcjar = 4;
+    }
+
+    message Inputs {
+        repeated string libraries = 1;
+        // Partitioned from the builder flags, expanding the source_jars.
```
hsyed: fix comment

kotlin/internal/js/js.bzl
```diff
@@ -0,0 +1,105 @@
...
+            mandatory = True,
+        ),
+        "srcjar": attr.label(
+            allow_files = ["-sources.jar"],
+            mandatory = False,
+            single_file = True,
+        ),
+        "runtime_deps": attr.label_list(
+            default = [],
+            allow_files = [".jar"],
```
hsyed: add _KtJsInfo provider requirement

alexeagle:
Why do you require users to specify whether the kotlin files should run as js as part of the build rule? What about a common library that works in both runtimes? It would be better to just make kt_library do the right thing.

hsyed:
@alexeagle thanks for having a look ! 

[This issue](https://github.com/bazelbuild/rules_kotlin/issues/89) covers the design of a `kt_library` rule. 

From what I understand you are implying a `kt_library` that is capable of being a `common`  library as well as a `js` and `jvm` library -- is this correct ?  There are some nuances to this  (see [kotlin multiplatform docs](https://kotlinlang.org/docs/reference/multiplatform.html)).

A simple package for a domain model which has:

a) No external dependencies.
b) No platform specific logic.

Would compile to JS and JVM with no problems.  If either cases above are not true it gets complicated. 

I am currently thinking of creating a  `kt_common_library` rule first which would declare the common files and then `kt_js_library` and `kt_jvm_library` could take this as an `actual_for` attr and compile a platform specific variant. 

This is how proto compilation works eventually once everything is working we could revisit a smarter `kt_library`.

hsyed:
@alexeagle what about the points I raised above, `module_name` and `module_root`. Could you add support to the node rules so that they are up from providers as well as the attrs ?

alexeagle:
Sorry, I was on my phone over the weekend so that comment was cryptic.

`proto_library` is an interesting example. There are two things to note there:
1) some package distributes a `proto_library`, but has no idea about my language `foo`. I can still have a `foo_proto_library` rule in my package which does the codegen for that `proto_library` and its dependency tree
2) `proto_library` was forced to do this to avoid our earlier architecture, where any dependency on a proto_library caused all the different codegen systems to be built, e.g. C++ users were waiting for the Java protoc target to build.

#1 is clearly undesirable, I'd rather just have a `java_library` depend on a `proto_library` without having to manually add a `java_proto_library` rule in between. It's only needed because `java_library` doesn't know how to walk its transitive deps looking for `proto_library` that requires codegen. It probably shouldn't. 

But in your case, we can totally make `nodejs_binary` -> `kotlin_library` dependency work correctly. The `kotlin_library` should expose a provider that offers to provide JS outputs from the Kotlin provider. These compiler actions would only run if some downstream rule in the graph requests that provider. Then any runtime-specific rule at the root of the tree (`nodejs_binary`, `rollup_bundle`, etc) already knows to walk the graph collecting this provider. This is how it works in TypeScript - there are different outputs from the TypeScript compiler and they're selected based on where they are used, since there is browser runtime and nodejs runtime.
Note, we should add a more general JS provider for this, the current setup is too specific to the TypeScript case.

I only know a bit about Kotlin, but I suspect that #2 is not an issue for you - you probably don't need to compile a bunch of extra tooling in order to request JS outputs from the kotlin compiler?

I had missed your point about needing a different API for `module_name`/`module_root` would you mind filing that feature request on rules_nodejs and we can get that in this week?

hsyed:
Your analysis on `kt_library` makes total sense. I think we need some real examples before making a start on a unified `library`. Such a rule would need to take platform specific attributes for each of the platforms. For example platform specific srcs --i.e., `actual_jvm`, `actual_js` (there is also native). The platform specific sources provide platform implementations. For example if some common library declares `expected fun debug_log(msg: String)`. On the JVM it might be `actual fun debug_log(msg: String) { System.err.println(msg) }` and in JS it would be `actual fun debug_log(msg: String) { console.log(msg) }`. 

The JVM aspect would probably need to support all of the attributes that `kt_jvm_library` supports. The native attributes will be as numerous as the JVM ones. Such a rule could quickly turn into a godzilla rule with many attributes. 

I am hoping the team internal to Google implementing the Kotlin rules would take lead on the design of the shape of the rules. 

> I had missed your point about needing a different API for module_name/module_root would you mind filing that feature request on rules_nodejs and we can get that in this week?

I've created the issue.

> I only know a bit about Kotlin, but I suspect that #2 is not an issue for you - you probably don't need to compile a bunch of extra tooling in order to request JS outputs from the kotlin compiler?

That is correct the Kotlin compiler distribution from jetbrains contains everything. The same persistent worker compiles jvm and js.

PR #117, opened 2018-08-13
Basic Android support

hsyed:
@jin Everything seems to be working. There is a single thin macro in `android.bzl`. Could you verify the macros are correct. 

Some notes / RFCs:

* This approach should work for mix java / kotlin compiles -- so a project can be gradually converted to Kotlin !
* AS 3.0.1 is working fine with the Kotlin. No new rules were added (might need to add a internal `kt_android_library` rule later on to pick up the android  toolchainWDYT ?). 
* I didn't look into the kotlin aidl compiler plugin. I don't think there is a need WDYT ? . 
* The example is not very nice.  It would be really useful to get a comprehensive idiomatic example. @jin could the android team contribute such a sample ?

### DIFF
--- a/kotlin/internal/jvm/BUILD
+++ b/kotlin/internal/jvm/BUILD
@@ -11,3 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+java_import(
+    name = "android_sdk",
+    jars = ["@bazel_tools//tools/android:android_jar"],
+    neverlink = 1,
+    visibility = ["//visibility:public"],
+)
--- a/kotlin/internal/jvm/android.bzl
+++ b/kotlin/internal/jvm/android.bzl
@@ -37,6 +37,9 @@ def _kt_android_artifact(name, srcs = [], deps = [], plugins = [], **kwargs):
     return [base_name, kt_name]
 
 def kt_android_library(name, exports = [], visibility = None, **kwargs):
+    """Creates a Android sandwich library. `srcs`, `deps`, `plugins` are routed to `kt_jvm_library` the other android
+    related attributes are handled by the native `android_library` rule.
+    """
     native.android_library(
         name = name,
         exports = exports + _kt_android_artifact(name, **kwargs),
--- a/kotlin/internal/jvm/android.bzl
+++ b/kotlin/internal/jvm/android.bzl
@@ -1,0 +1,44 @@
+# Copyright 2018 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


kotlin/internal/jvm/BUILD
```diff
@@ -11,3 +11,10 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
+java_import(
+    name = "android_sdk",
+    neverlink = 1,
+    jars = ["@bazel_tools//tools/android:android_jar"],
```
hsyed: Surely there must be a better place to import this from ?

kotlin/internal/jvm/android.bzl
```diff
@@ -0,0 +1,43 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+load(
+    ":jvm.bzl",
+    _kt_jvm_library = "kt_jvm_library",
+)
+
+def _kt_android_artifact(name, srcs = [], deps = [], **kwargs):
```
hsyed: fix spelling artefact !

kotlin/internal/jvm/BUILD
hsyed: @jin is this correct ?

kotlin/internal/jvm/android.bzl
hsyed: pull out the `plugins` attribute and route it to `kt_jvm_library`.

examples/android/android.bazelproject
```diff
@@ -0,0 +1,11 @@
+directories:
+  .
+
+targets:
+  //:android
```
jin: this seems to be `//examples/android:android`

kotlin/internal/jvm/android.bzl
```diff
@@ -0,0 +1,44 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+load(
+    ":jvm.bzl",
+    _kt_jvm_library = "kt_jvm_library",
+)
+
+def _kt_android_artefact(name, srcs = [], deps = [], plugins = [], **kwargs):
```
jin: Bazel uses `artifact` everywhere, let's standardize on that?

kotlin/internal/jvm/BUILD
jin: Yeah, that's the correct approach.

jin: (for now). This might change sometime in the future where it's moved into `@androidsdk`.

examples/android/android.bazelproject
hsyed: The bazelproject is relative to the example workspace.

kotlin/internal/jvm/android.bzl
hsyed: updated in new PR.

hsyed:
Removed example from this PR. don't want to lose all of this context.

hsyed:
@jin The current example is in branch `temp_android_example` for reference. 

Will you be able to port the codelab example ?

Issue #27, opened 2018-02-22
Fail to build on Windows


alex-chung:
When I run `bazel build //examples/dagger:coffee_app` on Windows 10, it fails with the following error message: `Error: Could not find or load main class org.jetbrains.kotlin.preloading.Preloader`

I am running Bazel 0.10.1.

hsyed:
This is likely down to path seperator issues. I’ll try my best to fix them but I do not have a windows machine to test with.

PR welcome :)

hsyed:
It will be a while till I get around to this, waiting on the Bazel CI system to support windows. Also #40 prints pretty colours to the console -- need to find a suitable lib or something in Bazel core that handles console printing.

buchgr:
@hsyed what's the blocker here? We do support Windows on Buildkite.

hsyed:
@buchgr i’ve made a bit of progress. I am blocked on this [issue](https://github.com/bazelbuild/bazel/issues/5523).

IljaKroonen:
@hsyed I'm trying to have a look at this but I have trouble connecting the dots. Is the https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template_windows.txt file supposed to somehow "fix" the classpath when running a command? Is it not possible to check if we are on a windows or on a linux machine to change the separator in the piece of code you linked in the other thread? I suppose that would be a hack? https://github.com/bazelbuild/rules_kotlin/blob/740b962803b0a128832c4970587cc8d26dc80546/kotlin/internal/utils.bzl#L187-L187

hsyed:
@IljaKroonen You are very brave !

If you are planning to work on the code I should push my current refactors. I've completely reorganized the skylark code.

There is an issue open in [bazel core](https://github.com/bazelbuild/bazel/issues/5523)  -- I couldn't figure out how to debug this particular script, it's like magic. I had to park the task and delete windows from virtualbox (80 gigs!).

It does more than just path replacements, it compensates for the lack of symlinks in windows. 

I'll try to help you as best as I can but I recommend chasing up the issue in bazel core.

hsyed:
@IljaKroonen ok I've just pushed my current refactor changes.

IljaKroonen:
@hsyed I'm new to bazel and I'm having trouble navigating all of this. Until now I haven't really observed the java_stub_template.txt file not working - it seems it is used many times successfuly.

[...]

PR #131, opened 2018-08-22
Fix usage of Java tools in rules_kotlin

nlopezgi:
Use of external/local_jdk is being deprecated, its also not compatible with remote execution.
This PR replaces all incorrect uses of java tools in the rules_kotlin rules.
This PR also adds CI config for testing rules_kotlin with remote execution via buildkite


kotlin/internal/jvm/jvm.bzl
```diff
@@ -125,10 +125,7 @@ _implicit_deps = {
         allow_files = True,
     ),
     "_java": attr.label(
-        executable = True,
-        cfg = "host",
```
hsyed: On windows singlejar is using the Java version. We probably need both the host and javabase version declared here ? (see last comment in #27).

WORKSPACE
```diff
@@ -27,6 +27,16 @@ http_jar(
     url = "https://github.com/hsyed/bazel-deps/releases/download/v0.1.0/parseproject_deploy.jar",
 )
 
+http_archive(
```
hsyed: Is this to get for getting the tests running in docker ? rbe_ubuntu1604 , is that a "docker host" ?

nlopezgi: Yes, this is to get the config info necessary to run on rbe_ubuntu1604

kotlin/internal/jvm/jvm.bzl
hsyed: Also for tests, Should tests be running under host or local jdk by default ?

nlopezgi:
A few test targets are still failing as (it seems) some actions are still running locally, will debug and fix soon.

kotlin/internal/jvm/jvm.bzl
nlopezgi: I'm going to try to get this to work with host jdk (embedded jdk if default), as you are right, it might be better to use host jdk for use of these host tools like singlejar

hsyed: > if it was a binary, but on windows its a jar. To fix that you probably need to condition how you invoke singlejar depending on the env.

If a level of indirection to chose between singlejar native and singlejar jvm. `_java` would become a tool dependency of the rule that merges jars. So I think both `_singlejar` and `_java` would need to be in cfg=`host`. 

[Some guidelines are documented here] (https://docs.bazel.build/versions/master/skylark/rules.html#configurations)

> My impression is that tests should run under target configuration (not host) but I might be mistaken (I'm far from expert in configurations).

Same.

Since java isn't used for any build actions at the moment I think the change should be fine. I'll add a  `_host_java` implicit if needed for windows.

hsyed: > current_host_java_runtime does not seem to have the files defined as deps

which  files were missing ?

nlopezgi: > which files were missing ?
 printing out ctx.files._java returned an empty set for embedded_jdk while local_jdk returns a bunch of files. Still not sure why.

> Since java isn't used for any build actions at the moment I think the change should be fine.

Thanks, will focus on getting the failing targets to build correctly on remote execution and let you know once this PR is ready for review again.

nlopezgi:
Ok so root caused why some actions were running locally: strategy=xxx=worker is not compatible with remote execution (i.e., exec strategies are either / or, so worker is by definition a different strategy to remote). 
After adding config to makes sure tests run remotely on rbe-ubuntu config I noticed there are still two tests that are failing, which seem to not be remote execution compatible at the moment: KotlinJvmDaggerExampleTest and KotlinJvmKaptAssertionTest.

These tests are failing with errors of this type:
1 error
src/test/data/jvm/kapt/kotlin/TestKtValue.kt:23:30: error: unresolved reference: AutoValue_TestKtValue
    fun builder(): Builder = AutoValue_TestKtValue.Builder()

To me this looks like a missing implicit dependency (see https://docs.bazel.build/versions/master/remote-execution-rules.html#managing-implicit-dependencies). Since the scope of this PR is just to get java tool usage in shape and get most test setup on remote execution I think this should be sufficient for now.

@hsyed ptal.

WORKSPACE
nlopezgi: forgot to clarify, rbe_ubuntu1604 is an experimental hosted remote execution service that we are opening up for rule owners to try (will be sending out email to Bazel community in a few weeks, but I'm testing out the process with rules_kotlin atm - Bazel head as well as some other rules are running with this config as well). Once this is setup you might notice faster CI results from rbe_ubuntu1604 than from other configs. Having these rules be compatible with remote execution will open the door for remote execution bazel customers to run projects that depend on your rules with the experimental hosted remote build service.

kotlin/internal/jvm/jvm.bzl
hsyed: I think I know what the problem is. When actions are not run as a worker there is a sandbox initialization issue. 

A resolve_command has to be sequenced into the action [see this](https://github.com/bazelbuild/rules_kotlin/blob/20313211b3caacc3cd7c376af1c3645edf3168e0/kotlin/internal/jvm/compile.bzl#L211). Could you try this ? I suspect you'll find that you'll be able leave the attrs `allow_files`, `default`.`executable` and `host` as is .

nlopezgi:
@hsyed please see my comment about the advice you gave. Let me know if you want me to try anything else or if this is good to be merged.

hsyed:
changes upstreamed in #134

nlopezgi:
Thanks!

PR #148, opened 2018-09-18
refactor the testbench mocks so they are  more composeable

hsyed:
Also Adds sources and output jar normalization tests.


hsyed:
@buchgr FYI: Apple worker seems to be stuck

PR #156, opened 2018-10-24
Make sure testonly is plumbed through the android rules.

cgruber:
Android rules break if they depend on testonly, even if you set testonly=1, as the testonly setting isn't passed on to the kt_android_library's native android_library wrapper, nor the underlying pass-through target (that suffixes with "_base").  This adds testonly to both.


cgruber:
Also could you look at this one, @kevin1e100?

kevin1e100:
I think you want to default to None but not sure

PR #151, opened 2018-10-10
Add exports to kt_jvm_import

wcurrie:
Adds an `exports` attribute to `kt_jvm_import`. For kotlin support in bazel-deps: https://github.com/johnynek/bazel-deps/pull/212

Example https://github.com/wcurrie/bazel-kt-test

Having `exports` allows `bazel-deps` to generate a `kt_jvm_import` rule for each external dependency (jar), avoiding ijar generation, but maintaining transitive dependencies via `exports`.

As a reminder, trying to run the kotlin compiler with ijars instead of full jars will fail on methods marked as inline.

I see `kt_jvm_import` was/is meant to be temporary (https://github.com/bazelbuild/rules_kotlin/issues/4#issuecomment-402570312).

Maybe ijar will support kotlin in someday soon (https://github.com/bazelbuild/bazel/commit/c1f6455d38399853ff9f34e3eb5e8ca26e590290):
> We have a Kotlin compiler plugin that will annotate inlined methods with @KeepForCompile and ijar can use this to include code that is a part of the library's compiletime interface into the stripped jar.

I couldn't find any description of the plugin or `@KeepForCompile`.

Is there any harm in merging this, even if `kt_jvm_import` will later be deleted?

PR #161, opened 2018-11-18
Add example of use with johnynek/bazel-deps

wcurrie:
Adds an example project for https://github.com/bazelbuild/rules_kotlin/pull/151. 

It's perhaps too complicated to keep in `rules_kotlin`. Happy to close if it's not worth maintaining.


cgruber:
This commit seems unrelated to the kotlin rules, but more related to managing third-party deps, which is not kotlin-specific.

PR #162, opened 2018-11-18
 Pass transitive deps to kotlinc

wcurrie:
When the java compiler is run by `java_binary` and `java_library` the compile classpath includes `exports` of `deps`: transitive dependencies. This currently doesn't happen with `kt_jvm_binary` or `kt_jvm_library`. This change makes the kotlin rules consistent (I believe/hope).

IIUC using `exports` in this way is not ideal. It's preferable to declare all compile `deps` explicitly. In practice having transitive dependencies is a valuable stepping stone for a large scale maven -> bazel migration.

Ideally this behaviour would be complemented by a `strict_java_deps` like flag? That's a much bigger job.


wcurrie:
This change may be meaningless without support for `exports` in `kt_jvm_library` https://github.com/bazelbuild/rules_kotlin/pull/151.

cgruber:
Why did this get closed? I think it's still a valid issue. I'm feeling this pain in our usage today.

wcurrie:
Happy to leave it open. Only closed it coz nothing seemed to be happening. I'm not working with this stuff in $DAY_JOB.

PR #163, opened 2018-11-20
docs: remove 'kotlin_release_version' as attribute of kotlin_repositories

gertvdijk:
Support appears to be removed with commit fe593c56551b304451, but docs weren't updated as such.


cgruber:
Docs are generated now, upon release.

PR #167, opened 2018-11-24
Set default file modification date in jars to 2010-1-1

mjduijn:
This change sets the default modification data files in jars created with rules_kotlin to 2010-1-1.

I've patched JarHelper.kt with the latest version of the [bazelbuild/bazel Jarhelper](https://github.com/bazelbuild/bazel/blob/master/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/jarhelper/JarHelper.java)

Reasoning: I noticed testJarNormalization failed to pass because the modification date for files in `test_embed_resources.jar` was set to 2010-1-1 instead of 1980-1-1.
In commit [bcefd9833cb5620fef8a27c37c2808a66b57c7e6 in bazelbuild/bazel](https://github.com/bazelbuild/bazel/commit/bcefd9833cb5620fef8a27c37c2808a66b57c7e6#diff-81c20aa091487ff7400c193d10ea25f2) the default date when normalizing Jars is set to 2010-1-1.

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/utils/jars/JarHelper.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/utils/jars/JarHelper.kt
@@ -80,7 +80,7 @@ open class JarHelper internal constructor (
     }
 
     /**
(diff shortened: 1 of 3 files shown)


cgruber:
No longer relevant to the code.

PR #170, opened 2018-12-21
Allowing no-source kt-lib

menny:
This may be useful when wanting to combine several libs together. Similar functionality
is available in java_library.
Ensuring that if  were specified we fail (should be converted to runtime_deps).


menny:
Being a bit selfish here, but this missing feature blocks my Maven deps generator from producing good rules for Kotlin: https://github.com/menny/bazel-mvn-deps#kotlin

cgruber:
For what it's worht, menny, your maven stuff shouldn't need kotlin in any way at all.  If you have jvm artifacts (jars) just use java_library/jar_import etc.  

This is a totally needed feature, however, especially since this also breaks if there are java sources but no .kt sources, and kotlin should be able to process that case.

cgruber:
This is being fixed in the current head, as a part of a broader set of changes to builder handling, so I'm going to close this.

PR #177, opened 2019-03-05
Update bazel toolchains pin and add toolchain config target

nlopezgi:


nlopezgi:
fyi @meteorcloudy (as he last approved changes to bazel-toolchains pin in this repo)

nlopezgi:
Thanks for the super quick response!

PR #178, opened 2019-03-11
Fix mixed UTF-8 Java and Kotlin builds

rcari:
When building .java files using the native Java rules, the environment
variable LC_LANG=en_US.UTF-8 is explicitely set for the actual build
action, see: https://git.io/fhhc7

When building a mixed Kotlin + Java module on Linux, the .java sources
are not handled by the native Java rules, and as a result are treated as
encoded using US-ASCII due to sandboxing which strips the environment
from all locale information, causing the compiler to default to the
'C' locale.

This patch adds this environment variable when building Kotlin + Java
modules and has no side effects on Kotlin-only modules as UTF-8 is the
only valid encoding for .kt source files.


rcari:
I signed it!

10-10-11:
Thanks for the fix!

PR #187, opened 2019-05-20
Added support for bazel generated embeded jar resources.

igorgatis:
Before this change, generated files could not be located when placed as `resources` attribute. Current code uses `.short_path` which works fine for regular files. Howver, generated files are placed under `bazel-out/k8-fastbuild/genfiles` which is not in the source tree. This change adds `bazel-out/k8-fastbuild/genfiles` to the list of `_CONVENTIONAL_RESOURCE_PATHS` thus this prefix is removed before adding the resource into the jar file.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -52,6 +52,8 @@ def _fold_jars_action(ctx, rule_kind, output_jar, input_jars):
 _CONVENTIONAL_RESOURCE_PATHS = [
     "src/main/resources",
     "src/test/resources",
+    "bazel-out/k8-fastbuild/genfiles",
+    "bazel-out/k8-fastbuild/bin",
 ]
 
 def _adjust_resources_path_by_strip_prefix(path, resource_strip_prefix):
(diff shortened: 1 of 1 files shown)


igorgatis:
@hsyed could you pls take a look?

igorgatis:
@jin Does this PR make sense to you?

cgruber:
As I understand it, this repo won't be taking more patches until the internal google rules are pushed out. I'm bringing my repo http://github.com/cgruber/rules_kotlin more up to date, if you want to contribute there, and we can chat with the google team once they're in a position to take patches on what parts of that fork to bring in.

kotlin/internal/jvm/compile.bzl
```diff
@@ -52,6 +52,8 @@ def _fold_jars_action(ctx, rule_kind, output_jar, input_jars):
 _CONVENTIONAL_RESOURCE_PATHS = [
     "src/main/resources",
     "src/test/resources",
+    "bazel-out/k8-fastbuild/genfiles",
```
jin: This seems odd. We shouldn't depend on bazel-* paths, especially if they have configuration parameters in them, like k8.

Issue #179, opened 2019-03-19
kt_jvm_test on Windows: MANIFEST did not contain path mapping for external/com_github_jetbrains_kotlin/lib/kotlin-compiler.jar


jin:
```
ERROR: C:/users/jingwen/code/rules_jvm_external/tests/unit/kotlin/BUILD:3:1: Compiling Kotlin to JVM //tests/unit/kotlin:inline_function_test { kt: 1, java: 0, srcjars: 0 } failed (Exit 1)
Exception in thread "main" java.lang.IllegalStateException: runfile manifest C:/users/jingwen/_bazel_jingwen/mejy7zte/execroot/rules_jvm_external/bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/builder.exe.runfiles/MANIFEST did not contain path mapping for external/com_github_jetbrains_kotlin/lib/kotlin-compiler.jar
        at io.bazel.kotlin.builder.utils.BazelRunFiles.resolveVerified(BazelUtils.kt:72)
        at io.bazel.kotlin.builder.toolchain.KotlinToolchain$Companion.createToolchain(KotlinToolchain.kt:68)
        at io.bazel.kotlin.builder.toolchain.KotlinToolchain.createToolchain(KotlinToolchain.kt)
        at io.bazel.kotlin.builder.KotlinBuilderMain.main(KotlinBuilderMain.java:28)
INFO: Elapsed time: 4.161s, Critical Path: 0.80s
INFO: 0 processes.
```

Repro on Windows:

```
$ git clone https://github.com/bazelbuild/rules_jvm_external
$ cd rules_jvm_external
$ git checkout kotlin
$ bazel test //tests/unit/kotlin/...
```

I managed to get further by adding the missing kotlin-compiler jar to the data runfiles of //src/main/kotlin:builder:

```python
# The builder artifact.
java_binary(
    name = "builder",
    data = [
        ":compiler_lib.jar",
        "@com_github_jetbrains_kotlin//:lib/kotlin-compiler.jar", # <----
    ],
    main_class = "io.bazel.kotlin.builder.KotlinBuilderMain",
    visibility = ["//visibility:public"],
    runtime_deps = ["//src/main/kotlin/io/bazel/kotlin/builder"],
)
```

Then the `kt_jvm_test` gets a bit further, and fails at it not being able to find `java.exe` from the local JRE.

```
==================== Test output for //tests/unit/kotlin:inline_function_test:
[...]

PR #193, opened 2019-07-02
BugFix: Windows: MANIFEST did not contain path mapping for kotlin-compiler.jar

mig35:
This fixes: https://github.com/bazelbuild/rules_kotlin/issues/179


mig35:
I signed it!

jin:
Thanks for looking into this!

PR #198, opened 2019-08-10
Add in a script to test these rules against a set of bazel versions. 

cgruber:
Also, update the readme with the resulting compatibility output. 

The script itself puts out a matrix.md file, which could be used in docs, but for now I'm using it just as a template for updating the readme.


cgruber:
This was in error - it was supposed to be a pr on my fork only.

PR #207, opened 2019-10-10
Update readme URL

keith:


jin:
thanks!

jdemeulenaere:
I think this commit should be revert as https://github.com/bazelbuild/rules_kotlin/archive/legacy-modded-1_0_0-01.zip leads to a 404.

PR #209, opened 2019-10-11
Add a facility for inferring the test class from the test name

### LINKED ISSUE #150: kt_jvm_test should infer the test_class if unspecified.
If the rule is named `FooTest` and the test file is `FooTest.java` or `FooTest.kt`, and the `test_class` is unspecified, the test rule should just work out that the test_class is `my.package.FooTest`.

> Note: it may be wrong if tests are specified differently in the `.kt` file, but that's totally fine, insofar as one can still override the `test_class`.

cgruber:
If no test_class is specified, do a best-effort inference of the test_class based on the test name, similar to the java_test rule.  It might fail, but if so, it'll fail either by not setting test_class, or by setting it to an object which doesn't exist, which would be reported as such, so the error is pretty clear. 

Fixes #150

### DIFF
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -19,7 +19,6 @@ load(
     "//kotlin/internal:defs.bzl",
     _KtJvmInfo = "KtJvmInfo",
 )
-
 load(
     "//kotlin/internal/utils:utils.bzl",
     _utils = "utils",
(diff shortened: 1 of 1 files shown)


kotlin/internal/jvm/impl.bzl
```diff
@@ -156,14 +155,33 @@ def kt_jvm_binary_impl(ctx):
...
     runtime_jars = depset(ctx.files._bazel_test_runner, transitive = [providers.java.transitive_runtime_jars])
+
+    test_class = ctx.attr.test_class
+
+    # If no test_class, do a best-effort attempt to infer one.
+    if not bool(ctx.attr.test_class):
+        for file in ctx.files.srcs:
+            if file.basename.split(".")[0] == ctx.attr.name:
+                for splitter in _SPLIT_STRINGS:
+                    elements = file.short_path.split(splitter, maxsplit = 1)
```
jin: Add some examples here on the string manipulation inputs and expected outputs?

cgruber: Not sure what you mean. Like a comment including examples like `common/foo/src/test/java/com/squareup/foo/TestFoo.kt` turning into `["common/foo", "com/squareup/foo/TestFoo.kt"]` ?

jin: Nvm, don't worry about it.

Issue #211, opened 2019-10-14
Broaden support for "friends" (access to `internal` members) in rules_kotlin


cgruber:
### Problem

Kotlinc supports a visibility `internal` which implies that all code with access to `internal` members are part of the same logical module.  Without modification, `internal` is granted to all code contained in the compilation unit.  However, there are several use-cases where code outside a compilation unit is to be appropriately granted access.  Examples of this are:

   1. Test code or test-library code which needs access to `internal` members of the code under test
   2. Implementation code which needs access to `internal` conveniences of public API which has been separated into a different compilation unit for build-system constraint (such as bazel's `visibility=` mechanism.
   3. Configuration code (e.g. dagger wiring) which needs to see `internal` implementation code such as internal constructors, etc., in order to create or otherwise configure them.
   4. Build flavors (e.g. src/debug/java) should be able to see main source code (e.g. src/main/java).

**kotlinc** supports these by providing a `-Xfriend-paths` flag, whose content is a classpath (.jars and/or class-folders) whose presence in the list is given as a signal that **kotlinc* should grant access to their `internal` members during this compilations.

**rules_kotlin** presently supports one specific and narrow use of `-Xfriend-paths`. `kt_jvm_test()` users can declare a `friends=` attribute pointing to a target containing the objects under test, and this will give the test access.  A library `B` which declares `A` as `friends=` shares its module definition (specifically, `KtJvmInfo` provider takes `A`'s `module` definition and uses it instead of generating its own, and using `kotlin_module=` and `friends=` attributes together is forbidden. 

Also, any target listed as `friends=` is added to the direct dependencies of that target. 

There are several problems with the present implementation, however, including the following:

  * There is a bug in the existing implementation hidden by a lax enforcement in kotlinc prior to 1.3.40 (see #210 for details)
  * The `friends=` mechanism is named as plural, but permits only a single friend target.
  * Friendship is (presently) not transitive, so `C--friends-->B--friends-->A` will result in `C` seeing `internal` in `B`, and `B` (during compilation) seeing `internal` in `A`, but `C` cannot see `internal` in `A`.  This prevents having, say, three tests, a shared test lib with `internal` and code under test containing `internal` and all working together.
  * (partly per #210, but also for these purposes) the whole compilation classpath is thrown at the `-Xfriend-paths` 

### Proposal

This issue proposes a reworking of the mechanism with one of two options on contraints.

[...]

Issue #210, opened 2019-10-14
`friends=` on tests is broken in 1.3.40+


cgruber:
Kotlinc prior to 1.3.40 used a more generous heuristic (a partial string match) on the contents of `-Xfriend-paths`.  This has caused a breakage when used with 1.3.40 or later versions of kotlinc in the present rules.  

The current rules throw the whole compilation class path at this flag, BUT, also use an incorrect path separator (namely `File.pathSeparator`).  However, the present implementation of kotlinc's JVM back-end does not use this separator, instead using a list-like property syntax, with comma-separated values (Note: The Javascript backend uses the system path separator).  This inconsistency is filed with Jetbrains as [KT-34277].  The currently rules therefore do not split the classpath given by `-Xfriend-paths` at all, returning a single-item list containing the whole string of the classpath.  The net effect is that prior to 1.3.40, the partial string match worked (though an over-zealous list of friends was supplied). As of 1.3.40, however, an exact match against the jarfile (implied by the given friends target) is used.  This fails to match against the string given to the flag.

The existing rules need to both switch to using the separator expected, and also need to back off of sending the compilation classpath, which results in access to internals that should not be granted. 

[KT-34277]: https://youtrack.jetbrains.com/issue/KT-34277

cgruber:
@jin and @djwhang and @timpeut for awareness. I have a plan to deal with this (and a PR), but this is for context.

gertvdijk:
Hi! I have issues with this and bisected the problem with rules_kotlin on a954e2ba578a1cf8ae687d90287cb97611590aaa. It happens when I use an extension function from another workspace as dependency.

```
/path/to/bazel-cache/execroot/workspace_name/src/test/kotlin/nl/myorg/mycomponent/subcomponent/MyTest.kt:252:40: error: cannot access 'toFooBar': it is internal in 'nl.myorg.mycomponent.subcomponent'
```

Here is `toFooBar` part of a kt_jvm dep from `@other_workspace//` and then it gets into trouble.

Could it be that paths to external workspaces are not taken into account?

(I could make a minimal working example to demonstrate the issue, but would take some time. Please let me know if you want me to take that time or that I could try something else or maybe I have missed something else?)

PR #216, opened 2019-10-16
Support friend access to a target's internal members.

cgruber:
> Note: This PR is against a design currently under discussion in #211. It's here for illustration, though it is complete, and a version of it is being used by Square internally. It's not intended to be reviewed as the chosen implementation, as there are some un-answered questions in the aforementioned design discussion. 

### Support `friend` access to a target's `internal` members.

Allow the rules to support kotlin `internal` visibility (usually accessible only to the compilation unit)
by allowing a library or test to specify that it is friends with another library like so:

```
kt_jvm_library(
    name = "foo",
    srcs = glob(["*.kt"]),
    friend = "//some/other:target",
    # ...
)
```

> Note: declaring friends of a kt_android_library requires adding `_kt` to the target, e.g.
> `//some/other:target_kt`.  This is because the `kt_android_library` rule is a macro with a
> `kt_jvm_library` under the hood, and the surfaced rule is an android rule which doesn't have
> kotlin aware bazel providers. This will be fixed in the future.

This grants access to `internal` members of `//some/other:target`. A library can only be friends
with one other library, which must be a kotlin-aware target (returns a KtJvmInfo from the rule). It inherits the module name from that library. Only one friend can be declared, because of the assumptions made here about module membership.  Since friendship can be transitive (see below), this constrains the visibility so it does not become an arbitrarily growing lattice of trust, defeating the purpose.

Very common use-cases for this are:

  * tests and test-libraries depending on internals of the system under test.
  * clusters of targets that represent one logical unit, with public, private,
     fake, testing-utilities, configuration, or other targets that comprise the
     whole unit.

Friendship has limited transitivity. Consider projects `C`, `B`, and `A` with a dependency
relationship `C->B->A`.  If `C` declares friendship in `B`, and `B` declares friendship with `A`,
then they are all treated as logically one module for `internal` purposes. However it doesn't
skip. Once the line of friendship is broken, a separate module is presumed by kotlinc. 

Addresses #211 (and #210 along the way).

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,7 @@ and `kt_js`, and `kt_android` typically applied to the rules (the exception bein
 `kt_android_local_test`, which doesn't exist. Use an `android_local_test` that takes a 
 `kt_android_library` as a dependency).
 
-Limited "friend" support is available, in the form of tests being friends of their library for the
-system under test, allowing `internal` access to types and functions.
+Friend support for tests and other use-cases (supports access to `internal` types and functions)
 
 Also, jvm rules support the following standard java rules attributes:
   * `data`
@@ -110,8 +109,8 @@ in your `WORKSPACE` file (or import from a `.bzl` file:
 ```
 load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kotlin_repositories")
 
-KOTLIN_VERSION = "1.3.31"
-KOTLINC_RELEASE_SHA = "107325d56315af4f59ff28db6837d03c2660088e3efeb7d4e41f3e01bb848d6a"
+KOTLIN_VERSION = "1.3.50"
+KOTLINC_RELEASE_SHA = "69424091a6b7f52d93eed8bba2ace921b02b113dbb71388d704f8180a6bdc6ec"
 
 KOTLINC_RELEASE = {
     "urls": [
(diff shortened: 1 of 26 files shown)


README.md
```diff
@@ -110,8 +109,8 @@ in your `WORKSPACE` file (or import from a `.bzl` file:
 ```
 load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kotlin_repositories")
 
-KOTLIN_VERSION = "1.3.31"
-KOTLINC_RELEASE_SHA = "107325d56315af4f59ff28db6837d03c2660088e3efeb7d4e41f3e01bb848d6a"
+KOTLIN_VERSION = "1.3.50"
```
cgruber: Note: Necessary to properly illustrate due to #210 - prior to ***1.3.40*** there is a different enforcement mechanism (a sloppy string comparison) which over-permits, concealing some issues.

kotlin/internal/jvm/compile.bzl
```diff
@@ -137,6 +137,21 @@ def _partition_srcs(srcs):
         src_jars = depset(src_jars),
     )
 
+def _handle_legacy_friends(ctx):
+    # TODO extract and move this into common. Need to make it generic first.
```
cgruber: Hmm. Need to fix that. This TODO is the point of the PR. :D

README.md
```diff
@@ -123,6 +122,43 @@ KOTLINC_RELEASE = {
...
+does not become an arbitrarily growing lattice of trust, defeating the purpose.
+
+Very common use-cases for this are:
+
+  * tests and test-libraries depending on internals of the system under test.
+  * clusters of targets that represent one logical unit, with public, private,
+     fake, testing-utilities, configuration, or other targets that comprise the
+     whole unit.
+
+Friendship has limited transitivity. Consider projects `C`, `B`, and `A` with a dependency
```
jin: Why is friendship transitive? Can we make it strict, like regular Java dependencies? It's easier to loosen transitivity than to make it stricter down the road.

kotlin/internal/jvm/jvm.bzl
```diff
@@ -161,6 +161,11 @@ _common_attr = utils.add_dicts(
             default = [],
             allow_files = False,
         ),
+        "friend": attr.label(
```
jin: The internal rules uses "friends". Please continue to use the plural form.

src/test/kotlin/io/bazel/kotlin/KotlinAssertionTestCase.kt
```diff
@@ -120,7 +120,7 @@ abstract class BasicAssertionTestCase {
         ProcessBuilder().command("bash", "-c", Paths.get(executable).fileName.toString())
             .also { it.directory(executable.resolveDirectory()) }
             .start().let {
-                it.waitFor(5, TimeUnit.SECONDS)
+                it.waitFor(10, TimeUnit.SECONDS)
```
cgruber: This just accounts for the e2e test now running a test with a 2-second delay in the production code.

README.md
cgruber: So the current implementation is that you declare one friend. This is how the orginal friends field worked for the test rule. So the transitivity lets the tree of things associated with the logical "module" to be realized.   If we went with the "friend-graph" approach (see #211 proposal) then it would be reasonable to make friends not-transitive (declare what you consume) but just have the "we're in the same logical module" constraint.

The problem, of course, is the same as iwht strict deps. There are lots fo cases where it's pure boilerplate.  Like how we export jsr-330 along with both dagger and guice, because there are no cases where you would declare the latter two and not the first. Likewise, this could force tests to explicitly declare their testlib and their system under test, when the testlib for that package always is friends with the system under test. So it could get needlessly verbose.  Because there's no performance gain, like with classpath ordering and strict-deps, I don't think it's a huge deal either way.

kotlin/internal/jvm/jvm.bzl
cgruber: Sigh. Ok. So if we do that, then we really should not have the one-friend-only restriction (e.g., friends-graph rather than friends-tree) otherwise this is terribly named. :/  But either way, I guess conformity iwth the internal is worth it.

cgruber:
I'm going to extract the very small part that fixes the bug (the comma vs. semicolon) into a separate PR, which doesn't affect the semantics of the feature the way this whole PR (and #211) does, I think, so I can include at least 1.3.40+ compatibility in the 1.3.0 release, regardless of what we decide about friends capabilities long-term.

README.md
cgruber: As @vRallev pointed out, according to the JetBrains folks, -Xfriend-paths is intended to be transitive, by design. We could make it strict, but that's constraining the compiler feature - though if we moved to the friends-graph approach it wouldn't precisely matter so much. But ultimately the flag is intended to be transitive, so I'd want to press for handling things the way Jetbrains intends their flag to be used.

PR #217, opened 2019-10-17
Fix builder crashes

cgruber:
A few factors have been adding up to compiler crashes without any output. (e.g. cgruber/rules_kotlin#24, which is relevant to this repo as well).  Several things are happening at once, all fixed in this PR.

  1. CompilationStatusExceptions are being handled, but not reported, so errors in the tooling are failing the build, but the information of what's going on isn't propagated.
  2. Java changed the way it reported version numbers from the system property `java.version` which caused the logic that chose whether to include --multi-release or not is wrong. This caused the jdeps processor to barf, which was swallowed by item 1 above.
  3. During debugging (or any other time bazel isn't cleaning its own working dirs up), executing the builder blows up when the source jar is created (because it already exists). 

These are all fixed here, along with some tidying, and an example workspace that exercises this code by using a dep that needs --multi-release to be set.  Adds it to the CI as well.  

This should unblock #206


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JDepsGenerator.kt
```diff
@@ -84,4 +80,16 @@ internal class JDepsGenerator @Inject constructor(
...
         }
     }
+}
+
+/**
+ * Extract the normalized major version from the java.version property.
+ *
+ * This is complex as later versions stopped reporting "1.x.y" format in favor of "x.y", so check
+ * for a major version of "1" and if it's so, use the minor version as the major one.
+ */
```
jin: can you add a comment here that the assumption here is called on the `java.version` property, and is assumed to be a well-formed version number?

cgruber: Sure!

examples/trivial/BUILD
```diff
@@ -0,0 +1,19 @@
+package(default_visibility = ["//visibility:public"])
+
+java_library(
+    name = "java_deps",
+    exports = [
+        "@maven//:com_expedia_graphql_kotlin",
+        "@maven//:com_expedia_graphql_kotlin_schema_generator",
+#        "@maven//:com_expedia_graphql_kotlin_federation",
```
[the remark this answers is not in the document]
cgruber: Done (and applied buildifier and ktlint)

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JDepsGenerator.kt
cgruber: Wait, I'm confused - isn't that what this comment says, if taken with the previous sentence (which refers to the java.version property directly)?

cgruber: I reworded the text. Hopefully that makes more sense.

PR #221, opened 2019-10-25
Stop passing source jars in kt_jvm_import() to the DefaultInfo(files=) attribute.

### LINKED ISSUE #208: android_binary is broken because it's not allowed to depend on .jar artifacts
Example code: https://github.com/fishy/onepwd

The current [master version](https://github.com/fishy/onepwd/tree/a24f4e9155a110e18699e756918eebd9618b163a) is using `rules_kotlin`@ 7543a917cda483c5856ec79b590343c9669eacd6 and it works. When I tried to upgrade it to newest master:

```
$ git diff
diff --git a/WORKSPACE b/WORKSPACE
index c25b6ea..236c6f5 100644
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -7,12 +7,12 @@ android_sdk_repository(

 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

-RULES_KOTLIN_VERSION = "7543a917cda483c5856ec79b590343c9669eacd6"
+RULES_KOTLIN_VERSION = "ab2f9f79912a35af3c1a07849d6fb457c2bfc54b"
 http_archive(
     name = "io_bazel_rules_kotlin",
     strip_prefix = "rules_kotlin-%s" % RULES_KOTLIN_VERSION,
     url = "https://github.com/bazelbuild/rules_kotlin/archive/%s.tar.gz" % RULES_KOTLIN_VERSION,
-    sha256 = "d335be363937ed62b5758cfdb33910ee52429835685345ef6b9f009c2bff9a9d",
+    sha256 = "1a18400bf92cba63554e2c71b9e74fa8569ab296e4954d232d5c643e19a2da02",
 )
 load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kotlin_repositories", "kt_register_toolchains")
```

I got this error when building `:app`:

```
ERROR: /Users/fishy/work/onepwd/BUILD:23:12: in deps attribute of android_binary rule //:app: Dependencies on .jar artifacts are not allowed in Android binaries, please use a java_import to depend on external/com_github_jetbrains_kotlin/lib/kotlin-stdlib-sources.jar. If this is an implicit dependency then the rule that introduces it will need to be fixed to account for it correctly.
```

Using `bazel query 'allpaths(//:app, @com_github_jetbrains_kotlin//:kotlin-stdlib)' --output graph` gives me this dependency graph:

dep

Another simpler android project have the same problem when upgraded `rules_kotlin` to ab2f9f79912a35af3c1a07849d6fb457c2bfc54b:

dep2

cgruber:
DefaultInfo(files=) gets passed to various things that interpret it as regular outputs, such as production jars, so things like source jars should not go in it. 

Fixes #208


cgruber:
I removed the usual reviewers, since this isn't up for review/acceptance, but for comment.

cgruber:
For the record, the example is a bit complex, to try to force a few scenarios - direct depending on a kt library, adding an intervening java target, etc.   the `//app:app` case should (in theory) mirror the case given in #208, unless I'm missing something.

cgruber:
Never mind, I found the bug, even though I can't replicate it.

examples/android/app/BUILD.bazel
```diff
@@ -0,0 +1,34 @@
+load("@build_bazel_rules_android//android:rules.bzl", "android_binary")
+
+# An app that consumes android-kt deps
+android_binary(
+    name = "app",
+    custom_package = "examples.android.app",
+    incremental_dexing = 0,
```
Kernald: Enabling the incremental dexing seems to reproduce the bug.

cgruber: Ohhhhh!  Interesting. Good to know.

I also found the error in the rules, but it's good to find a replication.

cgruber:
Ok, This should reproduce (and fix) the problem, and is suitable for review now.

examples/android/app/BUILD.bazel
cgruber: That was perfect, thank you. Definite replication. It fails before the final commit in this review is added.

Kernald:
Thanks for the fix! I confirm that it fixes #208 for me.

examples/android/WORKSPACE
```diff
@@ -0,0 +1,71 @@
...
+maven_install(
+    artifacts = [
+        "androidx.appcompat:appcompat:1.0.0",
+        "junit:junit:4.12",
+        "androidx.test.espresso:espresso-core:3.1.1",
+        "org.hamcrest:hamcrest-library:1.3",
+    ],
+    repositories = [
+        # Private repositories are supported through HTTP Basic auth
+        "http://username:password@localhost:8081/artifactory/my-repository",
```
jin: rm this

examples/android/WORKSPACE
```diff
@@ -0,0 +1,71 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+RULES_JVM_EXTERNAL_TAG = "2.8"
+
+RULES_JVM_EXTERNAL_SHA = "79c9850690d7614ecdb72d68394f994fef7534b292c4867ce5e7dec0aa7bdfad"
+
+ANDROID_API = 28
+
+ANDROID_TOOLS_VERSION = "28.0.2"
```
jin: let's not pin this. the CI may not always have 28.0.2.

examples/android/app/BUILD.bazel
```diff
@@ -0,0 +1,47 @@
...
+    custom_package = "examples.android.app",
+    incremental_dexing = 1,
+    manifest = "src/main/AndroidManifest.xml",
+    multidex = "native",
+    deps = [
+        "//lib",
+    ],
+)
+
+sh_test(
```
jin: you can use `build_test` for skylib for this.

cgruber: When did that get added!  Last I saw it was an internal google thing.  I'll update.

examples/android/WORKSPACE
cgruber: hah!  Totally cut and paste from the repro case.  Woops.

examples/android/WORKSPACE
cgruber: Done.

examples/android/app/BUILD.bazel
cgruber: Done

Issue #219, opened 2019-10-22
kt_js_import fails for kotlinx atomicfu import


jasonwyatt:
I'm seeing the following error when trying to set up a `kt_js_import` for AtomicFu:

```
ERROR: /Users/jwf/GitHub/[REDACTED]/BUILD:5:1: Action [REDACTED]/atomicfu-js.js failed (Exit 1) importer failed: error executing command bazel-out/host/bin/external/io_bazel_rules_kotlin/kotlin/internal/js/importer --jar external/maven/v1/https/jcenter.bintray.com/org/jetbrains/kotlinx/atomicfu-js/0.13.1/atomicfu-js-0.13.1.jar --out ... (remaining 3 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_jwf/c103462cb8cae3463e19adc48a04ad0a/sandbox/darwin-sandbox/87/execroot/__main__/bazel-out/host/bin/external/io_bazel_rules_kotlin/kotlin/internal/js/importer.runfiles/io_bazel_rules_kotlin/kotlin/internal/js/importer.py", line 64, in <module>
    _main(parser)
  File "/private/var/tmp/_bazel_jwf/c103462cb8cae3463e19adc48a04ad0a/sandbox/darwin-sandbox/87/execroot/__main__/bazel-out/host/bin/external/io_bazel_rules_kotlin/kotlin/internal/js/importer.runfiles/io_bazel_rules_kotlin/kotlin/internal/js/importer.py", line 51, in _main
    _extract_root_entry(args.jar, args.out)
  File "/private/var/tmp/_bazel_jwf/c103462cb8cae3463e19adc48a04ad0a/sandbox/darwin-sandbox/87/execroot/__main__/bazel-out/host/bin/external/io_bazel_rules_kotlin/kotlin/internal/js/importer.runfiles/io_bazel_rules_kotlin/kotlin/internal/js/importer.py", line 45, in _extract_root_entry
    raise ex
  File "/private/var/tmp/_bazel_jwf/c103462cb8cae3463e19adc48a04ad0a/sandbox/darwin-sandbox/87/execroot/__main__/bazel-out/host/bin/external/io_bazel_rules_kotlin/kotlin/internal/js/importer.runfiles/io_bazel_rules_kotlin/kotlin/internal/js/importer.py", line 38, in _extract_root_entry
    jar.read(basename)
[...]

PR #223, opened 2019-10-27
Use regular expressions to match JS filenames in kt_js_import.

jasonwyatt:
Adds a more nuanced approach to pulling `kt_js_import` files out of the jars from Maven.

Rather than `importer.py` looking for an exact match of the filenames its provided with: it now accepts regular expression patterns to match against when looking for output/auxillary files.

Also: removes the requirement to have the `name` match the name of the jar, because this lead to `module_name` issues in the built javascript. For example: the KotlinJS compiler uses `require('kotlinx-coroutines-core')` instead of `require('kotlinx-coroutines-core-js')`.

Finally: This PR adds example-usage of `kt_js_import` to the `express` nodejs example.

Fixes: #219


CONTRIBUTORS
```diff
@@ -10,3 +10,4 @@
 #     Name <email address>
 
 Hassan Syed <h.a.syed@gmail.com>
+Jason Feinstein <jwf@google.com>
```
cgruber: I think googlers are by definition assumed to be contributing. (I have no problems with adding explicit names here, but when I was at Google, we never used to include googlers, it was "Google engineers AND these folks" (see Guava for instance, where googlers appear only because of their pre-googler contributions)

kotlin/internal/js/impl.bzl
```diff
@@ -122,16 +122,16 @@ def kt_js_import_impl(ctx):
         fail("a single jar should be supplied, multiple jars not supported")
     jar_file = ctx.files.jars[0]
 
-    # Lock the jar name to the label name -- only make an exception for the compiler repo.
-    if not (ctx.label.workspace_root.startswith("external/") and ctx.label.workspace_root.endswith(_KT_COMPILER_REPO)):
-        expected_basename = "%s.jar" % ctx.label.name
-        if _strip_version(jar_file.basename) != expected_basename:
-            fail("label name %s is not the same as the jar name %s" % (jar_file.basename, expected_basename))
+    srcjar = None
```
cgruber: Nit: This could probably be a `foo if blah else baz` one-liner.

cgruber:
(Also, thanks for this!)

CONTRIBUTORS
jasonwyatt: Ah ok, I wasn't sure given the header for this file...

jasonwyatt:
I wasn't sure exactly what you meant WRT to copying examples-trivial and how that would play into the CI pipeline, but I was able to add another node example (fairly trivial), and created a test similar to `KotlinJvmDaggerExampleTest`

jasonwyatt:
Ah, nevermind, I see it now in `.bazelci`.  Will switch to that approach, since this approach seems to be failing (running :all_local_tests locally is fine though)

.bazelci/presubmit.yml
```diff
@@ -32,3 +32,12 @@ tasks:
...
       - //...
+  examples-nodejs:
+    name: Example - Node Helloworld
+    platform: ubuntu1804
+    working_directory: examples/node/coroutines-helloworld
+    include_json_profile:
+      - build
+      - test
+    build_targets:
+      - //coroutines-helloworld/...
```
cgruber: This is awesome.  Can you add building //express too?

jasonwyatt: Will-do!

cgruber:
Ready for me to merge? (Don't want to jump the gun if you have anything last minute).

jasonwyatt:
Ah, I'm ready whenever!

cgruber:
Done, though I probably now have to rebase my own work. :D

cgruber:
Hmm. Don't know why it undid my approval, but I did approve (and merged)

PR #233, opened 2019-12-05
Fix jdeps invocation

szinn:
Fix issue where jdeps was not being invoked correctly to compute .jdeps file


szinn:
@googlebot I consent

PR #234, opened 2019-12-05
Fixed invocation of jdeps

szinn:
Fix invocation of jdk jdeps to use the jar as the original source of code rather than a missing .jdeps file.


szinn:
This fixes https://github.com/bazelbuild/rules_kotlin/issues/235

szinn:
-v was added to the invocation to ensure that referenced libraries were also marked as explicit (jdeps seems to have different output for -v vs no -v for the same .jar file)

cgruber:
You ready for merge? Don't want to jump the gun.

PR #241, opened 2019-12-14
Add kotlin-reflect to jvm_runtime and jvm_stdlibs

szinn:
kotlin-reflect should be part of the standard libraries for Kotlin compiles

### DIFF
--- a/kotlin/internal/toolchains.bzl
+++ b/kotlin/internal/toolchains.bzl
@@ -117,7 +117,8 @@ _kt_toolchain = rule(
         "jvm_runtime": attr.label_list(
             doc = "The implicit jvm runtime libraries. This is internal.",
             default = [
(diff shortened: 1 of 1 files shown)


cgruber:
I'm not sure about this. The kotlin-reflect artifact is supplied separately precisely so as not to have it "come for free" where it's not needed.  This also doesn't match the maven dependency metadata for these, so in any other build system you'd need to add kotlin-reflect where you use it.  That seems to indicate (to me) that it shouldn't be "part of the standard library for kotlin compiles". No?

szinn:
SGTM - I was playing around with Guice using the maven_install rules to pull it in.  The kotlin compile indicated that the incorporated libraries also pulled in the standard libraries.  It seemed to me that having -reflect as part of the compile seemed not unreasonable

PR #242, opened 2019-12-14
Add InteliiJ aspect info

szinn:
- allows the bazel-intellij aspect processing to correctly identify the version
  of Kotlin being used and update the source accordingly
- part of the fix for https://github.com/bazelbuild/intellij/issues/1274

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,10 @@
 
 |  Date&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | News  | 
 | :----------- | -------- |
-| Dec 5, 2019 | Released version [1.3.0-rc2](https://github.com/bazelbuild/rules_kotlin/releases/tag/legacy-1.3.0-rc2). | 
-| Dec 5, 2019 | Fix for problem with jdeps generation (#235). | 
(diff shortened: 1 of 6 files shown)


cgruber:
So this seems reasonable to me, but I'm curious about passing it as a separate part of the struct, rather than baking it in to values in KtJvmInfo. Is there a strong reason to prefer that?  Is there prior art on passing toolchains in that providers struct?  

I see examples of passing back ToolchainInfo from rules that define the toolchain, but this feels different. I'm struggling to see a good example anywhere to resolve this in my mind. This isn't a toolchain rule you're editing.

Also, do you need the stdlib compilation path? (or is that just to force the stdlibs into the IDE thingy?)  Is this not something you can attach to an aspect?  The PR description references aspects, but I don't see an aspect given here (and the intellij aspect lives elsewhere).  Could you possibly get the ctx.rule.toolchain in the intellij aspect, to get this information?

szinn:
I was seeing that this kind of information was passed back in the toolchain as you described.  As well, the version of the tool is more a toolchain info, but could easily be passed in the KtJvmInfo as well and extracted from there.

The stdlib path was purely to fill out the intellij structure.  Perhaps it isn't needed there and just the version can be put into the KtJvmInfo.  Shall I make this change to the PR?

As well, this change does nothing for the tool, it is putting the information into the return structure so that it can be extracted by the intellij aspect code (which is the other PR I shared)

szinn:
@cgruber This change has been updated to put the language_version in the KtJvmInfo struct.

cgruber:
I'm just assuming this can be picked up on other side by the plugin (or its aspect).  But this change is fine, as it is, assuming something will use it.

szinn:
I have a change for bazel-intellij to pick up this change

cgruber:
I'll be releasing 1.3.1 this weekend or next week, I expect, so it'll be available very soon.

PR #255, opened 2019-12-19
Passthrough tags to the container android_library

justhecuke:
Tags were being dropped which meant that tag filters targets wasn't working properly.


cgruber:
That's great!  I'll be releasing an RC shortly (like, probably today) so this should be available nearly immediately.

PR #260, opened 2020-01-17
Fix some incompatible native android_* rules references

cgruber:
This only fixes these references within the library, and does not address third-party violations from things like @build_tools.  Also, updates and reorganizes the android sample's WORKSPACE file. Minor buildifier formatting also.

Partly addresses #251

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -13,6 +13,15 @@
 # limitations under the License.
 workspace(name = "io_bazel_rules_kotlin")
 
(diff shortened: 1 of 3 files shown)


cgruber:
crap - this change adds a real dependency on the android rules which, if you don't use android, bleeds into your project if you use the kotlin rules. 

I think this actually needs a bit of a project reorganization so it doesn't require the android stuff to be present in the midst of the non-android stuff. :/ 

(Thanks, CI, for pointing this out)

PR #263, opened 2020-01-29
Filter non-kotlin code out of generated sources

justhecuke:
Filters out non-JVM generated sources.

Annotation processors (like dagger) can generate non-JVM code (like `.dot` files) and this can cause an IllegalStateException due to having a non-JVM source file.

Since we cannot expect all annotation processors to only generate JVM source code, and to be generally useful for generic annotation processors, we need to gracefully handle generated non-JVM code rather than having a strict requirement that only JVM code can be generated since that would probably preclude the use of  many useful annotation processors.

Also, added a print of errors hit during compilation. Before it would simply error out mysteriously with no error message. Now, we will actually get the error message. It was very helpful in debugging this particular issue.


WORKSPACE
```diff
@@ -126,3 +126,8 @@ maven_install(
     ],
 )
 
+android_sdk_repository(
```
justhecuke: My local tests seem to have required this and I'm not sure why the buildkite test runs are fine without it.

justhecuke:
Any way to get someone to take a quick look? It's a very minor change that unblocks usage of some annotation processors.

WORKSPACE
jin: Can you remove build_tools_version and api_level, please? The CI may not have these versions pinned.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
```diff
@@ -210,6 +210,7 @@ class KotlinJvmTaskExecutor @Inject internal constructor(
         expandWithSources(
             File(directories.generatedSources).walkTopDown()
                 .filter { it.isFile }
```
jin: combine `it.isFile && IS_JVM_SOURCE_FILE.test(it.name)` for a single pass?

justhecuke: Done.

WORKSPACE
justhecuke: Done.

WORKSPACE
```diff
@@ -131,8 +131,12 @@ maven_install(
...
```
cgruber: Wait, why do you need this in the root workspace?

justhecuke: When I tried to run the tests locally it complained about an android sdk not having been configured. Adding this solved the problem. I can't recall the exact error message.

WORKSPACE
cgruber: Your local tests needed this? Which tests? This shouldn't be necessary. :/

WORKSPACE
justhecuke: Seems that this was added at head so it's a moot point now.

justhecuke:
> Also, can you please put in some sort of test for this behavior? I'm concerned that this will get broken quickly with some of the changes we're going to put in.

I'm in the process of adding tests. It's a bit more involved than I thought it would be.

justhecuke:
> > Also, can you please put in some sort of test for this behavior? I'm concerned that this will get broken quickly with some of the changes we're going to put in.
> 
> I'm in the process of adding tests. It's a bit more involved than I thought it would be.

I added a unit test although I had to change the visibility of some methods to get everything working.

I can also make a simple java plugin that generates a non-jvm file and make a E2E test of the kotlin compilation rules, but I wasn't sure if that was something that would be OK.

justhecuke:
@cgruber @restingbull I think this can be reviewed again.

justhecuke:
@cgruber friendly ping since the last one I gave was late on a friday.

restingbull:
@cgruber is unavailable at the moment. I'll squash and merge.

justhecuke:
Thanks.

cgruber:
Sorry for the dead air @justhecuke  - was travelling. Thanks for picking up the slack @restingbull

PR #268, opened 2020-02-01
Allow resources to be in a kotlin directory

rockwotj:
We just put everything in a `kotlin` directory (which the test rules allow for) so it'd be nice if resources also didn't need this prefix.

---
This change is [Reviewable](https://reviewable.io/reviews/bazelbuild/rules_kotlin/268)

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -52,6 +52,7 @@ def _fold_jars_action(ctx, rule_kind, output_jar, input_jars):
 _CONVENTIONAL_RESOURCE_PATHS = [
     "src/main/resources",
     "src/test/resources",
+    "kotlin",
 ]
 
 def _adjust_resources_path_by_strip_prefix(path, resource_strip_prefix):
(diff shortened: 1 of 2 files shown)


cgruber:
Do you put them in "<project>/kotlin" or do you mean "src/main/kotlin"? I'm a little hesitant to just make "kotlin" the split as it's not the most uncommon package name which could result in some bad (and breaking) source root inferences. I'd be more comfortable breaking on src/kotlin or src/*/kotlin.

rockwotj:
I put them in `/kotlin`, the `kt_jvm_test` rules already support this.

https://github.com/bazelbuild/rules_kotlin/blob/4512a83053489326a3643ef9d84e3e15420eb58e/kotlin/internal/jvm/impl.bzl#L157-L164

In general I'm not a fan of the unneeded maven structure of separating sources and tests.

I understand you can specify a prefix per binary, but we have many binaries and specifying the strip resource prefix is... annoying.

cgruber:
Ahhh. I'm sorry, I get it now. I realize I had made the change I was thinking about already, and you're expanding it. I apparently can't read a simple diff. :/

This is fine.

PR #271, opened 2020-02-05
Release the Kotlin rules in a simplified workspace relying on precompiled jars.

restingbull:
Minor fix for the kt_jvm_import to work with the IntelliJ Bazel Plugin (Issue: bazelbuild/intellij#1616)
Move android_sdk import to third_party
Introduce release_archive macro to package the repository
Add //:rules_kotlin_release to generate a .tgz suitable for release 

TESTED: All kotlin examples.


restingbull:
@googlebot I signed it!

Covered under the Square CLA

kotlin/internal/utils/packager.bzl
```diff
@@ -0,0 +1,151 @@
...
+
+def _rename_impl(ctx):
+    out_file = ctx.actions.declare_file(ctx.label.name + "/" + ctx.attr.target)
+    in_file = ctx.file.source
+    ctx.actions.run_shell(
+        inputs = [in_file],
+        outputs = [out_file],
+        progress_message = "%s -> %s" % (in_file, ctx.attr.target),
+        command = "mkdir -p {dir} && cp {in_file} {out_file}".format(
+            dir = ctx.label.name,
```
cgruber: The dir is the label.name? That's the target, yeah?  You don't want label.package?

src/main/kotlin/BUILD
```diff
@@ -48,13 +49,50 @@ java_binary(
...
+)
+
+release_archive(
+    name = "pkg",
+    srcs = [
+        ":compiler_lib.jar",
+    ],
+    # Generate the release build file, relying on precompiled jars.
+    dist_files = {
+        "BUILD": """
```
cgruber: Nit: You might want to break these build files into private vars, like _RELEASE_LIB_BUILD_CONTENT or something.  I find it makes the whole thing read more cleanly if it's done that way, because you don't end up with weird scoping/indenting oddities.

src/main/kotlin/BUILD
```diff
@@ -48,13 +49,50 @@ java_binary(
...
+    ":compiler_lib",
+    "@com_github_jetbrains_kotlin//:lib/kotlin-compiler.jar",
+  ],
+)
+""",
+    },
+    package_dir = "src/main/kotlin",  # explicitly set the package directory, as there are no parent release_archives.
+    renames = {
+        # must rename, otherwise it creates a cyclical dependency in the release.
+        ":builder_deploy.jar": "kotlin_deploy.jar",
```
cgruber: maybe "kotlinc_worker.jar". I don't think you're under any obligation to use _deploy suffix, and this is closer to what it is. I'd also name the import rule similarly, so when it shows up as a target, it's not confusing. "kotlin_deploy" isn't a really obvious name. And "import_builder" isn't "what it is" it's "what it does" - and in this case, I think the target describing what it is probably fits better.

third_party/BUILD
```diff
@@ -92,3 +93,10 @@ java_binary(
     visibility = ["//visibility:public"],
     runtime_deps = ["//third_party/jvm/org/pantsbuild:jarjar"],
 )
+
+java_import(
```
cgruber: Note - I really want to separate the android and jvm stuff so a pure jvm project doesn't get android shit dragged into it. Totally for the future, but just a thing to keep in mind.

WORKSPACE
```diff
@@ -80,6 +80,11 @@ rbe_autoconfig(
     name = "buildkite_config",
 )
 
+load(
```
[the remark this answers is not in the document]
restingbull: Buildifier. I never format if I don't have to.

examples/node/WORKSPACE
```diff
@@ -9,6 +9,9 @@ load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kotlin_repositories", "kt_reg
 kotlin_repositories()
 kt_register_toolchains()
 
+load("@io_bazel_rules_kotlin//third_party/jvm:workspace.bzl", "maven_dependencies")
```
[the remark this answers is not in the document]
restingbull: I'll update the buildkite and fix the examples.

kotlin/internal/jvm/jvm.bzl
```diff
@@ -336,6 +336,9 @@ kt_jvm_import = rule(
             doc = """The sources for the class jar.""",
             allow_single_file = True,
             cfg = "target",
+            # TODO(https://github.com/bazelbuild/intellij/issues/1616): Remove when the Intellij Aspect has the
+            #  correct null checks.
+            default = "@io_bazel_rules_kotlin//third_party:empty.jar",
```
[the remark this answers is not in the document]
restingbull: c'est la vie.

kotlin/internal/toolchains.bzl
```diff
@@ -195,3 +195,26 @@ def define_kt_toolchain(
...
+        name = "builder_debug_timings",
+        values = {"define": "kt_timings=1"},
+    )
+
+    native.config_setting(
+        name = "builder_debug_trace",
+        values = {"define": "kt_trace=1"},
+    )
+
+    native.toolchain_type(
```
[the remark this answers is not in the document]
restingbull: It's not moved -- the invocation is still in the same place, but the code isn't repeated.

Once we sort out the layout this will be resolved. Till then, it makes more sense to keep a single source of truth.

kotlin/internal/utils/packager.bzl
restingbull: Naw. Using the label name to generate a directory namespaces the file and avoids overlaps with existing files and outputs. The package would dup the directory tree and double package it.

third_party/BUILD
restingbull: We should be pulling the jar from the android sdk, not the jar from tools. It ignores the toolchain and sdk, so this is suspicious. It's on the list.

kotlin/internal/utils/packager.bzl
```diff
@@ -0,0 +1,151 @@
...
+
+    if "package_dir" not in kwargs:
+        pkg_name = native.package_name()
+        local_pkg_index = pkg_name.rfind("/")
+        if local_pkg_index > -1:
+            pkg_name = pkg_name[local_pkg_index:]
+        kwargs["package_dir"] = pkg_name
+
+    # release_archives are always public.
+    kwargs["visibility"] = ["//visibility:public"]
```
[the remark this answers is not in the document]
restingbull: Reworked to hack less.

kotlin/internal/utils/packager.bzl
```diff
@@ -0,0 +1,151 @@
...
+        srcs += [rename_name]
+
+    pkg_tar(
+        name = name,
+        srcs = srcs,
+        **kwargs
+    )
+
+_file_header = """
+# Copyright 2020 The Bazel Authors. All rights reserved.
```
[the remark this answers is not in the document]
restingbull: No more generated files.

src/main/kotlin/BUILD
restingbull: Made them separate files.

kotlin/internal/toolchains.bzl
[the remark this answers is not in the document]
cgruber: Make's sense.

kotlin/internal/utils/packager.bzl
cgruber: Ah, ok.

third_party/BUILD
cgruber: Nod.

cgruber:
Note, this PR sets the stage for what will become a new way of building from head - but at this point, this PR does not require such changes.

PR #280, opened 2020-02-13
Cleanup src tree

restingbull:
Split build target into separate packages for easier navigation.


restingbull:
@googlebot I consent.

PR #287, opened 2020-02-24
Remove tree artifacts

restingbull:
Small optimization for kotlin compilation. May result in improvements for remote builds.

Depends on: bazelbuild/rules_kotlin/pull/288


restingbull:
@googlebot I consent.

cgruber:
I consent. (In case it's me)

cgruber:
I consent

cgruber:
@googlebot I consent.

kotlin/internal/jvm/plugins.bzl
```diff
@@ -19,40 +19,51 @@ load(
...
-        "annotation_processors": "a serializeable list of structs containing annotation processor definitions",
+        "annotation_processors": "depset of structs containing annotation processor definitions",
     },
 )
 
-_EMPTY_PLUGIN_INFO = [KtJvmPluginInfo(annotation_processors = [])]
+_EMPTY_PLUGIN_INFO = [KtJvmPluginInfo(annotation_processors = depset())]
 
+# Mapping functions for args.add_all.
+# These preserve the transitive depsets until needed.
```
cgruber: Wait... they do? I thought `[x for x in y]` was a list comprehension. This doesn't convert `y` into a list?

cgruber:
I'm LGTM on these two subsequent commits, but let's get the other one merged and rebase this on that, so we can make sure I'm looking at the correct set of commits.

kotlin/internal/jvm/plugins.bzl
restingbull: If you use args.add_all(<depset<Foo>>, map_each=fooToString) the depset isn't flattened until the command line is needed. 

However, this looks like a partial change. I think I merged it wrong. They should be processor.class and processor.cp

I'll go look.

restingbull: A commit snuck on the wrong pr. Resolve.

cgruber:
Ok, so post-hoc LGTM I guess. :)

PR #298, opened 2020-03-06
Ensure that KotlionBuilder workers use a clean directory for each compilation.

restingbull:


src/main/kotlin/io/bazel/kotlin/builder/utils/WorkingDirectoryContext.kt
```diff
@@ -0,0 +1,34 @@
...
+
+/** WorkingDirectoryContext provides a consistent base directory that is removed on close. */
+class WorkingDirectoryContext(val dir: Path) : Closeable {
+  companion object {
+    fun newContext(): WorkingDirectoryContext = WorkingDirectoryContext(
+        Files.createTempDirectory("working"))
+  }
+
+  override fun close() {
+    Files.walk(dir).sorted(Comparator.reverseOrder()).forEach(Files::delete)
```
cgruber: I'm concerned about this - I believe walk(dir) follows symlinks, nad you want to delete the symlinks. 

Here's a good [stackoverflow](https://stackoverflow.com/questions/779519/delete-directories-recursively-in-java/27917071#27917071) with some thoughts about it

Also, guava (after 23, I believe) has [MoreFiles.deleteRecursively](https://guava.dev/releases/23.0/api/docs/com/google/common/io/MoreFiles.html#deleteRecursively-java.nio.file.Path-com.google.common.io.RecursiveDeleteOption...-) which has various symlink and security options.

src/test/kotlin/io/bazel/kotlin/builder/KotlinAbstractTestBuilder.java
```diff
@@ -124,35 +126,57 @@ public final Path writeGeneratedSourceFile(String filename, String[] lines) {
     return writeFile(DirectoryType.SOURCE_GEN, filename, lines);
   }
 
+  private static void assertFileExistence(Stream<Path> pathStream, boolean shouldExist) {
```
cgruber: This method is confusing. I don't see an actual existence test, just behavior based on the should boolean.

src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinWorkerTest.kt
```diff
@@ -0,0 +1,180 @@
...
+import java.nio.file.Files
+import java.nio.file.Path
+import java.util.zip.ZipEntry
+import java.util.zip.ZipFile
+import kotlin.streams.toList
+
+@RunWith(JUnit4::class)
+class KotlinWorkerTest {
+
+  private val wrkDir = Files.createTempDirectory("KotlinBuilderEnvironmentTest")
```
cgruber: Of course, if we WERE injecting FileSystem, then we could be using JIMFS. :)

src/main/kotlin/io/bazel/kotlin/builder/tasks/BazelWorker.kt
```diff
@@ -48,117 +54,109 @@ interface CommandLineProgram {
...
+            val invocationWorker = InvocationWorker(delegate, buffer)
+            while (true) {
+              val status =
+                  WorkerProtocol.WorkRequest.parseDelimitedFrom(realStdIn)?.let { request ->
+                    invocationWorker.invoke(loadArguments(request.argumentsList, true))
+                  }?.also { (status, log) ->
+                    with(WorkerProtocol.WorkResponse.newBuilder()) {
+                      exitCode = status
+                      output = log
+                      build().writeDelimitedTo(realStdOut)
```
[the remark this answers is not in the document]
restingbull: Yup.

src/main/kotlin/io/bazel/kotlin/builder/KotlinBuilderComponent.java
```diff
@@ -1,17 +1,18 @@
 /*
- * Copyright 2018 The Bazel Authors. All rights reserved.
```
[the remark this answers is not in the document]
restingbull: Arg. IntelliJ is getting in it again.

src/main/kotlin/io/bazel/kotlin/builder/toolchain/CompilationTaskContext.kt
```diff
@@ -22,151 +23,159 @@ import java.io.ByteArrayInputStream
...
-    private val executionRoot: String = Paths.get(
-        System.getProperty("user.dir")
-    ).toAbsolutePath().toString() + File.separator
-    private var timings: MutableList<String>?
-    private var level = -1
-    private val isTracing: Boolean
+class CompilationTaskContext(
+  val info: CompilationTaskInfo,
+  private val out: PrintStream,
+  private val executionRoot: String = FileSystems.getDefault().getPath("").toAbsolutePath()
```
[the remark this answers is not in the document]
restingbull: Paths.get is semi-discouraged, as it is not explicit about the file system.

At some point, I'll have cleaned up the apis enough to inject paths.

src/main/kotlin/io/bazel/kotlin/builder/utils/WorkingDirectoryContext.kt
restingbull: We don't care about following the symlink. We just want the reference gone -- we might even prefer it in the long run, as following the symlink and removing the file is an extra hop.

src/main/kotlin/io/bazel/kotlin/builder/tasks/KotlinBuilder.kt
```diff
@@ -47,16 +50,71 @@ class KotlinBuilder @Inject internal constructor(
...
+  override fun apply(workingDir: Path, args: List<String>): Int {
     val (argMap, context) = buildContext(args)
     var success = false
     var status = 0
     try {
       @Suppress("WHEN_ENUM_CAN_BE_NULL_IN_JAVA")
       when (context.info.platform) {
-        Platform.JVM -> executeJvmTask(context, argMap)
+        Platform.JVM -> executeJvmTask(context, workingDir, argMap)
         Platform.JS -> executeJsTask(context, argMap)
```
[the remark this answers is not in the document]
restingbull: Fixed. Missed the fact that js derived the working directory from the output.

src/test/kotlin/io/bazel/kotlin/builder/KotlinAbstractTestBuilder.java
restingbull: path.toFile().exists() is a test, albeit the File api, not the Path api.

It was like this when I got here -- Since it's private, it can wait until the test structure is refactored.

src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinWorkerTest.kt
restingbull: Well. Um. Not really. The compiler only consumes strings, not paths.

PR #315, opened 2020-04-16
Add build target for serialization compiler plugin.

jongerrish:
Add test to verify generated serializer() method.


jongerrish:
@googlebot I signed it!

PR #318, opened 2020-04-17
Fix missing plugin inputs

justhecuke:
java_plugin runtime jars were not being added to the KotlinCompile action's input list which meant that bazel was not properly ensuring that they were available.

This PR just adds the plugin jars we need to the action input and lets bazel handle the rest.

Before this, kotlin's plugin implementation was fundamentally broken and only worked by chance.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -225,6 +225,7 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar):
     friend = _compiler_friends(ctx, friends = getattr(ctx.attr, "friends", []))
     compile_deps = _compiler_deps(toolchains, friend, deps = ctx.attr.deps + ctx.attr.plugins)
     annotation_processors = _plugin_mappers.targets_to_annotation_processors(ctx.attr.plugins + ctx.attr.deps)
+    transitive_runtime_jars = _plugin_mappers.targets_to_transitive_runtime_jars(ctx.attr.plugins + ctx.attr.deps)
     plugins = ctx.attr.plugins
 
     _run_kt_builder_action(
@@ -236,6 +237,7 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar):
         friend = friend,
         compile_deps = compile_deps,
         annotation_processors = annotation_processors,
+        transitive_runtime_jars = transitive_runtime_jars,
         plugins = plugins,
         outputs = {
             "output": output_jar,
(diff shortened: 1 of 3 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -325,7 +325,9 @@ def _run_kt_builder_action(ctx, rule_kind, toolchains, dirs, srcs, friend, compi
 
     ctx.actions.run(
         mnemonic = "KotlinCompile",
-        inputs = depset(ctx.files.srcs, transitive = [compile_deps.compile_jars]),
+        inputs = depset(
+            ctx.files.srcs,
+            transitive = [compile_deps.compile_jars] + [ap.classpath for ap in annotation_processors.to_list()]),
```
justhecuke, line 330: I'm not sure if the `to_list()` is strictly required, but I'm not sure how this would handle a depset of structs rather than a depset of files.

justhecuke:
PTAL @cgruber

justhecuke:
PTAL @restingbull

kotlin/internal/jvm/compile.bzl
restingbull: We want to avoid to_list like the plague, unfortunately. It would be better to export the processor classpath from https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/internal/jvm/plugins.bzl#L36 to limit the number of times we call to_list. (probably should even happen there, tbh.

justhecuke:
> A little restructure to collect the classpaths in the aspect to avoid flattening a depset until action execution would go a long way.
> 
> qq -- what was the bug that came up from missing the classpath?

Bug is improper/non execution of plugins and sometimes missing classes when running the plugin. Errors during the build from dagger and such not generating classes.

Jars on the processorpath didn't exist and were pointing to the target config rather than host. Especially bad when building for android since it'd try to compile for arm, possibly meaning that valid java/kotlin wouldn't be able to build due to android sdk not using java/kotlin toolchain although I didn't witness this.

End result is flaky failures during build that generally required a `clean` to fix up. If you hit a build failure you tended to hit it 100% until a clean. If your build worked then it tended to work 100% until a clean. Especially bad in CI since builds were always clean.

kotlin/internal/jvm/compile.bzl
justhecuke: I think I've made the change you requested. I couldn't do it within the exact method you pointed to, not that I could figure anyways, but I've made changes similar to the existing pattern.

Issue #323, opened 2020-05-03
Missing file in the release archive


Kernald:
I'm trying to use the prebuilt release (1.4.0-rc1). In my `WORKSPACE`:

```
http_archive(
    name = "io_bazel_rules_kotlin",
    sha256 = "a8c40b42b74ce2f15a3a614cc80b8b7ac8f2ad7069731f5898b89c606811907b",
    url = "https://github.com/bazelbuild/rules_kotlin/releases/download/legacy-1.4.0-rc1/rules_kotlin_release.tgz",
)

load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kotlin_repositories", "kt_register_toolchains")

kotlin_repositories(
    compiler_release = {
        "sha256": "ccd0db87981f1c0e3f209a1a4acb6778f14e63fe3e561a98948b5317e526cc6c",
        "urls": [
            "https://github.com/JetBrains/kotlin/releases/download/v1.3.72/kotlin-compiler-1.3.72.zip",
        ],
    },
)

kt_register_toolchains()
```

Building anything fails:

```
ERROR: missing input file 'external/io_bazel_rules_kotlin/src/main/kotlin/compiler_lib.jar', owner: '@io_bazel_rules_kotlin//src/main/kotlin:compiler_lib.jar'
ERROR: /private/var/tmp/_bazel_kernald/3e87e5361dca121403e1b8033631cff0/external/io_bazel_rules_kotlin/src/main/kotlin/BUILD.bazel:25:1: @io_bazel_rules_kotlin//src/main/kotlin:builder: missing input file '@io_bazel_rules_kotlin//src/main/kotlin:compiler_lib.jar'
ERROR: /private/var/tmp/_bazel_kernald/3e87e5361dca121403e1b8033631cff0/external/io_bazel_rules_kotlin/src/main/kotlin/BUILD.bazel:25:1 1 input file(s) do not exist
INFO: Elapsed time: 12.819s, Critical Path: 0.37s
```

Looking at the content of this folder, I can see a `compiler.jar`, but no `compiler_lib.jar`.

cgruber:
Working to reproduce.

Kernald:
Closing as this has been fixed in #326 which has been released already.

PR #326, opened 2020-05-07
Fix missing file in the release archive

restingbull:
* fixes incorrect release_archive rules, looks like a bad merge. #323
* updated to use stardoc (so ugly).
* add a new release script to check the validity of release and generate new documentation.


cgruber:
I didn't see a test for the release artifact - I think we can add it in a separate PR, but we should maybe have one pipeline step that exercises that script.

PR #327, opened 2020-05-07
Fix a badly formatted comment killing skydoc, and add tests.

cgruber:


docs/kotlin.md
```diff
@@ -6,41 +6,40 @@
...
+            Define a plugin for the Kotlin compiler to run. The plugin can then be referenced in the `plugins` attribute
```
restingbull: Does this look right when processed? There shouldn't be an indent, otherwise the markdown fails.

cgruber: No, it's all bad, and I don't konw why. I reverted it - for now, I need the release script to not block. We can sort skydoc/stardoc/whatever issues separately.

Issue #325, opened 2020-05-06
Kotlin workers mask the underling issue


nkoroste:
Version: Legacy Kotlin Rules 1.4.0 (Release Candidate 1)

Noticed new issue with the new rules. For example if I have a missing dependency in a Kotlin target I get the following worker error that masks the underlying issue of the missing dep.

```
 failed: Worker process quit or closed its stdin stream when we tried to send a WorkRequest:

---8<---8<--- Exception details ---8<---8<---
java.io.IOException: Stream closed
	at java.base/java.lang.ProcessBuilder$NullOutputStream.write(Unknown Source)
	at java.base/java.io.OutputStream.write(Unknown Source)
	at java.base/java.io.BufferedOutputStream.flushBuffer(Unknown Source)
	at java.base/java.io.BufferedOutputStream.write(Unknown Source)
	at com.google.protobuf.CodedOutputStream$OutputStreamEncoder.doFlush(CodedOutputStream.java:3062)
	at com.google.protobuf.CodedOutputStream$OutputStreamEncoder.write(CodedOutputStream.java:2994)
	at com.google.protobuf.CodedOutputStream$OutputStreamEncoder.writeLazy(CodedOutputStream.java:3013)
	at com.google.protobuf.ByteString$LiteralByteString.writeTo(ByteString.java:1392)
	at com.google.protobuf.CodedOutputStream$OutputStreamEncoder.writeBytesNoTag(CodedOutputStream.java:2801)
	at com.google.protobuf.CodedOutputStream$OutputStreamEncoder.writeBytes(CodedOutputStream.java:2775)
	at com.google.devtools.build.lib.worker.WorkerProtocol$Input.writeTo(WorkerProtocol.java:235)
	at com.google.protobuf.CodedOutputStream$OutputStreamEncoder.writeMessageNoTag(CodedOutputStream.java:2855)
	at com.google.protobuf.CodedOutputStream$OutputStreamEncoder.writeMessage(CodedOutputStream.java:2824)
	at com.google.devtools.build.lib.worker.WorkerProtocol$WorkRequest.writeTo(WorkerProtocol.java:1081)
	at com.google.protobuf.AbstractMessageLite.writeDelimitedTo(AbstractMessageLite.java:95)
	at com.google.devtools.build.lib.worker.Worker.putRequest(Worker.java:127)
	at com.google.devtools.build.lib.worker.WorkerSpawnRunner.execInWorker(WorkerSpawnRunner.java:352)
[...]

PR #328, opened 2020-05-09
Fix the worker contract.

restingbull:
Restructure the worker for improved (hopefully) clarity and better management of empty pipes.
Adds BazelWorkerTest that checks the behaviours of both invocation workers and persistent.

Fixed: #325.

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/BazelWorker.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/BazelWorker.kt
@@ -16,23 +16,30 @@
  */
 package io.bazel.kotlin.builder.tasks
 
-import com.google.devtools.build.lib.worker.WorkerProtocol
+import com.google.devtools.build.lib.worker.WorkerProtocol.WorkRequest
+import com.google.devtools.build.lib.worker.WorkerProtocol.WorkResponse
 import io.bazel.kotlin.builder.utils.WorkingDirectoryContext
 import io.bazel.kotlin.builder.utils.wasInterrupted
+import java.io.BufferedInputStream
 import java.io.ByteArrayInputStream
 import java.io.ByteArrayOutputStream
+import java.io.Closeable
 import java.io.IOException
+import java.io.InputStream
 import java.io.PrintStream
 import java.nio.charset.StandardCharsets.UTF_8
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
+import java.util.logging.Level
+import java.util.logging.Logger
 
 /**
  * Interface for command line programs.
  *
  * This is the same thing as a main function, except not static.
  */
+@FunctionalInterface
 interface CommandLineProgram {
   /**
    * Runs blocking program start to finish.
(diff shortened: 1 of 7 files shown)


src/main/kotlin/io/bazel/kotlin/builder/tasks/BazelWorker.kt
```diff
@@ -56,107 +63,164 @@ interface CommandLineProgram {
...
-          else BazelWorker.ERROR_STATUS.also {
-            System.err.println(
-                "ERROR: Worker threw uncaught exception with args: ${args}"
-            )
-            e.printStackTrace(System.err)
+      val (status, exit) = WorkingDirectoryContext.newContext()
+        .runCatching {
+          request.argumentsList
+            ?.let {
+              maybeExpand(it)
```
cgruber, line 167: nit: I'd make these one-liners. But this isn't wrong.

src/main/kotlin/io/bazel/kotlin/builder/tasks/BazelWorker.kt
```diff
@@ -56,107 +63,164 @@ interface CommandLineProgram {
...
+                       e)
+            Status.ERROR to 1
+          } else {
+            Status.INTERRUPTED to 1
           }
-        } to buffer.toString()
+        }
+        .getOrThrow()
+
+      val response = with(WorkResponse.newBuilder()) {
```
cgruber, line 186: Nit: With is fine, but also
```
val response = WorkResponse.newBuilder().apply {
  output = String(io.execution.toByteArray(), UTF_8)
  exitCode = exit
  setRequestId(request.requestId)
}.build()
```

It's a more common pattern for working with java builders, though it's the same thing.

src/main/kotlin/io/bazel/kotlin/builder/tasks/BazelWorker.kt
```diff
@@ -56,107 +63,164 @@ interface CommandLineProgram {
...
+class InvocationWorker(
+  private val io: WorkerIO,
+  private val program: CommandLineProgram
+) : Worker {
+  val logger: Logger = Logger.getLogger(InvocationWorker::class.java.canonicalName)
+  override fun run(args: List<String>): Int {
+    return WorkingDirectoryContext.newContext().runCatching {
+      program.apply(dir, maybeExpand(args))
+    }.recover { e ->
+      logger.log(Level.SEVERE,
```
cgruber, line 217: Nit. static import Level and that can be a one-liner.

Makes it a wee bit easier to see what's happening with the return code. Also, is this a place where oen of those `return@`'s will work?

src/main/kotlin/io/bazel/kotlin/builder/tasks/BazelWorker.kt
```diff
@@ -56,107 +63,164 @@ interface CommandLineProgram {
...
-          } catch (e: IOException) {
-            throw RuntimeException(e)
-          }
-        }
+/** Defines the common worker interface. */
+interface Worker {
+  fun run(args: List<String>): Int
+}
+
+class WorkerIO(
```
cgruber, line 108: This whole design is so much cleaner, thank you.

src/main/kotlin/io/bazel/kotlin/builder/tasks/BazelWorker.kt
restingbull: I get curly bracket blindness when the density is too tight.

src/main/kotlin/io/bazel/kotlin/builder/tasks/BazelWorker.kt
restingbull: Ah, cargo-culted the old code. Much cleaner, as with requires using the builder function on the last line.

src/main/kotlin/io/bazel/kotlin/builder/tasks/BazelWorker.kt
restingbull: Still too long for one line, I'm afraid.

PR #329, opened 2020-05-11
Fix the yaml to exercise release artifact in build kite

restingbull:


src/main/kotlin/BUILD
```diff
@@ -54,7 +54,7 @@ release_archive(
         ":builder_deploy.jar": "kotlin_worker.jar",
         "BUILD.release.bazel": "BUILD.bazel",
     },
-    deps = [
-        "//src/main/kotlin/io/bazel/kotlin/compiler:pkg"
-    ]
+#    deps = [
```
cgruber, line 57: What's going on with this comment?

restingbull: Trying to make it fail. Still trying.

restingbull: Works!

Issue #188, opened 2019-05-21
JAVA_HOME doesn't exist within the Builder's environment (even with --action_env) so java code is compiled with the embedded JDK


cgruber:
Ran into the one case where compiling with JDK9+ (with target 1.8) and running on pre-JDK9 breaks (java.nio.ByteBuffer on 9 implements a non-covariant overload). 

In investigating, I discovered that within the working environment of the KotlinBuilder, JAVA_HOME is not set, regardless of the use of `--action_env` or other considerations. This causes the javac used to be the one from the embedded jdk, which in the case of use of java.nio.ByteBuffer, explosions at runtime with NoSuchMethod errors. 

The workaround is a simple cast, but there should be a way to specify the java environment to the kotlin rules, so that it does the right thing.

davidsantiago:
Aha! I believe I was [struggling with this a week or two ago](https://github.com/bazelbuild/bazel/issues/8313), at least it was what confused the heck out of me to the point that I filed that issue. I was trying to build a Kotlin 1.3 project, but was getting [an error message](https://youtrack.jetbrains.com/issue/KT-28587) that appears related to jdk11 being used to build Kotlin 1.3. As I don't have JDK11 installed, but it appears to be what is used as the embedded JDK, I think your catch would explain what I was seeing.

davidsantiago:
If you get a second, can you elaborate on the "simple cast" workaround? I looked at the KotlinBuilder source code, but can't figure out how to make use of this workaround.

cgruber:
Sorry - you cast the `ByteBuffer` to just `Buffer`, which forces the compiler to link against the intended method.

cgruber:
That doesn't solve all the problems, but it solves these coveriant overload issues.

ThomasCJY:
Any update on this? Not being able to configure the java version correctly really tied our hands on the bazel migration.

nkoroste:
Shouldn't Kotlin rules honor `--java_toolchain=@bazel_tools//tools/jdk:toolchain_java8` flag?

ThomasCJY:
I have a tmp work around for this issue but it will breaks the cache since JAVA_HOME will be different for builder https://github.com/bazelbuild/rules_kotlin/pull/330 https://github.com/bazelbuild/bazel/issues/11382. To fix the root issue, we might need to pull in the embdded_jdk8 from bazel and use that during compilation.

cgruber:
[...]

PR #330, opened 2020-05-15
[rfc] Enable `use_default_shell_env` in kotlin compiling action to allow passing in `action_env`

ThomasCJY:
### Background
https://github.com/bazelbuild/rules_kotlin/issues/188
Previously, `action_env` cannot be passed into kotlin compile environment due to the reason that `use_default_shell_env` is not set. This PR fixes it.

### Regrading Kotlin being compiled by Jdk11 Issue
There are two issues under the hood which leads to the annoying fact that kotlin target is always being compiled using jdk11.

**Problem 1**
The first problem is as https://github.com/bazelbuild/rules_kotlin/issues/188 stated, `JAVA_HOME` hasn't been defined in kotlin rule's compiling sandbox environment. As a result, it's using the default jdk11 which comes with bazel tool. The fix for this problem, is to enable `use_default_shell_env` in kotlin rule's compile action so that it will take in the `JAVA_HOME` we defined locally.  

**Note: It is required to specify `--action_env=JAVA_HOME` in `.bazelrc` file of your codebase to make sure this env variable is passed to default environment [doc link](https://bazel.build/designs/2016/06/21/environment.html#new-flag---action_env)** 

**Problem 2**
However, even after fixing the first problem, the kotlin target doesn't compile and keep throwing NPE error, which is caused by having two different jdk version being loaded into classloader.
```
java.lang.NullPointerException
	at com.sun.tools.javac.file.Locations.getPathEntries(Locations.java:149)
	at com.sun.tools.javac.file.Locations.getPathEntries(Locations.java:134)
```

This problem, is the result of 
bazel build tool's Java Builder Stub Action is by default using jdk11 as javabin https://github.com/bazelbuild/bazel/issues/11382 and kotlin_rule is using the stub java action directly

Fix for problem 2 see https://github.com/bazelbuild/bazel/pull/11402

**Other options**
- Can we reuse javaToolchain? 
  - JavaToolchain's logic is embedded inside bazel's internal codebase. Kotlin rule, as an external rule, cannot easily access and utilize it. 
- We've already define `jvm_target` inside KotlinToolChain, why does it not help?
  - Well, this `jvm_target` is never used anywhere in kotlin rule's code. Probably never been used correctly.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -338,6 +338,9 @@ def _run_kt_builder_action(ctx, rule_kind, toolchains, dirs, srcs, friend, compi
         execution_requirements = {"supports-workers": "1"},
         arguments = [args],
         progress_message = progress_message,
+        # In order to make kotlin targets being compiled by java 8. Please define --action_env=JAVA_HOME
+        # in the `.bazelrc` file in your repository to pass in JAVA_HOME to default shell
+        use_default_shell_env = True,
         env = {
             "LC_CTYPE": "en_US.UTF-8",  # For Java source files
         },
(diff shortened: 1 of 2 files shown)


restingbull:
An alternative would be to compile the java sources using java_common -- as opposed to the current approach of invoking the java compiler directly.

PR #338, opened 2020-06-08
Fix deep dependency generation

szinn:
Consider this code:
```
package example

import io.ktor.http.HttpMethod
import io.ktor.http.HttpStatusCode
import io.ktor.server.testing.handleRequest
import io.ktor.server.testing.withTestApplication
import org.assertj.core.api.Assertions.assertThat
import org.junit.jupiter.api.Test

class ExampleTest {
    @Test
    fun testRoot() {
        withTestApplication({ module(testing = true) }) {
            handleRequest(HttpMethod.Get, "/").apply {
                assertThat(response.status()).isEqualTo(HttpStatusCode.OK)
                assertThat(response.content).isEqualTo("<html><table></table></html>")
            }
        }
    }
}
```
with the build file
```
package(default_visibility = ["//visibility:private"])

kt_junit_test(
    name = "Example",
    srcs = glob([
        "*/**",
    ]),
    size = "small",
    main_class = "TestRunner",
    deps = [
        "//src:example",
        "@maven//:io_ktor_ktor_server_netty",
        "@maven//:io_ktor_ktor_server_core",
        "@maven//:io_ktor_ktor_server_tests",
        "@maven//:io_ktor_ktor_server_test_host",
    ],
    args = [
        "--select-package",
        "example",
    ],
)
```
In this case, IntellliJ complains about the handleRequest entry because it cannot find a
base class for the TestApplication.  The base class is in io_ktor_ktor_server_host_common but
doesn't get included in the dependencies.

This change does a deeper search for all dependencies and marks the top level ones as explicit and
the deeper ones as implicit and resolves the missing dependency

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsParser.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsParser.kt
@@ -26,12 +26,8 @@ internal class JdepsParser private constructor(
   private val filename: String,
   private val isImplicit: Predicate<String>
 ) {
-  private val packageSuffix: String = " ($filename)"
-
   private val depMap = HashMap<String, Deps.Dependency.Builder>()
-  private val packages = HashSet<String>()
-
-  private var mode = Mode.COLLECT_DEPS
+  private val moduleDeps = HashMap<String, MutableList<String>>()
 
   private fun consumeJarLine(classJarPath: String, kind: Deps.Dependency.Kind) {
     val path = Paths.get(classJarPath)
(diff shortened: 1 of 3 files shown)


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsParser.kt
```diff
@@ -57,51 +53,29 @@ internal class JdepsParser private constructor(
...
-      Mode.COLLECT_PACKAGES_JDK8 -> when {
-        trimmedLine.endsWith(packageSuffix) -> packages.add(
-          trimmedLine.substring(
-            0,
-            trimmedLine.length - packageSuffix.length
-          )
-        )
-        trimmedLine.startsWith("-> ") -> {
-          // ignore package detail lines, in the jdk8 format these start with arrows.
+    val parts = line.split(" -> ".toRegex()).dropLastWhile { it.isEmpty() }.toTypedArray()
```
cgruber, line 64: Is `" -> ".toRegex()` interned? This feels like it should be constant-ish, especially if it's going to do a Regex compilation on every line.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsParser.kt
```diff
@@ -57,51 +53,29 @@ internal class JdepsParser private constructor(
...
+    if (parts.size == 2 && parts[1].endsWith(".jar")) {
+      addModuleDependency(parts[0], parts[1]);
```
cgruber, line 66: worth a `trim()`? Not sure if the output needs it.

szinn: Not sure what you mean here

cgruber: You split on that regex, but do the parts not need string trimming? They may not, if there's no spare whitespace in the file, but just checking.

szinn:
@restingbull It shall be done and was done

szinn:
It worked with my project which is Java 11 based - Not sure if there are any Kotlin Java 8 based projects that use this.  I did generate the test output from Java 8 jdeps and noted it had the same format as the Java 11 version.

PR #359, opened 2020-08-26
Added output of diagnostics

andrefmrocha:
In the light of the developments regarding [BSP support for bazel](https://github.com/bazelbuild/bazel/pull/11766), added the output of a diagnostics file for `kotlinc` compilation.

With this, diagnostics information will be served to clients through BSP.


cgruber:
Looking at this, the core mechanism of piping diagnostics through BEP isn't merged in bazel, and has stalled. We can't really move forward with this until it is. I'm going to close for now, but I'd be willing to look at a replacement or reopened PR once that's in.

PR #360, opened 2020-08-27
Compile Java source generated from annotation processors run by kapt

jongerrish:
Where kapt runs annotation processors that generate Java source code (e.g: Dagger) that generated Java source must be compiled with any project Java source code.


jeffzoch:
@jongerrish Do you know if this would fix #412?

jongerrish:
@jeffzoch - this is an old PR but all of this work has already been committed. I've left a comment on your issue.

PR #361, opened 2020-08-27
Support multiplexed workers.

jongerrish:
See: https://docs.bazel.build/versions/2.0.0/multiplex-worker.html


jongerrish:
Need more than just enabling the flag. I think @restingbull has a full implementation for this

PR #363, opened 2020-08-28
Kotlin 1.4

### LINKED ISSUE #295: Validate rules against 1.4.x kotlinc.
Moving from 1.2 to 1.3, there were structure differences in the kotlinc zip file, so we should make sure that all the present rules assumptions apply to 1.4, or that we account for the differences.

fkorotkov:
Just bumped all the version. Seems there were not much changes on the compiler side since 1.3.x beside adding new language features and removing deprecating stdlib methods. Should be pretty painful update. 

Fixes #295


fkorotkov:
@googlebot I signed it!

kotlin/internal/repositories/setup.bzl
```diff
@@ -23,32 +23,33 @@ def kt_configure():
...
             "com.google.code.findbugs:jsr305:3.0.2",
-            "junit:junit:4.13-beta-3",
-            "com.google.protobuf:protobuf-java:3.6.0",
-            "com.google.protobuf:protobuf-java-util:3.6.0",
-            "com.google.guava:guava:27.1-jre",
-            "com.google.truth:truth:0.45",
-            "com.google.auto.service:auto-service:1.0-rc5",
-            "com.google.auto.service:auto-service-annotations:1.0-rc5",
+            "junit:junit:4.13",
+            "com.google.protobuf:protobuf-java:3.13.0",
```
restingbull, line 32: Is there a reason why these are rolled up into 1.4?

restingbull:
Brief pause on this -- I'm seeing some oddities where using 1.3 is getting 1.4 libraries without changing the default version.

kotlin/internal/repositories/setup.bzl
fkorotkov: Nope. I just update everything I found. Can revert and leave only kotlin and coroutines.

restingbull: Please -- best to keep things focused.

fkorotkov: Done.

fkorotkov:
@restingbull if oddities are hard to investigate, alternatively we can keep the 1.3 default and just allow 1.4 language version with a custom toolchain.

rockwotj:
> @restingbull if oddities are hard to investigate, alternatively we can keep the 1.3 default and just allow 1.4 language version with a custom toolchain.

I'd really love to move forward with 1.4 - can we do this and document the 1.4 custom toolchain? I think a separate change could make it the default...

/cc @restingbull @cgruber @fkorotkov

cgruber:
If I understand comments from @restingbull - this can't work as-is, as there are some issues with hand-crafted classpaths in the builder that need fixing. I believe he's going to take a crack at it shortly, but we've been on-deadline internally, and haven't had the time to take care of this.

We absolutely do need 1.4 support ourselves, and soon, so this will get done.

cgruber (approved):

cgruber:
I've merged this, as despite us having some theoretical misgivings that we're missing something, we're just not moving forward fast enough.  So this way the community is freed up to detect it.

rockwotj:
Thanks @cgruber should we wait for a release or just point something to master?

cgruber:
You can point to master if you want, but I plan to release in a day or two. There are one or two last little fixes i would like to get in together, but if they take more than tomorrow I'll roll a release on Friday.  Nothing ever stops you from pointing to a specific commit though. 

That said, please remember that if you point at master, and not from a release binary, you'll need to do the extra developer setup, so you can build all the tooling itself. The release package will have all pre-built binaries.

rockwotj:
@cgruber I finally got around to testing this with my company's codebase and 1.4 works beautifully. We're planning a release tomorrow using master.

Is there a reason https://github.com/bazelbuild/rules_kotlin/releases/tag/legacy-1.4.0-rc4 didn't include this PR? (see [here][1] for example)

[1]: https://github.com/bazelbuild/rules_kotlin/blob/legacy-1.4.0-rc4/kotlin/internal/toolchains.bzl#L95-L99

PR #364, opened 2020-09-03
Add 0 to allowed package name characters

jyuhengwei:
For some unknown reason, `0` is the only alphanumeric character that wasn't recognized as in a package name. This led to odd scenarios that when you tried to compile code that had a `0` in the package name, the kotlin builder would package the code as the string before the `0` which led to some problems in auto-generated code.


jyuhengwei:
@googlebot I signed it!

restingbull (changes requested):
Add a test, so we don't do this again?

jyuhengwei:
> Add a test, so we don't do this again?

I added a test.

jyuhengwei:
@restingbull Any chance of a re-review?

restingbull:
Apologies, missed the email.

restingbull (approved):

jyuhengwei:
> Apologies, missed the email.

No problem. It happens.

jyuhengwei:
@restingbull Do I need to do something else to get this merged?

restingbull:
Apologies -- I forgot you can't merge it.

Issue #365, opened 2020-09-23
'noarg' and 'allopen' plugins fail to compile correctly when used together


dotdat:
Simplified use case:
https://github.com/dotdat/rules_kotlin/tree/example_allopen_with_noarg/examples/plugin/src/allopennoarg

It appears that when both noarg and allopen plugins are introduced, neither of them work.
When running the command:
`bazel build examples/plugin/src/allopen_with_noarg:user_is_open_test`

The following output is produced:
```
INFO: Analyzed target //examples/plugin/src/allopennoarg:user_is_open_test (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /home/user/Development/rules_kotlin/examples/plugin/src/allopennoarg/BUILD:48:1: KotlinCompile //examples/plugin/src/allopennoarg:user_is_open_test { kt: 1, java: 0, srcjars: 0 } for k8 failed (Exit 1)
examples/plugin/src/allopennoarg/UserIsOpenTest.kt:5:18: error: this type is final, so it cannot be inherited from
class Subclass : User(UUID.randomUUID(), "test@example.org")
                 ^
Compilation failure: compile phase failed:
/home/user/.cache/bazel/_bazel_user/7eb9b19529baec16e97dbc2b12ec3ade/execroot/io_bazel_rules_kotlin/examples/plugin/src/allopennoarg/UserIsOpenTest.kt:5:18: error: this type is final, so it cannot be inherited from
class Subclass : User(UUID.randomUUID(), "test@example.org")
                 ^
Target //examples/plugin/src/allopennoarg:user_is_open_test failed to build
INFO: Elapsed time: 0.226s, Critical Path: 0.13s
INFO: 0 processes.
FAILED: Build did NOT complete successfully

```

Running `bazel run examples/plugin/src/allopen_with_noarg:user_has_noarg_constructor_test` results in:
```
INFO: Analyzed target //examples/plugin/src/allopennoarg:user_has_noarg_constructor_test (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //examples/plugin/src/allopennoarg:user_has_noarg_constructor_test up-to-date:
  bazel-bin/examples/plugin/src/allopennoarg/user_has_noarg_constructor_test.jar
INFO: Elapsed time: 0.120s, Critical Path: 0.00s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
[...]

PR #366, opened 2020-09-28
Separate Xplugin options to support multiple compiler plugins at a time

dotdat:
When two plugins are joined together by `:` in the same -Xplugin option, kotlinc does not apply these. Separating out the options appears to solve the issue.

Fixes https://github.com/bazelbuild/rules_kotlin/issues/365


dotdat:
@googlebot I signed it!

Issue #367, opened 2020-09-29
Allow inline classes via -Xinline-classes


jeffzoch:
Inline classes are an experimental yet widely used feature that should be enabled given that there is currently no way to pass flags to kotlinc via kotlin_rules. This will allow people to use inline classes like Result (or define their own).

jeffzoch:
#23 is the bigger issue encompassing this one

phildawson-ct:
It looks like this was added in #421 so I think this issue is out of date?

PR #368, opened 2020-09-29
Add -Xinline-classes to kotlinc args

jeffzoch:
For #367 

Does not address long-term stated goals of providing a mechanism to do this.


jeffzoch:
@googlebot I signed it!

restingbull:
We aren't going to increase the free kotlinc arguments -- to best handle the variations and change in the compiler we need to define them in a structured manner.

PR #377, opened 2020-11-04
Support multiple source jars

jongerrish:
Including multiple source jars currently does not work as SourceJarExtractor fails to overwrite META-INF/MANIFEST.MF which is present in all jars.

Fix this by having SourceJarExtractor.preWrite() return true only for directories and files that match the extraction pattern.

### DIFF
--- a/examples/dagger/BUILD
+++ b/examples/dagger/BUILD
@@ -34,11 +34,29 @@ rm TeaPot.kt
     toolchains = ["@bazel_tools//tools/jdk:current_host_java_runtime"],
 )
 
+genrule(
+    name = "chai_lib_src",
+    outs = ["chai_lib_src.srcjar"],
+    cmd = """
+cat << EOF > ChaiCup.kt
+package chai
+object ChaiCup {
+    fun isEmpty() = true
+}
+EOF
+$(JAVABASE)/bin/jar -cf $@ ChaiCup.kt
+rm ChaiCup.kt
+""",
+    toolchains = ["@bazel_tools//tools/jdk:current_host_java_runtime"],
+)
+
+
 kt_jvm_library(
     name = "coffee_lib",
     srcs = glob(["src/**"]) + [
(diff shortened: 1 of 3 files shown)


nkoroste:
Nice fix but wrong repo? though would be nice to upstream it while we're at it :)

cgruber:
@nkoroste - why is this the wrong repo?

examples/dagger/BUILD
```diff
@@ -34,11 +34,29 @@ rm TeaPot.kt
...
+    toolchains = ["@bazel_tools//tools/jdk:current_host_java_runtime"],
+)
+
+
 kt_jvm_library(
     name = "coffee_lib",
     srcs = glob(["src/**"]) + [
         # Adding a file ending with .srcjar is how code generation patterns are implemented.
         ":tea_lib_src",
+        ":chai_lib_src",
```
cgruber, line 59: Just to confirm, adding this here fails when the SourceJarExtractor change isn't present, right? Do we also have confirmation that this source is compiled? (that is, it doesn't merely "not break" from a starlark-related constraint, but actually gets the expected compiled output?)

cgruber (changes requested):

examples/dagger/BUILD
jongerrish: Good call. Included code from `:chai_lib_src` in the Coffee app to demonstrate it gets compiled.

cgruber (approved):

examples/dagger/src/coffee/CoffeeApp.kt
```diff
@@ -30,7 +31,7 @@ class CoffeeApp {
   companion object {
     @JvmStatic
     fun main(args: Array<String>) {
-      if (TeaPot.isEmpty()) {
+      if (TeaPot.isEmpty() && ChaiCup.isEmpty()) {
```
cgruber, line 34: Nice.

PR #379, opened 2020-11-05
Rework actions for building Kotlin + Java.

jongerrish:
Rework the actions involved in building Kotlin + Java.

This is upstreamed from our internal fork and brings some improved performance and correctness.

1. Extract KAPT to separate action - this allows for some action caching in some cases, e.g: In the case of a Kotlin non-ABI affecting change the JavaBuilder action is cached. 
1. Prefer header jars generated by Java builder (there is no need to run ijar)
1. Build source jar in separate action (off the critical path since it is now run outside the Kotlin builder)
1. Use java_common.compile() for main code path (no need to build Java code in Kotlin builder anymore even when not using experimental ABI jar mode)
1. Run ABI plugin in same action as main kotlinc (unfortunately there is no performance advantage to running separately since a full compile always happens anyway, I created https://youtrack.jetbrains.com/issue/KT-43086 to track having a liteparse version of ABI plugin)
1. Specify the same Java compiler options as used for in the in process Java compilation, in particular disable annotation processors in java_common.compile() since they ran already in KAPT


kotlin/internal/jvm/compile.bzl
```diff
@@ -162,9 +190,11 @@ def _compiler_directories(ctx):
     """Creates a dict of the necessary compiler directories for generating compile actions"""
     return dict(
         classdir = _output_dir_path(ctx, "jvm", "classes"),
+        abi_classdir = _output_dir_path(ctx, "jvm", "abi_classes"),
```
jongerrish, line 193: I'll probably want to create a separate dictionary for KotlinKAPT + KotlinCompile actions which will avoid creating directories which are not needed in either case, e.g: no need for `kapt_gen_stub_dir` when running regular `kotlinc`

kotlin/internal/jvm/compile.bzl
```diff
@@ -237,81 +267,174 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar, compile_jar):
     dirs = _compiler_directories(ctx)
     srcs = _partitioned_srcs(ctx.files.srcs)
     friend = _compiler_friends(ctx, friends = getattr(ctx.attr, "friends", []))
-    compile_deps = _compiler_deps(toolchains, friend, deps = ctx.attr.deps + ctx.attr.plugins)
+    compile_deps = _compiler_deps(toolchains, friend, deps = ctx.attr.deps)
```
jongerrish, line 270: I removed plugins from compiler deps because `java_common.compile()` "discovers" plugins on the classpath and runs them, whether or not you specify `-proc:none`

kotlin/internal/jvm/compile.bzl
```diff
@@ -237,81 +267,174 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar, compile_jar):
...
             },
-            mnemonic = "KotlinCompileAbi",
+            build_java = False,
+            build_kotlin = False,
+            mnemonic = "KotlinKapt",
         )
-        if not srcs.java:
-            compile_jar = kt_compile_jar
+        generated_src_jars.append(kapt_generated_src_jar)
+        output_jars.append(kapt_generated_class_jar)
```
jongerrish, line 307: Hmm, actually, I don't think we want to merge `kapt_generated_class_jar` into the outputs here.

I _believe_ this jar is the compiled KAPT generated sources (perhaps stub versions), so I can try including them on the `KotlinCompile` classpath. It should be faster than parsing the sources, and if they're actually the stubs would allow action caching if the KAPT sources don't have API changes.

 I _think_ we could possibly pass this jar on the classpath to `kotlin` rather than the generated sources. I can give this a try.

kotlin/internal/jvm/compile.bzl
```diff
@@ -237,81 +267,174 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar, compile_jar):
...
+            dirs = dirs,
+            srcs = srcs,
+            generated_src_jars = generated_src_jars,
+            friend = friend,
+            compile_deps = compile_deps,
+            annotation_processors = [],
+            transitive_runtime_jars = transitive_runtime_jars,
+            plugins = plugins,
+            outputs = outputs,
+            build_java = False,
```
jongerrish, line 339: I left these janky `build_java` & `build_kotlin` flags in here, but I'm hoping we can just remove the in-KotlinWorker Java compilation and simplify things.

kotlin/internal/jvm/compile.bzl
```diff
@@ -237,81 +267,174 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar, compile_jar):
...
+            build_kotlin = True,
+            mnemonic = "KotlinCompile",
+        )
+
+        compile_jars.append(kt_compile_jar)
+        output_jars.append(kt_runtime_jar)
+        if not annotation_processors:
```
jongerrish, line 346: Having KAPT as a separate action means `kotlinc` and `JavaBuilder` can run in parallel. If there are no annotation processors then it has to happen in sequence.

What we could consider doing is to always run KAPT and when there are no annotation processors just set `aptMode=stubs`

Stub generation seems to be super fast comparatively. It would simplify the code and allow `CompileKotlin` and `JavaBuilder` to always run in parallel which is probably good for small incremental builds.


```
build //src/test/data/jvm/kapt:ap_kotlin_mixed

info: [kapt] Java stub generation took 23 ms
info: [kapt] Annotation processing took 270 ms
```

kotlin/internal/jvm/compile.bzl
```diff
@@ -237,81 +267,174 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar, compile_jar):
...
+        ]
+
+        # Kotlin takes care of annotation processing.
+        if len(srcs.kt) > 0:
+            javac_opts += ["-proc:none"]
+
+        java_info = java_common.compile(
+            ctx,
+            source_files = srcs.java,
+            source_jars = generated_src_jars,
```
jongerrish, line 378: As mentioned earlier, maybe we can pass the compiled KAPT stub classes on the classpath rather than the KAPT generated source.

kotlin/internal/jvm/compile.bzl
```diff
@@ -237,81 +267,174 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar, compile_jar):
...
+        action_type = "Runtime",
+        input_jars = output_jars,
     )
 
+    mergee_jdeps = []
+    for java_info in java_infos:
+        if java_info.outputs.jdeps:
+            mergee_jdeps.append(java_info.outputs.jdeps);
+
+#    jdeps = _run_merge_jdeps_action(
```
jongerrish, line 423: Need to figure out what to do about jdeps.

It seems pretty slow, on our fork I've just removed it temporarily, but I see two possibilities
1. Break it out of the Kotlin Builder and extract it to its own Builder and run it on the output jar. We should probably run it on the compile jar also so we can support classpath shortening.
1. Implement a Kotlin compiler plugin to generate the Kotlin half of the jdeps and then write a new builder to merge the two jdeps (I already have the jdeps merger in our fork). Ideally `java_common.merge()` or a new API would support this use case. I can also see about contributing something to Bazel for this.

```
Task timings for //src/test/data/jvm/kapt:ap_kotlin_mixed (total: 1322 ms):
|   * expand sources: 0 ms
|   * kapt (com.google.auto.value.processor.AutoValueProcessor): 546 ms
|   * compile classes: 764 ms
|     * kotlinc: 241 ms
|     * javac: 83 ms
|     * create jar: 45 ms
|     * produce src jar: 6 ms
|     * generate jdeps: 389 ms
```

kotlin/internal/jvm/compile.bzl
```diff
@@ -389,37 +514,34 @@ def _run_kt_builder_action(ctx, rule_kind, toolchains, dirs, srcs, friend, compi
...
             transitive = [compile_deps.compile_jars, transitive_runtime_jars],
         ),
         tools = tools,
         input_manifests = input_manifests,
         outputs = [f for f in outputs.values()],
         executable = toolchains.kt.kotlinbuilder.files_to_run.executable,
-        execution_requirements = {"supports-workers": "1"},
+        execution_requirements = {
+            "supports-workers": "1",
+            "worker-key-mnemonic": "KotlinCompile"
```
jongerrish, line 528: This means `KotlinCompile` and `KotlinKAPT` share the same workers. I'd like to support Multiplexed workers too. It looks like JavaBuilder supports multiplexed workers in Bazel 3.7 and someone submitted maybe a fix for the "multiplexed workers sometimes hang" bug reported by some users which I guess will make it into Bazel 3.8

kotlin/internal/jvm/compile.bzl
```diff
@@ -389,37 +514,34 @@ def _run_kt_builder_action(ctx, rule_kind, toolchains, dirs, srcs, friend, compi
...
+        },
         arguments = [args],
         progress_message = progress_message,
         env = {
             "LC_CTYPE": "en_US.UTF-8",  # For Java source files
         },
     )
 
-def kt_jvm_produce_jar_actions(ctx, rule_kind):
+def kt_jvm_produce_output_jar_actions(ctx, rule_kind):
```
jongerrish, line 537: I should rename this method or maybe just inline or refactor it into the final output jar merge section above.

kotlin/internal/jvm/compile.bzl
```diff
@@ -428,21 +550,18 @@ def kt_jvm_produce_jar_actions(ctx, rule_kind):
...
-        _fold_jars_action(ctx, rule_kind, output_jar, output_merge_list)
+    Returns:
+        see `kt_jvm_compile_action`.
+    """
 
     # Setup the compile action.
     return kt_jvm_compile_action(
         ctx,
-        rule_kind = rule_kind,
-        output_jar = kt_compile_output_jar,
```
jongerrish, line 446: I think I can probably restore this API of passing in both the output and compile jar. This is kind of butchered because our local fork has a version of `kt_android_library` that calls the hidden Skylark APIs for Android resource processing and we needed to do some hijinks with passing in Android SDK + resource jars to the `KotlinBuilder` classpath... I removed all that stuff, but this is the long answer to why I changed this API, but I can change it back :-)

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
```diff
@@ -60,26 +60,29 @@ class KotlinJvmTaskExecutor @Inject internal constructor(
...
-                )
+                  )
+                } else {
+                  emptyList()
+                }
               }
             },
             runCatching {
               context.execute("javac") {
-                if (outputs.jar.isNotEmpty()) {
```
jongerrish, line 82: I think we can remove the Java compiler stuff from the Kotlin Builder but I'm not sure if you really want to keep it for backwards compatibility / safety net / preserve monolithic way as the default and have this PR behind a flag or something..... all this works nicely on Snaps Android codebase so I'm reasonably confident in it.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
```diff
@@ -106,14 +109,24 @@ class KotlinJvmTaskExecutor @Inject internal constructor(
 
         if (outputs.jar.isNotEmpty()) {
           context.execute("create jar", ::createOutputJar)
-          context.execute("produce src jar", ::produceSourceJar)
+          // TODO: Remove - source jar is generated in a separate action
+//          context.execute("produce src jar", ::produceSourceJar)
```
jongerrish, line 113: To be removed, unless you want to preserve the in-process stuff for safety?

kotlin/internal/jvm/compile.bzl
nkoroste: I think this is a good comment to live directly in the source code

kotlin/internal/jvm/compile.bzl
nkoroste: This is a great idea, specially if you think we're going to see a visible performance improvement.

kotlin/internal/jvm/compile.bzl
nkoroste: I think it's important to note that jdeps never actually worked properly for Kotlin. It does't handle things like inline methods for example.

kotlin/internal/jvm/compile.bzl
restingbull: Last I checked kapt compiled stubs aren't particularly valid bytecode. They provide just enough to run the annotation processors. 

Double check by dumping the stub bytecode, of course.

(I would keep them out of the classpath, on the principle that a leak is ugly.)

I *believe* those generated sources are the only output from the annotation processing. I could be mis-remembering how it works.

kotlin/internal/jvm/compile.bzl
restingbull: It is better than nothing, unfortunately. We need a kdeps so that we can get strict deps working.

kotlin/internal/jvm/compile.bzl
restingbull: I'm working on multiplexer support. It has some pitfalls. The way kt workers are intialized is... interesting.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
restingbull: Stepwise. We need to put these changes behind a flag.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
restingbull: Moving the source jar generation seems safe enough.

restingbull (changes requested):
Please move these changes behind a flag of some sort -- we have very poor testing coverage, and these shuffle bytecode a bit.

kotlin/internal/jvm/compile.bzl
jongerrish: KAPT is pretty neat in that it generates compiled stubs for the Kotlin code, which the Java code can then compile against.

Right now we have:-

If there are annotation processors
  - Run KAPT -> generates: kapt-gen-stub.jar (Kotlin Stubs), kapt-gen-java-src.jar (Annotation processor gen-src)
  - In parallel
    Run Kotlinc -> inputs: src/main/**/*.kt (to compile), src/main/**/*.java + kapt-gen-java-src.jar (for Java symbol resolution)
  Run Javac -> inputs src/main/**/*.java + kapt-gen-java-src.jar (to compile), kapt-gen-stub.jar (for Kotlin symbol resolution)

With no annotation processors
  In series run
    - kotlinc -> inputs: src/main/**/*.kt (to compile), src/main/**/*.java (Java symbol resolution) outputs -> output.jar, abi.jar
    - javac -> inputs: inputs src/main/**/*.java (to compile), abi.jar (kotlin symbol resolution)

The idea I had above was for the use case without annotation processors to run KAPT anyway, but with aptMode=stubs which seems fast. It adds an extra action but means that we don't need to wait for full Kotlinc before starting Java compilation. Javac can run in parallel with Kotlinc using the kapt-gen-stubs.jar on its classpath rather than the abi.jar which is probably faster since ABI jar has to go along with a full compile and better caching, i.e: stubs don't include what Java would consider implementation details, e.g: inline method changes, plus, if you're not running in ABI mode then the caching is worse as Java compiles against the full jar rather than the ABI jar.

Let me know if you want me to pursue this change in the PR, I'm quite happy to and can run performance tests against out repo (we have a bunch of dashboards tracking different cases in our repo, ABI changes, non-ABI changes, resource changes etc).

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
jongerrish: Sounds good. I'll do that this weekend.

Note that now Skylark supports build options, how do you feel if I create a Skylark Option for this or prefer an option in the tool chain like you did for experimental_abi_jars?

How about
`experimental_always_use_java_builders`
or
`experimental_new_action_flow`
or.... blanking on good names.

Skylark build flags: see: https://docs.google.com/document/d/1vc8v-kXjvgZOdQdnxPTaV0rrLxtP2XwnD2tAZlYJOqw/edit#heading=h.pn8nbhno42p5

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
jongerrish: Sounds good. I'll keep this change outside of the flag then, but keep the original flow and have this new action flow behind afore mentioned flag. Sound good?

kotlin/internal/jvm/compile.bzl
jongerrish: Actually, I see what `kapt_generated_class_jar` is now, it basically contains anything that should be in the classes folder that isn't classes (since KAPT doesn't compile anything), in practice this means generated resource files, META-INF/services files etc, so we DO need to keep this, but the good thing is it doesn't include classes so won't affect performance.

kotlin/internal/jvm/compile.bzl
jongerrish: I think, if you want to keep the existing in-Kotlin worker Java compilation as the main code path (and this new stuff behind a flag) then probably I should keep the build_java / build_kotlin flags for clarity / simplicity? Otherwise we're doing a lot of if/elif checks on the various inputs to determine which mode to run the builder in.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
jongerrish: Actually, what do you think if I break the source jar generation into a separate PR and remove all traces of source jar packing from the Kotlin builder, might be easier to reason about these as separate changes?

jongerrish: Nevermind, I updated this PR to pull source jar + jdeps generation to their own actions.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
jongerrish: Added an experimental flag to use Java builders.

kotlin/internal/jvm/compile.bzl
restingbull: These directories are a bit of cruft due to poor documentation of tree artifacts. They can be removed, as they don't participate in the compilation.

kotlin/internal/jvm/compile.bzl
restingbull: Let's layer it into another pr -- this one moves a lot around. I'm optimistic -- an extra action should be less painful with multiplexing.

kotlin/internal/jvm/compile.bzl
```diff
@@ -67,6 +67,81 @@ def _fold_jars_action(ctx, rule_kind, output_jar, input_jars, action_type = ""):
...
+    ])
+    args.add("--output", output_jar);
+
+    args.add_all("--resources", ctx.files.srcs);
+
+    ctx.actions.run(
+        mnemonic = "KotlinSrcJar",
+        inputs = ctx.files.srcs,
+        outputs = [output_jar],
+        executable = ctx.executable._singlejar,
```
restingbull, line 90: We can pull singlejar off the java toolchain, or add it to the kotlin one. As long as we don't make a loop, I favor pushing tools to the toolchain.

kotlin/internal/jvm/compile.bzl
restingbull: Agreed -- I did not know that, though it's the kind of shennigans I'd expect from java_common.

kotlin/internal/jvm/compile.bzl
```diff
@@ -237,81 +314,197 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar, compile_jar):
...
-    compile_deps = _compiler_deps(toolchains, friend, deps = ctx.attr.deps + ctx.attr.plugins)
+    compile_deps = _compiler_deps(toolchains, friend, deps = ctx.attr.deps)
     annotation_processors = _plugin_mappers.targets_to_annotation_processors(ctx.attr.plugins + ctx.attr.deps)
     transitive_runtime_jars = _plugin_mappers.targets_to_transitive_runtime_jars(ctx.attr.plugins + ctx.attr.deps)
     plugins = ctx.attr.plugins
 
-    if toolchains.kt.experimental_use_abi_jars:
-        kt_compile_jar = ctx.actions.declare_file(ctx.label.name + "-kt.abi.jar")
+    output_jars = []
+    if toolchains.kt.experimental_use_java_builder:
```
restingbull, line 323: Pull this into a separate function? It's long enough before forking the new way and the old way.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
restingbull: I'm leaning towards toolchain support. The build flags are great to have, but command line options tend to proliferate and be non-discoverable.

restingbull (changes requested):

kotlin/internal/jvm/compile.bzl
jongerrish: This went in a previous PR actually - https://github.com/bazelbuild/rules_kotlin/pull/388 - plugins shouldn't be on the compile classpath anyway.

kotlin/internal/jvm/compile.bzl
jongerrish: Good point, they're internal implementation details of the builder.

Removed.

kotlin/internal/jvm/compile.bzl
jongerrish: Added a comment below when adding `-proc:none` to `javac_opts`

kotlin/internal/jvm/compile.bzl
jongerrish: Inlined / removed this method.

kotlin/internal/jvm/compile.bzl
jongerrish: Just inlined the method.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
jongerrish: Added an option `experimental_use_java_builder` to the tool chain to enable the JavaBuilder flow.

jongerrish:
> Please move these changes behind a flag of some sort -- we have very poor testing coverage, and these shuffle bytecode a bit.

Done.

kotlin/internal/jvm/compile.bzl
jongerrish: Done.

Discovered `java_common.pack_sources()` so using this to create the source jar. For the existing Fold Jars actions getting singlejar from the java toolchain now.

kotlin/internal/jvm/compile.bzl
jongerrish: Done.

restingbull (approved):

PR #380, opened 2020-11-05
Fix the tables in the kotlin documentation

tapthaker:


tapthaker:
If this documentation is autogenerated, please let me know where and we can fix that instead.

restingbull:
template lives over here: https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/doc-templates/rule.vm

render: `bazel build //kotlin: stardoc`

run `scripts/release.sh` when done tweaking to update the md file.

caveat: autoformatting the markup usually breaks it. It's wonderful.

restingbull:
Fixed in master.

PR #383, opened 2020-11-08
Add option rules for kotlinc and javac.

restingbull:


jongerrish (commented):
Perfect, just what we need!

src/main/protobuf/kotlin_model.proto
```diff
@@ -106,28 +106,30 @@ message JvmCompilationTask {
   }
 
   message Inputs {
-    // The full classpath
+      // The full classpath
```
cgruber, line 109: Wierd formatting. Why did the comments indent, but the decl didn't?

cgruber (approved):
This looks fine and straightforward. It's pretty limited, with no arbitrary flags passable, which I assume is the intended design philosophy.  I think this is the right start. We can carefully add more options where reasonable.

There's areas of work and design around use-cases where people want per-target options (very typically annotation processing options), but this is fine to start, and doesn't preclude any further work that I can see.

restingbull:
> This looks fine and straightforward. It's pretty limited, with no arbitrary flags passable, which I assume is the intended design philosophy. I think this is the right start. We can carefully add more options where reasonable.
And you can also enforce options linked to level using https://docs.bazel.build/versions/master/be/common-definitions.html#common.exec_compatible_with. For example, most -X options should be version limited.

> 
> There's areas of work and design around use-cases where people want per-target options (very typically annotation processing options), but this is fine to start, and doesn't preclude any further work that I can see.

Annotation processing is a slippery slope. While java happily allows arbitrary flags whenever, it's not a maintainable approach outside of a mono-repo. I suspect we'd be better off wrapping annotation processor arguments with the plugin rather that attaching them per library. Discovery on that (when removing the processor, for example) is painful.

PR #384, opened 2020-11-10
Support the new Kotlin IR backend

Bencodes:
In order to support Jetpack Compose we need to support enabling the experimental IR backend.

Compose setup examples: https://developer.android.com/jetpack/androidx/releases/compose#declaring_dependencies
Docs for the new IR backend: https://kotlinlang.org/docs/reference/whatsnew14.html#new-jvm-ir-backend

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -342,6 +342,12 @@ def _kotlinc_options_provider_to_flags(opts):
         flags.append("-nowarn")
     if opts.x_use_experimental:
         flags.append("-Xuse-experimental=kotlin.Experimental")
+    if opts.x_use_ir:
+        flags.append("-Xuse-ir")
+    if opts.x_allow_jvm_ir_dependencies:
+        flags.append("-Xallow-jvm-ir-dependencies")
+    if opts.x_skip_prerelease_check:
+        flags.append("-Xskip-prerelease-check")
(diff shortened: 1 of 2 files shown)


Bencodes:
@googlebot I signed it!

kotlin/internal/jvm/compile.bzl
```diff
@@ -342,6 +342,12 @@ def _kotlinc_options_provider_to_flags(opts):
         flags.append("-nowarn")
     if opts.x_use_experimental:
         flags.append("-Xuse-experimental=kotlin.Experimental")
+    if opts.x_use_ir:
+        flags.append("-Xuse-ir")
```
restingbull, line 346: That's it? We don't need extra libs or some such?

Bencodes: The extra libs that are needed can be provided via compiler plugins

restingbull: For IR or compose?

Bencodes: For compose. The new IR backend is just the new compiler backend that Jetbrains has been working on. Jetpack Compose is a compiler plugin that was built to leverage the new backend (enabled via `-Xplugin=...` or `kt_compiler_plugin(...)`.

restingbull (approved):

PR #385, opened 2020-11-11
Several small fixes for previous PR #383

jongerrish:
Fixes required to use `kt_kotlinc_options` and `kt_javac_options`


Bencodes:
Just opened up the same pull request here https://github.com/bazelbuild/rules_kotlin/pull/386 but also exposed `kt_kotlinc_options` and `kt_javac_options` so that we can override the default opts.

restingbull (approved):

PR #388, opened 2020-11-12
Fix kt_compiler_plugin not working with android_binary

jongerrish:
`kt_compiler_plugin` currently doesn't work when used from anywhere in the deps of an `android-binary` target, failing as follows:-

```
ERROR: /Users/jgerrish/indiekid/rules_kotlin/examples/android/app/BUILD.bazel:29:15: in deps attribute of android_binary rule //app:app3: Dependencies on .jar artifacts are not allowed in Android binaries, please use a java_import to depend on kotlin_rules_maven/v1/https/jcenter.bintray.com/org/jetbrains/kotlin/kotlin-serialization-unshaded/1.4.10/kotlin-serialization-unshaded-1.4.10.jar. If this is an implicit dependency then the rule that introduces it will need to be fixed to account for it correctly.. Since this rule was created by the macro 'android_binary', the error might have been caused by the macro implementation
ERROR: Analysis of target '//app:force_build_apks_test' failed; build aborted: Analysis of target '//app:app3' failed
```

The fix is not including plugins as part of the compile deps.

Add `kt_compiler_plugin` to the `kt_android_library` target referenced by an `android_binary` update Activity to reference generated code to validate it was run correctly.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -236,7 +236,7 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar, compile_jar):
     dirs = _compiler_directories(ctx)
     srcs = _partitioned_srcs(ctx.files.srcs)
     friend = _compiler_friends(ctx, friends = getattr(ctx.attr, "friends", []))
-    compile_deps = _compiler_deps(toolchains, friend, deps = ctx.attr.deps + ctx.attr.plugins)
+    compile_deps = _compiler_deps(toolchains, friend, deps = ctx.attr.deps)
     annotation_processors = _plugin_mappers.targets_to_annotation_processors(ctx.attr.plugins + ctx.attr.deps)
     transitive_runtime_jars = _plugin_mappers.targets_to_transitive_runtime_jars(ctx.attr.plugins + ctx.attr.deps)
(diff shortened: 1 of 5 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -236,7 +236,7 @@ def kt_jvm_compile_action(ctx, rule_kind, output_jar, compile_jar):
     dirs = _compiler_directories(ctx)
     srcs = _partitioned_srcs(ctx.files.srcs)
     friend = _compiler_friends(ctx, friends = getattr(ctx.attr, "friends", []))
-    compile_deps = _compiler_deps(toolchains, friend, deps = ctx.attr.deps + ctx.attr.plugins)
+    compile_deps = _compiler_deps(toolchains, friend, deps = ctx.attr.deps)
```
restingbull, line 239: Did this sneak in? Or is it part of the fix?

restingbull (changes requested):

kotlin/internal/jvm/compile.bzl
jongerrish: That's the fix. Including plugins as compile deps produces the error in the PR description

restingbull (approved):

PR #381, opened 2020-11-05
Support defining extra compiler arguments in the tool chain.

jongerrish:
Supports the following attributes in the toolchain:-

`extra_copts` - supports passing extra arguments to kotlinc
`jvm_opts` - supports passing extra arguments to the Java compiler + KAPT

PR #390, opened 2020-11-12
Expose -Werror as a compiler option

Kernald:
The user-facing option name (`all_warnings_as_errors`) matches Gradle's equivalent option (`AllWarningsAsErrors`). Not sure if that makes lots of sense, and I'm happy to change it for a better suggestion - `werror` or `w_error` don't sound exactly great.

This PR does this for both javac and kotlinc.

### DIFF
--- a/kotlin/internal/opts.bzl
+++ b/kotlin/internal/opts.bzl
@@ -15,10 +15,11 @@
 _KOPTS = {
     "warn": struct(
         args = dict(
-            default = True,
-            doc = "Enable or disable compiler warnings.",
+            default = "report",
+            doc = "Control warning behaviour.",
+            values = ["off", "report", "error"],
         ),
-        type = attr.bool,
+        type = attr.string,
     ),
     "x_use_experimental": struct(
         args = dict(
@@ -56,10 +57,11 @@ kt_kotlinc_options = rule(
 _JOPTS = {
     "warn": struct(
         args = dict(
-            default = True,
-            doc = "Enable or disable compiler warnings.",
+            default = "report",
+            doc = "Control warning behaviour.",
+            values = ["off", "report", "error"],
         ),
-        type = attr.bool,
+        type = attr.string,
     ),
     "x_ep_disable_all_checks": struct(
         args = dict(
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -338,8 +338,10 @@ def _kotlinc_options_provider_to_flags(opts):
     if not opts:
         return ""
     flags = []
-    if not opts.warn:
+    if opts.warn == "off":
         flags.append("-nowarn")
+    elif opts.warn == "error":
+        flags.append("-Werror")
     if opts.x_use_experimental:
         flags.append("-Xuse-experimental=kotlin.Experimental")
     return flags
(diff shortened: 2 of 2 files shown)


Bencodes:
This is another flag that we require in our codebase so +1 for merging this!

artem-zinnatullin:
Honestly, I don't quite understand why don't we just allow free compiler args instead of adding each possible arg as starlark attribute, did I miss some discussion?

Bencodes:
The discussion happened here https://github.com/bazelbuild/rules_kotlin/pull/381

kotlin/internal/opts.bzl
```diff
@@ -13,6 +13,13 @@
 # limitations under the License.
 
 _KOPTS = {
+    "all_warnings_as_errors": struct(
```
jongerrish, line 16: You might want to keep the consistent naming pattern used above, so this would be `w_error`

kotlin/internal/opts.bzl
```diff
@@ -54,6 +61,13 @@ kt_kotlinc_options = rule(
 )
 
 _JOPTS = {
+    "all_warnings_as_errors": struct(
```
jongerrish, line 64: `w_error`?

kotlin/internal/opts.bzl
Kernald: I have to say, I'm not sure I get the point of this abstraction if it doesn't provide a higher-level API over the compiler configuration (the fact that they directly map to flags isn't especially relevant IMHO). I'm happy to rename it for the sake of consistency, but I don't think it's the best approach here. As it is, you would have this configuration:

```
kt_kotlinc_options(
  name = "toolchain_options",
  all_warnings_as_errors = True
)
```

vs:

```
kt_kotlinc_options(
  name = "toolchain_options",
  w_error = True
)
```

I personally find the first one clearer and more obvious.

kotlin/internal/opts.bzl
restingbull: I would update "warn" to be a trinary: "off", "report", "error", eg:

```
    "warn": struct(
        args = dict(
            default = "report",
            doc = "Control warning behaviour.",
            values = ["off", "report", "error"],
        ),
        type = attr.string,
    ),
```

Kernald: That's a really good idea, I'll update this later today.

keith: it seems like we're barreling towards reproducing all compiler flags and their options in this dictionary. Are we sure it doesn't make sense to allow arbitrary flags instead? Seeing a few more PRs open adding more of these as well

restingbull: The dictionary is merely a code reuse tool. The flags are typesafe, and as suggested above, can provide a richer api.

Having maintained and worked on rules that allow arbitrary string lists (or worse, just a string) the long term maintenance outweighs the expediency. There are plans to restructure the compiler calls to use the embeddable api, which means reparsing them in the worker. Finally, removal of 

From a code repository maintenance perspective, abstracting the flags allows:
* Defining flagset options and constraints (visibility, environment, etc.) them (makes large scale code changes a lot easier)
* Feedback for incorrect values happens in analysis time, as well as improved discovery as the compiler options are embedded with the rules.
* Auditing flag usage.
* Deprecations are immediately clear via warnings from the rules.

The initial commit was address the need to add flags in a  maintainable manner. I expect there will be a lot of prs adding to the options for a bit as the most used options are added.

Kernald: Thanks for all the details! I updated this to use a trinary, for both `javac` and `kotlinc`.

restingbull (approved):

cgruber:
> Honestly, I don't quite understand why don't we just allow free compiler args instead of adding each possible arg as starlark attribute, did I miss some discussion?

You did miss discussions with Syed Hassan in the very early days, and some side discussions in which we were very hesitant to open up the capability arbitrarily without more concentrated design effort. But your issue #396 can serve as a platform for that design discussion.

PR #393, opened 2020-11-16
Restingbull/compiler plugin rework

restingbull:
Merging fixes for the compiler plugins. (long overdue)
Add small optimization to skip building without sources.
First pass to sort out export vs. runtime vs neverlink.
Update examples to cover more cases.


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/compilation_task.kt
```diff
@@ -97,17 +114,68 @@ internal fun JvmCompilationTask.produceSourceJar() {
   }
 }
 
+internal fun encodeMap(options: Map<String, String>): String {
```
jongerrish, line 117: FYI - there is a CLI interface to KAPT now which handles the encoding of parameters `org.jetbrains.kotlin.kapt.cli`

examples/android/lib/BUILD.bazel
```diff
@@ -1,7 +1,10 @@
-load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_compiler_plugin", "kt_android_library")
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kt_android_library", "kt_compiler_plugin")
 
 kt_compiler_plugin(
     name = "serialization_plugin",
+    compile_phase = True,
+    id = "org.jetbrains.kotlin.serialization",
+    stubs_phase = True,
```
jongerrish, line 7: Does this mean the plugin and the code it generates is run twice, once during KAPT and once during regular kotlinc?

restingbull: Yep. If you don't, the stubs can be bad.

examples/android/lib/BUILD.bazel
```diff
@@ -12,12 +15,12 @@ kt_android_library(
...
+        ":serialization_plugin",
+    ],
     visibility = ["//visibility:public"],
     deps = [
-        "@maven//:org_jetbrains_kotlinx_kotlinx_serialization_runtime",
         "@maven//:androidx_appcompat_appcompat",
-    ],
-    plugins = [
-        ":serialization_plugin",
+        "@maven//:org_jetbrains_kotlinx_kotlinx_serialization_runtime",
```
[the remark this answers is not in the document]
restingbull: Huh. This isn't what I have locally.

examples/android/lib/BUILD.bazel
cgruber: So, after discussion, I think we can default these to true, because most cases it's what folks will want. Badly behaved plugins may trip this up, but most users of compiler plugins likely won't understand the dynamics, and all of the existing corpus of plugins seems to be safe for these as a default.

examples/plugin/src/noarg/BUILD
```diff
@@ -2,10 +2,12 @@ load("//kotlin:kotlin.bzl", "kt_compiler_plugin", "kt_jvm_library", "kt_jvm_test
 
 kt_compiler_plugin(
     name = "no_arg_plugin",
+    compile_phase = True,
```
cgruber, line 5: Setting these to default=true makes even more sense when you see all of these additions. It is less explicit than ex-googlers want to be, but most people in the wild are going to be naive users of published plugins, not necessarily build experts.

examples/plugin/src/serialization/empty.kt
```diff
@@ -0,0 +1,4 @@
+
+package plugin.serialization
+
+// Exists to satisfy the "must have a source file" for kt_jvm_library, which will go away eventually
```
cgruber, line 4: Didn't you fix "must have a source?"

src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
```diff
@@ -45,7 +45,12 @@ class KotlinToolchain private constructor(
     "org.jetbrains.kotlin.kapt3"
   ),
   val jvmAbiGen: CompilerPlugin,
-  val skipCodeGen: CompilerPlugin
+  val skipCodeGen: CompilerPlugin,
+  val extensionsCompiler: CompilerPlugin = CompilerPlugin(
+    kotlinHome.resolveVerified(
+      "lib", "android-extensions-compiler.jar").absolutePath,
```
cgruber, line 51: o.O? Formatter is weird.

cgruber (changes requested):
A few nits and issues, but nearly there.

cgruber:
Probably want to rebase from the changes from the release branch.

.bazelignore
```diff
@@ -1,6 +1,7 @@
 # Some examples have their own WORKSPACE. Ignore as part of this root WORKSPACE so
 # we don't break trying to build separate workspaces using wildcards like //...
 # examples/dagger doesn't have its own workspace, so don't do all of examples.
-examples/android
-examples/node
-examples/trivial
+examples / android
```
cgruber, line 4: Did your linter apply starlark linting to a path?

scripts/reflow_skylark
```diff
@@ -17,5 +17,6 @@
 
 buildifier $(find . -type f \
     -iname "*.bzl" -or \
-    -name "BUILD*"
-)
\ No newline at end of file
+    -name "*.bazel" -or \
```
cgruber, line 20: I think this is what's killing your .bazelignore and making buildifier edit it.

cgruber (approved):
Just the formatting issue for .bazelignore, and the commented find configuration for reflow.

scripts/reflow_skylark
restingbull: Nope.

.bazelignore
restingbull: No -- I think it's intellij.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/compilation_task.kt
restingbull: Pick that up shortly -- though it would be a breaking change for older plugin versions.

examples/android/third_party/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
+java_plugin(
+    name = "dagger_component_plugin",
+    generates_api = 1,
```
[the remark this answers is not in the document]
restingbull: Cargo culting. I think we use this in the dagger example.

.bazelignore
restingbull: Nope. Bad githook grep.

PR #395, opened 2020-11-16
Support -no-stdlib and -no-reflect kotlinc args

Bencodes:
https://kotlinlang.org/docs/reference/compiler-reference.html#-no-reflect and https://kotlinlang.org/docs/reference/compiler-reference.html#-no-stdlib

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -205,6 +205,10 @@ def _kotlinc_options_provider_to_flags(opts):
         flags.append("-Xallow-jvm-ir-dependencies")
     if opts.x_skip_prerelease_check:
         flags.append("-Xskip-prerelease-check")
+    if opts.include_stdlibs == "stdlib":
+        flags.append("-no-reflect")
+    elif opts.include_stdlibs == "none":
+        flags.append("-no-stdlib")
     return flags
 
 def _javac_options_provider_to_flags(opts):
--- a/kotlin/internal/opts.bzl
+++ b/kotlin/internal/opts.bzl
@@ -21,6 +21,14 @@ _KOPTS = {
         ),
         type = attr.string,
     ),
+    "include_stdlibs": struct(
+        args = dict(
+            default = "all",
+            doc = "Don't automatically include the Kotlin standard libraries into the classpath (stdlib and reflect).",
+            values = ["all", "stdlib", "none"],
+        ),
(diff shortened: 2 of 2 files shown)


Kernald:
In the same idea as https://github.com/bazelbuild/rules_kotlin/pull/390#discussion_r524530762, I would expose these two flags in a single one with multiple options rather than two booleans, given that `-no-stdlib` implies `no-reflect`. Maybe a `include_stdlibs` with `all` (default), `stdlib` (`-no-reflect`), `none` (`-no-stdlib`)?

Bencodes:
Updated to use string options with `all`, `stdlib`, and `none`.

kotlin/internal/jvm/compile.bzl
```diff
@@ -205,6 +205,10 @@ def _kotlinc_options_provider_to_flags(opts):
         flags.append("-Xallow-jvm-ir-dependencies")
     if opts.x_skip_prerelease_check:
         flags.append("-Xskip-prerelease-check")
+    if opts.include_stdlibs == "stdlib":
+        flags.append("-no-reflect")
```
restingbull, line 209: Doesn't reflect the include_stdlibs attribute.

restingbull (changes requested):

kotlin/internal/jvm/compile.bzl
Bencodes: @restingbull mind elaborating?

Bencodes: The `-no-reflect` flag strips out the Kotlin reflect library but keeps the standard library. The `-no-stdlib` flag strips both the reflect and standard libraries out. Not passing either of these flags will result in both the standard library and reflect being on the classpath.

When the `include_stdlibs ` is set to `all` we want to just include everything (pass no flags). When `include_stdlibs` is set to `stdlib` we want to pass -no-reflect which just strips the reflect standard libraries but keeps the kotlin standard library. Lastly when we pass `none` we want to just strip everything by passing `-no-stdlib`.

This logic also mirrors what @Kernald mentioned here https://github.com/bazelbuild/rules_kotlin/pull/395#issuecomment-728301116

restingbull: Nevermind, I'm going brain dead. I read the code as taking a list no less than 5 times. My apologies.

restingbull (approved):

PR #398, opened 2020-11-17
Add opt-in flags

jeffzoch:


alexjlockwood:
@jeffzoch FYI your code isn't compiling right now, I think you have a small syntax error.

alexjlockwood:
This is required for us to get Jetpack Compose compiling using Bazel in our codebase. It would be great if it could be reviewed!

jeffzoch:
@alexjlockwood I need to actually fix a bug i just found yesterday in this PR: the separator should be `=` instead of `:`. Ill push a fix in a bit. Also hoping it could be reviewed

kotlin/internal/opts.bzl
```diff
@@ -49,6 +49,13 @@ _KOPTS = {
         ),
         type = attr.bool,
     ),
+    "x_opt_in": struct(
+        args = dict(
+            default = [],
+            doc = "opt in to experimental features",
```
restingbull, line 55: Hmmmm... https://kotlinlang.org/docs/reference/opt-in-requirements.html

This looks like it's going to be a user definable configuration over time. Could you open an issue for deciding the direction of the opt in api? 

If there are specific usecases you need to unblock, we can work around the larger design considerations for now.

restingbull (changes requested):

kotlin/internal/opts.bzl
jeffzoch: Good point - it IS possible for library writers to define something needs to be opt in. That + the need for experimental coroutine apis (which are sadly still very common) are why I figured it be possible to define this. Ill open the issue but would love for this to be added soon as a lot of APIs are unavailable for use until this flag is here.

restingbull (approved):

restingbull (changes requested):
Please apply changes discussed in https://github.com/bazelbuild/rules_kotlin/issues/422.

jeffzoch:
@restingbull I went ahead and took a stab at it - had a few questions ill post on the PR

examples/trivial/app/BUILD.bazel
```diff
@@ -7,7 +7,10 @@ kt_jvm_library(
         "//:java_deps",
     ],
 )
-
+kt_kotlinc_options(
```
jeffzoch, line 10: in #422 you proposed 
```
kt_compiler_options(
   opt_in = [
        "boo.bar.Baz"
   ]
)
```
but there is no kt_compiler_options construct currently. It felt strange to introduce that rather than just use kotlinc_options so I added it there

kotlin/internal/opts.bzl
```diff
@@ -86,6 +86,13 @@ _KOPTS = {
         ),
         type = attr.bool,
     ),
+    "opt_in": struct(
```
jeffzoch, line 89: Should this be in _KOPTS? It also felt weird to add it to a structure that is essentially identical but I suppose the important distinction is that toolchain overrides do not apply to this (based on compile.bzl)

examples/trivial/app/BUILD.bazel
restingbull: 🤦  I meant the kotlinc options. The naming is new.

kotlin/internal/opts.bzl
restingbull: This is sort of the tricky bit -- we only have one options rule for the moment. I'm inclined to split it out into two rules (lib and toolchain). What are your thoughts? 

What I'd like to avoid is the toolchain having optins that are ignored or error -- the idea is to make the distinction clear.

I'm also willing to just let optins be on the toolchain.

What are your thoughts?

restingbull (changes requested):

kotlin/internal/opts.bzl
jeffzoch: @restingbull i think two distinct ones sounds fine to me - would we duplicate the flags in each that can be overridden or would is there someway to embed all toolchain flags into rule flags (if thats the desired behavior -not aware of toolchain flags that shouldnt be overriden by rules)?

jeffzoch: @restingbull just wanted to bump this as im not sure if you saw or not - its been a bit. Any thoughts on duplicating all flags vs embedding toolchain to be a strict superset or rule flags? The overlap is enticing to mitigate but may not be worthwhile

PR #400, opened 2020-11-19
Upgrade rules_nodejs to 1.7.0

meteorcloudy:
rules_nodejs pre 1.6.0 had invalid escape sequences in its Starlark code, which was silently ignored but now throws an error (due to bazelbuild/bazel@73402fa).


restingbull (approved):
Well. 0.36.1 -> 1.7.0 is scary.

But the tests pass, so let's find out.

PR #408, opened 2020-11-23
Java builder flow fixes

jongerrish:
1. experimental_use_java_builder flow is broken at head. Fix breakages, convert examples/android project to use custom toolchain with both experimental_use_java_builder + experimental_use_abi_jars to validate this flow.

2. Java source version must be passed to KAPT so that generated source matches the target. When targeting Java 8 but omitting the source value, incorrectly Dagger generates @javax.annotation.processing.Generated when it actually needs to generate @javax.annotation.Generated.

3. ABI classes should be written to a separate output folder to regular compilation output so that output in the compile jar is not polluted by non-ABI classes.

### DIFF
--- a/examples/android/WORKSPACE
+++ b/examples/android/WORKSPACE
@@ -64,7 +64,7 @@ load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kotlin_repositories", "kt_reg
 
 kotlin_repositories()
 
-kt_register_toolchains()
+register_toolchains("//bzl:experimental_toolchain")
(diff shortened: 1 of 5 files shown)


restingbull:
Appears to be broken in the release.

restingbull (approved):

cgruber:
The merged PR is failing in the android example. @jongerrish can you peek at it and see if it's related?

jongerrish:
I'll take a look as soon as I'm back at my desk.

PR #415, opened 2020-12-03
Merge Kotlin jdeps with Java jdeps produced from Starlark APIs

jongerrish:
Create a Jdeps merger action to merge Kotlin Builder produced Jdeps
with jdeps produced from `java_common.compile()`


cgruber:
Super initial thought.

So, I don't have a strong opinion on moving jdeps production into the builder. I don't recall if @restingbull pulled it out (and therefore has opinions) or if it was out from a prior decision.  I haven't paid much attention to the jdeps stuff. That said, I'm very excited about parts of this - notably the prospect of both strict deps and unused_deps, and I can see how this sets the stage for that.

jongerrish:
> Super initial thought.
> 
> So, I don't have a strong opinion on moving jdeps production into the builder. I don't recall if @restingbull pulled it out (and therefore has opinions) or if it was out from a prior decision. I haven't paid much attention to the jdeps stuff. That said, I'm very excited about parts of this - notably the prospect of both strict deps and unused_deps, and I can see how this sets the stage for that.

Actually I pulled it out and then regretted it so want to put it back hah hah ;-P

src/main/kotlin/io/bazel/kotlin/builder/JdepsMergerComponent.java
```diff
@@ -0,0 +1,55 @@
...
+  @Component.Builder
+  interface Builder {
+    JdepsMergerComponent build();
+  }
+
+  @dagger.Module
+  class Module {
+
+    @Provides
+    public PrintStream provideDebugPrintStream() {
```
cgruber, line 41: Super-nit: Technically these can all be static methods and the module declared abstract. This avoids creating an instance.

cgruber (approved):
This seems pretty reasonable. I'll let @restingbull comment in case there's something I missed, but it seems straightforward.

kotlin/internal/jvm/impl.bzl
```diff
@@ -34,7 +34,7 @@ def _make_providers(ctx, providers, transitive_files = depset(order = "default")
             providers.java,
             providers.kt,
             DefaultInfo(
-                files = depset([ctx.outputs.jar]),
+                files = depset([ctx.outputs.jar, ctx.outputs.jdeps]),
```
restingbull, line 37: Does this need to be in the default info for the IDE? I'd prefer to leave it as a "build on request" artifact. Also should it go into JavaInfo?

kotlin/internal/jvm/compile.bzl
```diff
@@ -348,7 +341,7 @@ def _run_kt_builder_action(
 
     args.add_all("--kotlin_passthrough_flags", _kotlinc_options_provider_to_flags(toolchains.kt.kotlinc_options))
     args.add_all("--javacopts", _javac_options_provider_to_flags(toolchains.kt.javac_options))
-
+    args.add_all("--direct_dependencies", _java_infos_to_compile_jars(compile_deps.deps))
```
restingbull, line 344: Open an issue and todo this one -- it isn't used in the builder.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
```diff
@@ -0,0 +1,88 @@
...
+import java.nio.file.FileSystems
+import java.nio.file.Files
+import java.nio.file.Path
+import java.nio.file.Paths
+
+/**
+ * Persistent worker capable command line program for merging multiple Jdeps files into a single
+ * file.
+ */
+class JdepsMerger: CommandLineProgram {
```
restingbull, line 21: Test?

restingbull (changes requested):

src/main/kotlin/io/bazel/kotlin/builder/JdepsMergerComponent.java
jongerrish: Done

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
jongerrish: Done

kotlin/internal/jvm/compile.bzl
jongerrish: Done

kotlin/internal/jvm/impl.bzl
jongerrish: Done - for some reason I couldn't get it working without this, but I just removed it... and its working :-)

restingbull (approved):

PR #416, opened 2020-12-03
Add anvil example to CI, and fix release artifact so jarjar works for un-shaded compiler uses in plugins.

cgruber:
- Add the anvil example to the CI.
  - Explicitly declare the test class for anvil's test, since src/test/java isn't inferred for kotlin tests.
  - Add in the dependencies of jarjar, so compiler de-shading works in the release package.
  - Temporarily disable `--override_repository` until #417 is figured out.


cgruber:
This PR as of this comment fixes the release script's exercize of the anvil example, but doesn't fix the CI version. Looking at it.

restingbull (approved):

PR #418, opened 2020-12-03
Add support for target level compiler opt

Bencodes:
Adds support for target-level control over compiler flags. Example usage:

```kotlin
# Provided to the toolchain as the default
kt_kotlinc_options(name = "kt_kotlinc_options", x_use_ir = False)

define_kt_toolchain(
    name = "kotlin_toolchain",
    ...
    kotlinc_options = "//:kt_kotlinc_options",
)

# Define a custom toolchain that uses the IR backend
kt_kotlinc_options(name = "kt_kotlinc_options_compose", x_use_ir = True)

kt_android_library(
    name = "compose_demo",
    ...
    kotlinc_options = "//:kt_kotlinc_options_compose",
)
```

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -25,6 +25,11 @@ load(
     _plugin_mappers = "mappers",
 )
 load(
+    "//kotlin/internal:opts.bzl",
+    _KotlincOptions = "KotlincOptions",
+    _JavacOptions = "JavacOptions",
+)
+load(
     "//kotlin/internal:compiler_plugins.bzl",
     _plugins_to_classpaths = "plugins_to_classpaths",
     _plugins_to_options = "plugins_to_options",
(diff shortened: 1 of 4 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -346,8 +351,12 @@ def _run_kt_builder_action(
     for f, path in outputs.items():
         args.add("--" + f, path)
 
-    args.add_all("--kotlin_passthrough_flags", _kotlinc_options_provider_to_flags(toolchains.kt.kotlinc_options))
-    args.add_all("--javacopts", _javac_options_provider_to_flags(toolchains.kt.javac_options))
+    # Unwrap kotlinc_opts/javac_options options or default to the ones being provided by the toolchain
+    kotlinc_options = ctx.attr.kotlinc_opts[_KotlincOptions] if ctx.attr.kotlinc_opts else toolchains.kt.kotlinc_options
```
restingbull, line 355: We should make an explicit decision about how the merging strategy should go. I'm not opposed to rule prioritized over toolchain, but it needs documentation. And possibly an RFC. @cgruber, your thoughts?

Bencodes: We looked into merging the opts before passing them to the builder but landed on just prioritizing the rule over toolchain because it makes opts consumption more explicit and has room for hidden side effects.

restingbull: I consider prioritizing the rule over toolchain a merging strategy, and would need to be documented as such. It is the most intuitive approach.

restingbull (changes requested):
Add documentation for merging (e.g. rule options are prioritized over toolchain.)

Bencodes:
Adding some docs to the readme.

Bencodes:
@restingbull any thoughts on the documentation here?

restingbull (approved):

restingbull:
Apologies for the quiet.

PR #424, opened 2020-12-08
Make values passed to --direct_dependencies unique.

jongerrish:
(Validated the parameter is identical to that used by JavaBuilder plus
the additional Kotlin compiler jar dep required to build Java)

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -114,7 +114,7 @@ def _jvm_deps(toolchains, friend, deps, runtime_deps = []):
     )
 
 def _java_infos_to_compile_jars(java_infos):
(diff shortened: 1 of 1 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -114,7 +114,7 @@ def _jvm_deps(toolchains, friend, deps, runtime_deps = []):
     )
 
 def _java_infos_to_compile_jars(java_infos):
-    return [compile_jar.path for java_info in java_infos for compile_jar in java_info.compile_jars.to_list()]
```
restingbull, line 117: I'm falling down the job -- should have called out the depset in the first pass.

restingbull (approved):

PR #433, opened 2020-12-28
Remove hard coding "io_bazel_rules_kotlin" to allow alternative names.

restingbull:
Surprisingly complicated, but there were quite a few instances where the repository name had to be "io_bazel_rules_kotlin"

Aside from being a better behaved rule set, this opens the door to build rules_kotlin using rules_kotlin.


kotlin/internal/jvm/impl.bzl
```diff
@@ -230,6 +233,8 @@ def kt_jvm_junit_test_impl(ctx):
             transitive = [runtime_jars],
             direct = ctx.files._java_runtime,
         ),
+        # adds common test variables, including TEST_WORKSPACE.
+        testing.TestEnvironment({}),
```
cgruber, line 237: Holy crap, we didn't have this? Urgh.... :(

kotlin/internal/jvm/jvm.bzl
```diff
@@ -513,13 +513,13 @@ kt_compiler_plugin = rule(
             default = False,
         ),
         "_jetbrains_deshade_rules": attr.label(
-            default = "@io_bazel_rules_kotlin//kotlin/internal/jvm:jetbrains-deshade.jarjar",
+            default = "//kotlin/internal/jvm:jetbrains-deshade.jarjar",
```
cgruber, line 516: Sometimes here you're using a plain string `"//third_party/foo"`, sometimes a `Label("//third_party/foo")`.  Should we be consistent, or is this a place where that's actually advisable or necessary?

kotlin/internal/repositories/release_repositories.bzl
```diff
@@ -55,3 +60,46 @@ def kotlin_repositories(compiler_release = KOTLIN_CURRENT_COMPILER_RELEASE):
...
+    )
+    repository_ctx.file(
+        "WORKSPACE",
+        content = """workspace(name = "%s")""" % attr.name,
+    )
+    repository_ctx.template(
+        "BUILD.bazel",
+        attr._template,
+        substitutions = {
+            "{{.KotlinRules}}": attr.kotlin_rules,
```
cgruber, line 81: This is one of the grossest things I've seen, but I get it. Haven't used repository_ctx.template() much myself. (gross in convention, not in concept - it's normal templating in concept)

third_party/BUILD
```diff
@@ -86,13 +86,9 @@ release_archive(
...
         "empty.jdeps",
         "jarjar.bzl",
-        "@kotlin_rules_maven//:org_pantsbuild_jarjar",
-        "@kotlin_rules_maven//:org_ow2_asm_asm_tree",
-        "@kotlin_rules_maven//:org_ow2_asm_asm",
-        "@kotlin_rules_maven//:org_ow2_asm_asm_commons",
-        "@kotlin_rules_maven//:org_ow2_asm_asm_analysis",
     ],
     src_map = {
+        "jarjar_runner_deploy.jar": "jarjar.jar",
```
cgruber, line 91: Just to clarify - you're yoinking the deploy jar and releasing that?

cgruber (approved):
Minor consistency nits, and a few questions out of curiosity, but nothing blocking. It's generally quite straightforward. LGTM

kotlin/internal/repositories/release_repositories.bzl
restingbull: I'm honestly not sure how else to do it.

kotlin/internal/jvm/jvm.bzl
restingbull: Good catch. We don't use compiler plugins in base, so I missed it.

We should *always use Label*. Label provides the fully qualified target, where as string is resolved relatively against the main repo.

third_party/BUILD
restingbull: Yes. No fuss, no muss.

jongerrish (approved):

PR #434, opened 2020-12-28
Fix bazel kapt to support generating non-compilable `source` files

ThomasCJY:
**Background**
`kt_xxx_library` doesn't support annotation processor generating source files other than `java` and `kt` classes. For example, if we are trying to use a plugin similar to [auto service](https://github.com/google/auto/tree/master/service) but changing the output option to be [StandardLocation.SOURCE_OUTPUT](https://github.com/google/auto/blob/afe607c395bc34efe56ac34bd3613c3d9a04d0eb/service/processor/src/main/java/com/google/auto/service/processor/AutoServiceProcessor.java#L173), the compile will fail and report `IllegalStateException: invalid source file type xxx`. The reason is that we assume all the generated `source` files to be compilable, which is not always the case.

**Change**
* Keep `generated_java_srcjar` (which contains all kapt generated files under StandardLocation.SOURCE_OUTPUT) as an output in both `experimental_use_abi_jars` and normal mode.
* Filter out non-compilable files before running `expandWithSources`

Note: `android_library` doesn't have this issue and it always put generated files under `xx-gensrc.jar`. Therefore, in this PR, I'm unifying the naming for output to both end with `-gensrc.jar`

**Test**
* Tested on local repo, both normal mode and
 `abit_jar` mode can generate `-gensrc.jar` target which contains the generated files

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -508,6 +508,7 @@ def kt_jvm_produce_jar_actions(ctx, rule_kind):
 
     else:
         kt_java_output_jar = ctx.actions.declare_file(ctx.label.name + "-kt-java.jar")
+        kt_generated_java_srcjar = ctx.actions.declare_file(ctx.label.name + "-gensrc.jar")
         _run_kt_builder_action(
             ctx = ctx,
             rule_kind = rule_kind,
@@ -522,6 +523,7 @@ def kt_jvm_produce_jar_actions(ctx, rule_kind):
             outputs = {
                 "output": kt_java_output_jar,
                 "kotlin_output_jdeps": ctx.outputs.jdeps,
+                "generated_java_srcjar": kt_generated_java_srcjar,
             },
         )
         compile_jar = kt_java_output_jar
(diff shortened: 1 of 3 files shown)


restingbull:
What is the java_library functionality for generated sources?

ThomasCJY:
@restingbull The `-generated-src.jar` file being used right now is behind a flag `experimental_use_abi_jars`. Here I'm renaming it to be as the same format as `android_library`, which put output under `-gensrc.jar`. I think it will be better if we align kotlin rule's output file name pattern with `android_library` rule when it's not being widely used by the community.   cc @jongerrish

jongerrish:
> @restingbull The `-generated-src.jar` file being used right now is behind a flag `experimental_use_abi_jars`. Here I'm renaming it to be as the same format as `android_library`, which put output under `-gensrc.jar`. I think it will be better if we align kotlin rule's output file name pattern with `android_library` rule when it's not being widely used by the community. cc @jongerrish

I think its fine to change the output artifact name since this work is behind an experimental flag and consistency with {java,android}_library is nice.

restingbull:
> > @restingbull The `-generated-src.jar` file being used right now is behind a flag `experimental_use_abi_jars`. Here I'm renaming it to be as the same format as `android_library`, which put output under `-gensrc.jar`. I think it will be better if we align kotlin rule's output file name pattern with `android_library` rule when it's not being widely used by the community. cc @jongerrish
> 
> I think its fine to change the output artifact name since this work is behind an experimental flag and consistency with {java,android}_library is nice.

Ah, gotcha -- I've lost track of exactly what the lives in the java builder path. I'd like to make the default, but until we've got a work around the abi issues, we are stuck.

Issue #423, opened 2020-12-08
Jdeps report unused deps when referencing a constant.


jongerrish:
When depending only on a constant value in another module jdeps reports that module as UNUSED.

I suspect this is because it is inlined by the compiler but haven't checked the bytecode yet.

// Foo module
fun foo() {
  val barRef = Bar.someNotConst // Reports EXPLICIT
  val bazRef = Baz.someConst // Reports UNUSED
}

// Bar module
object Bar {
   val someNotConst = "Not a const"
}

// Baz Module
object Baz {
   val someConst = "Constant Val"
}

PR #435, opened 2020-12-29
Rework Jdeps

jongerrish:
* Create Kotlin compiler plugin to generate Bazel Jdeps output.
* Existing dependency analysis (post process output jar file) now only occurs on Java compilation performed in the KotlinBuilder.
* Kotlin Jdeps now merged with Java Jdeps in all cases.
* Direct dependencies instead of full classpath now passed in to both Kotlin and Java Jdeps generators, jdeps files should now be a lot smaller.

Future Work:- Implement Kotlin strict deps inside the compiler plugin, now we can tell if a symbol is loaded from a jar contained in direct dependencies or not.

Fixes: https://github.com/bazelbuild/rules_kotlin/issues/423

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
@@ -45,7 +45,8 @@ class KotlinToolchain private constructor(
     "org.jetbrains.kotlin.kapt3"
   ),
   val jvmAbiGen: CompilerPlugin,
-  val skipCodeGen: CompilerPlugin
+  val skipCodeGen: CompilerPlugin,
+  val jdepsGen: CompilerPlugin
 ) {
 
   companion object {
@@ -69,6 +70,11 @@ class KotlinToolchain private constructor(
       RULES_REPOSITORY_NAME,
       "src", "main", "kotlin",
       "skip-code-gen.jar").toPath()
+
+    private val JDEPS_GEN_PLUGIN = BazelRunFiles.resolveVerified(
+      RULES_REPOSITORY_NAME,
+      "src", "main", "kotlin",
+      "jdeps-gen.jar").toPath()
 
     internal val NO_ARGS = arrayOf<Any>()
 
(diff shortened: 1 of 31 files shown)


tyvsmith:
Should the compiler plugin itself live in the Kotlin repo itself?

jongerrish:
> Should the compiler plugin itself live in the Kotlin repo?

As it's planned, its going to generate Bazel's jdeps proto format here:- https://github.com/bazelbuild/bazel/blob/master/src/main/protobuf/deps.proto

It's probably reasonable to extract out the collecting of the dependency information from the output format at some point if other formats are desired.

But there are another couple of other features I'm planning to add and not sure exactly how it will shake out yet....
1. Kotlin strict deps - enforce that types explicitly referenced in Kotlin source are declared as direct dependencies of the target containing those references - Bazel already supports this for Java.
2. Unused dependency warning/errors

The first one of which might end up in this plugin or a separate one, I don't really have a clear idea yet of what would be the best implementation.
The second one would be probably implemented in the JdepsMerger builder since its only there that you know if a dependency is unused from both Kotlin and Java source.

Right now I'm still banging my head on the wall how to look up the jar from which the compiler loaded a class from based on a `ClassDescriptor`.... Kotlin compiler plugins are the wild west it seems, not much out there in terms of docs / good examples for what we want to do.

src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
```diff
@@ -63,6 +64,11 @@ class KotlinToolchain private constructor(
       "src", "main", "kotlin",
       "skip-code-gen.jar").toPath()
 
+    private val JDEPS_GEN_PLUGIN = BazelRunFiles.resolveVerified(
+      "io_bazel_rules_kotlin",
```
restingbull, line 68: We can't hard code the repository rule, so use the hack when you merge from master.

We really need to pass in the paths via flags. It's on the list.

restingbull (changes requested):
Few tweaks that will sort out in the merge.

Will need to fix the commits -- you are missing a CLA.

src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
jongerrish: Done

oliviernotteghem:
@googlebot I consent.

restingbull:
Seems like the Ci doesn't like the jdeps test: https://buildkite.com/bazel/rules-kotlin-kotlin/builds/1372#1e3542f3-0976-4ab7-b2eb-d4a0fbb812fa

Can you take a look?

restingbull (approved):

PR #437, opened 2021-01-01
Implement (experimental) unused dependency checking feature.

jongerrish:
- Add a tristate (off/warn/error) experimental_report_unused_deps attribute to the toolchain.
- Implement unused dependency checking in `JDepsMerger` tool.
- Collect jdeps dependency artifacts if experimental_report_unused_deps is enabled (warn/error) and pass to KotlinBuilder. Passing these inputs triggers `JDepsMerge` action for each target causing the unused deps checking logic to be executed. Currently these inputs are not used by the `KotlinBuilder` but will be required for future work implementing a reduced classpath feature.
- Add tests for experimental_report_unused_deps off/warn/error
- Enable experimental_report_unused_deps as "warn" for examples/android

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
@@ -5,14 +5,15 @@ import io.bazel.kotlin.builder.tasks.CommandLineProgram
 import io.bazel.kotlin.builder.utils.ArgMap
 import io.bazel.kotlin.builder.utils.ArgMaps
 import io.bazel.kotlin.builder.utils.Flag
-import java.io.BufferedInputStream
-import java.io.BufferedOutputStream
-import java.io.File
+import io.bazel.kotlin.builder.utils.jars.JarHelper.Companion.INJECTING_RULE_KIND
+import io.bazel.kotlin.builder.utils.jars.JarHelper.Companion.TARGET_LABEL
+import java.io.*
 import java.nio.charset.StandardCharsets
 import java.nio.file.FileSystems
 import java.nio.file.Files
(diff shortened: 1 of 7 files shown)


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
```diff
@@ -64,9 +83,33 @@ class JdepsMerger: CommandLineProgram {
...
+          .mapNotNull { readJarOwnerFromManifest(Paths.get(it.path)).label }
+          .filter {it != label}
+
+        if (unusedLabels.isNotEmpty()) {
+          val open = "\u001b[35m\u001b[1m"
+          val close = "\u001b[0m"
+          val command =
+            """
+            |$open ** Please remove the following dependencies:$close ${unusedLabels.joinToString(" ")} from $label 
+            |$open ** You can use the following buildozer command:$close buildozer 'remove deps ${unusedLabels.joinToString(" ")}' $label
```
Kernald, line 99: I'm not entirely sure what the `\u001b[0m` ANSI code is, but if the `buildozer` command matches `^buildozer '(.*)'\\s+(.*)$`, `ibazel` can automatically run it when it sees it (see https://github.com/bazelbuild/bazel-watcher#output-runner)

restingbull (approved):

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
jongerrish: I copied these codes from Bazel's Java strict deps compiler plugin :-)

Looks like you can set any regex patterns anyway with the output runner, so should be good?

PR #443, opened 2021-01-05
Fix examples/android code to actually run successfully

nkoroste:
Changes:
- Add a manifest with a main activity to launch
- Use ``${lib_name}`` manifest variable in order to reuse the same manifest from multiple libs and inherit the right `MainActivity` from the right lib
- Define `resource_files` attr to actually include the resource dir. Modify the code to actually use the resources.
- Fixed the code to extend from `AppCompatActivity` in order to actually use `androidx.appcompat` and not crash on launch
- Renamed `lib` target to `my_kt` and `lib2` to `my_android` in order to make it easier to distinguish between the generated output files during debugging.
- Added `my_java` target to help with debugging when comparing target output file generation for java vs native android vs kotlin
- Added `.bazelproject` file for IDE


examples/android/lib/src/main/kotlin/examples/android/lib/MainActivity.kt
```diff
@@ -1,11 +1,11 @@
 package examples.android.lib
 
 import android.app.Activity
+import android.app.AlertDialog
 import android.os.Bundle
 import android.widget.Button
 import android.widget.LinearLayout
 import android.widget.LinearLayout.LayoutParams
-import androidx.appcompat.app.AlertDialog
```
restingbull, line 8: We leaned on appcompat to ensure it was included as part of the build.

Can you add something that requires it?

restingbull (changes requested):

examples/android/app/BUILD.bazel
```diff
@@ -8,6 +8,9 @@ android_binary(
     incremental_dexing = 0,
     manifest = "src/main/AndroidManifest.xml",
     multidex = "native",
+    manifest_values = {
+        "lib_name": "lib",
```
Kernald, line 12: Shouldn't these depend on the binary name rather than a dependency? This feels closer to how it would actually be used - the libraries have their own manifests. Right now you also still have two binaries with the same package name (`app` and `app3`) as they both define this as `lib`.

restingbull (approved):

nkoroste:
@googlebot I consent.

PR #446, opened 2021-01-08
Add docs for `experimental_use_abi_jars`

jongerrish:


CompileAvoidance.md
```diff
@@ -0,0 +1,38 @@
...
+    experimental_use_abi_jars = True,
+)
+```
+
+This feature is implemented using the Jetbrains Kotlin JvmABI compiler plugin for generating headers
+for Kotlin code. Unfortunately there are some known bugs with this plugin that affect less than 1%
+of targets.
+* https://youtrack.jetbrains.com/issue/KT-40340
+* https://youtrack.jetbrains.com/issue/KT-40133
+If you encounter such bugs, ABI generation can be disabled by setting the following tag
```
restingbull, line 29: one more line. :)

restingbull (approved):

Issue #403, opened 2020-11-20
DESIGN: defining java_plugin annotation processor arguments


restingbull:
Many java annotation processors allow configuration via command line arguments and system properties. These are consumed by kapt and javac,  and therefore best kept separate from java compiler arguments.

## Design Goals
* allow rule association
* avoid unnecessary boilerplate
* highly discoverable

### Nice to haves
* associate with a single java_plugin rule

PR #448, opened 2021-01-12
Add support for javac KAPT arguments

benjaminRomano:
**Background**
Internally, we needed to support javac arguments in kotlin rules to pass through dagger options to KAPT such as fastInit and disable format generated sources. 

This is our own implementation to satisfy our requirements, but I suspect this is not the desirable approach for this based on this issue: https://github.com/bazelbuild/rules_kotlin/issues/403. However, it may be helpful for getting us closer to the right APIs here.

**Change**
* Add a `kapt_javac_arguments` attribute to the kotlin toolchain rule
* Pass through `kapt_javac_arguments` to the KotlinWorker through a `javac_arguments` field in the kotlin model proto.

To encode the key/value pairs through to the kotlin worker we use a `:` delimiter. This can cause issues if the key used for an annotation processor's option includes a `:` which I'm not sure is possible or not.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -452,6 +452,13 @@ def _run_kt_builder_action(
         omit_if_empty = True,
     )
 
+    # Add KAPT javac arguments. This is only defined on the toolchain for now, but could be added to rules if need be
+    args.add_all(
+        "--javac_arguments",
+        ["%s:%s" % (k, v) for k, v in toolchains.kt.kapt_javac_arguments.items()],
+        omit_if_empty = True,
+    )
+
     args.add_all(
         "--compiler_plugin",
         [j for p in compiler_compiler_plugins for j in p.plugin_jars],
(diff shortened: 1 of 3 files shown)


restingbull:
Apologies for the delay.

Quick question: how often is the toolchain args used vs. rule args?

benjaminRomano:
In this PR, I only added support to toolchain, we didn't have a use case where we'd want this to be applied to a specific rule.

I believe its quite rare for us to use rule args to configure kt targets.

nkoroste:
I think we might need to consider new `kt_plugin` or `kt_annotation_processor` rule to allow us to pass-through args to KAPT.

Currently with Ben's change we apply Dagger flags to the entire KT toolchain which causes warnings like this:
```
warning: warning: The following options were not recognized by any processor: '[dagger.fastInit, dagger.nullableValidation, dagger.experimentalDaggerErrorMessages, dagger.formatGeneratedSource]'
```

because we end up applying Dagger flags to all actions even if Dagger AP is not used. 

@jongerrish

PR #451, opened 2021-01-14
Add support for kt_android_local_test

Bencodes:
Adding support for Kotlin and `android_local_test`. Things that changed in this PR:
- The `friends` attr has been moved from the `kt_jvm_test` group to the common group. This is to support creating `kt_jvm_library` bundles that can be packaged up us a dependency in a test target.
- `compile.bzl` was updated to enforce that `testonly` targets have `friends` associated with them. This was to keep compatibility with the old behavior.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -90,6 +90,9 @@ def _compiler_toolchains(ctx):
 def _compiler_friends(ctx, friends):
     """Creates a struct of friends meta data"""
 
+    if len(friends) > 0 and ctx.attr.testonly == False:
+        fail("only testonly targets can have friends associated with them")
+
     # TODO extract and move this into common. Need to make it generic first.
     if len(friends) == 0:
         return struct(
--- a/kotlin/internal/jvm/android.bzl
+++ b/kotlin/internal/jvm/android.bzl
@@ -18,7 +18,7 @@ load(
 
 _ANDROID_SDK_JAR = "%s" % Label("//third_party:android_sdk")
 
-def _kt_android_artifact(name, srcs = [], deps = [], plugins = [], kotlinc_opts = None, javac_opts = None, enable_data_binding = False, **kwargs):
+def _kt_android_artifact(name, srcs = [], deps = [], plugins = [], friends = [], kotlinc_opts = None, javac_opts = None, enable_data_binding = False, **kwargs):
     """Delegates Android related build attributes to the native rules but uses the Kotlin builder to compile Java and
     Kotlin srcs. Returns a sequence of labels that a wrapping macro should export.
     """
@@ -41,7 +41,8 @@ def _kt_android_artifact(name, srcs = [], deps = [], plugins = [], kotlinc_opts 
         srcs = srcs,
         deps = base_deps + [base_name],
         plugins = plugins,
-        testonly = kwargs.get("testonly", default = 0),
+        friends = friends,
+        testonly = kwargs.get("testonly", default = False),
         visibility = ["//visibility:private"],
         kotlinc_opts = kotlinc_opts,
         javac_opts = javac_opts,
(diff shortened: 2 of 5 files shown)


keith (approved):

cgruber:
Glad you've done this - I've been too busy to get to freinds support. :(. 

That said, we had a bunch of conversations about this in a few places, and we were thinking of taking this moment to change from `friends` to `associate_to`, to match the terminology of the gradle plugin, the main other way people use kotlinc.

I'll need to dig into the review to see how you're handling the actual details. Just wanted to start with an initial comment about naming.

kotlin/internal/jvm/compile.bzl
```diff
@@ -90,6 +90,9 @@ def _compiler_toolchains(ctx):
 def _compiler_friends(ctx, friends):
     """Creates a struct of friends meta data"""
 
+    if len(friends) > 0 and ctx.attr.testonly == False:
```
cgruber, line 93: Ouch - this seems a bit tricky - adding an API (attribute on a rule) for friends, then breaking if you use it, seems like a bad pattern.

Was there a reason you didn't want to extend this across the libraries as well?

cgruber: If you dont' want to tackle the full friends problem, that's fine - but in that case, please just add a duplicate of this to both test rules, or make a common test attributes section. I don't mind duplication here - or rather I prefer it to opening up an API we clobber, more broadly in the rule-set.

cgruber (changes requested):

cgruber:
Given that this is not really expanding the functionality or availability of friends, really just keeping kt_android_local_test having parity, I'm inclined to not push to expand this much. I can roll my extended freinds support PR on top of this, once it's merged.

kotlin/internal/jvm/compile.bzl
Bencodes: I opted to preserve the previous behavior under the assumption that there was some reason we didn't support the friends attribute outside of tests. I'm happy to just remove it and allow all targets to have friends.

cgruber: No, there's a broader set of concerns to address if we want to open it to libraries. I'll roll that PR. I've already done it twice internally.

For now, the main issue is just that we don't actually expose those properties on any rule that doesn't use them, so we dont' have this "if ! testonly" thing.  Just duplicate the attribute in the android test, and don't include this check, and we're fine. I'll adjust when I add the other functionality.

Bencodes: > Just duplicate the attribute in the android test

Not quite sure what you mean here. Are you wanting to remove the `testonly` check and leave the `friends` attribute in `_common_attr` so that it's available to non `_test` targets?

cgruber: No, I mean restore the attribute where it was (in kt_jvm_test), remove the conditional, and add a duplicate of the attribute on kt_android_local_test, so both tests have it, but nothing else. The present logic would actually let a kt_jvm_library use friends if it was testonly=true, which isn't what we want (at this point, in this PR) and it would make for the attribute being available for code-completion on rules that it will ultimately throw errors on, which is bad ergonomics.

Bencodes: Right now with the implementation I have here the `friends` attribute has to be added to the `kt_jvm_library` rule because that's what's handling the underlying sources (via `_kt_android_artifact`). Without that attr being exposed we can't support test friends in `kt_android_local_test`.

cgruber: Oooh. Ok - I see it now.  Crap.  That piece I completely misunderstood. 

Ok, there are ways of hacking this to preserve the API surface, but given that I'm going to update this to support it more broadly, lets' just leave it as it is.

cgruber (approved):
Given the fact that kt_jvm_library needs this feature for test-lib handling inside of kt_android_local_test, this is regrettable, but necessary. So let's just move on with it.

PR #465, opened 2021-01-27
Expose kotlinc's -Xfriend-paths to all jvm/android kt rules under the attribute 'assocates='

### LINKED ISSUE #211: Broaden support for "friends" (access to `internal` members) in rules_kotlin

cgruber:
Associates lets a library associate it self to other libraries, making them part of the same module. This is constrained such that while multiple libraries may be associated, they must all shard the same module, and so cannot associate to anything that is part of a different module. These module relationships are in the bazel build graph, not the contents of the jars as such.

This module membership is transitive (within the above-mentioned constraint), though strict-deps would stop that. Also, only kotlin targets can be associated.

Per discussions across several media, the name "associates" was chosen over "friends" (despite the kotlinc flag being -Xfriend-paths) as that is the terminology used in the gradle kotlin plugin, which is kotlin's primary delivery vehicle, and to avoid confusion with the C++ friends concepts.  The pre-existing "friends" attribute is preserved for backward compatibility with a warning. Future PR will add a flag to turn off that support, and then we'll delete it.

kt_jvm_import does not include this facility, but these can just set their module_name in common to participate.

Android should work, but because kt_android_* is a macro not a rule, the implicit target `//my/android/library:mytarget_kt` should be friended, since it has a KtJvmInfo. The `//my/android/library` will macro-resolve into an android_library. Until the android rules get the right kind of love, such that we can make a rule that has KtJvmInfo AND android-whatever providers, this simply is a known limitation we'll have to live with.

Also, the prior implementation shoved the full transitive closure (all jars, kotlin or no) of the friend= into the `-Xfriends-paths` flag, which is awful. This PR does break that, in case people were relying on that by some oddity. The fix is to just add the targets directly. 

Fixes #211

### DIFF
--- a/kotlin/internal/jvm/associates.bzl
+++ b/kotlin/internal/jvm/associates.bzl
@@ -63,16 +63,21 @@ def _get_associates(ctx):
             module_name = list(module_names)[0],
         )
 
-def _get_jars(associates):
-    # Uses depset() expansion because there is no flatten/flatmap list function.
-    jars = [
-        a[JavaInfo].compile_jars
-        for a in depset(transitive = associates.jars).to_list()
-        if a[JavaInfo].compile_jars
+def _flatten_jars(nested_jars_depset):
+    """Returns a list of strings containing the compile_jars for depset of targets.
+
+    This ends up unwinding the nesting of depsets, since compile_jars contains depsets inside
+    the nested_jars targets, which themselves are depsets.  This function is intended to be called
+    lazily form within Args.add_all(map_each) as it collapses depsets.
+    """
+    compile_jars_depsets = [
+        target[JavaInfo].compile_jars
+        for target in nested_jars_depset.to_list()
+        if target[JavaInfo].compile_jars
     ]
-    return depset(transitive = jars).to_list()
+    return [file.path for file in depset(transitive = compile_jars_depsets).to_list()]
 
 associate_utils = struct(
     get_associates = _get_associates,
-    get_jars = _get_jars,
+    flatten_jars = _flatten_jars,
 )
--- a/kotlin/internal/jvm/associates.bzl
+++ b/kotlin/internal/jvm/associates.bzl
@@ -1,0 +1,78 @@
+# Copyright 2020 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+load(
+    "//kotlin/internal:defs.bzl",
+    _KtJvmInfo = "KtJvmInfo",
+)
+load(
+    "//kotlin/internal/utils:sets.bzl",
+    _sets = "sets",
+)
+load(
+    "//kotlin/internal/utils:utils.bzl",
+    _utils = "utils",
+)
+
+def _get_associates(ctx):
+    """Creates a struct of associates meta data"""
+
+    friends_legacy = getattr(ctx.attr, "friends", [])
+    associates = getattr(ctx.attr, "associates", [])
+
+    if friends_legacy:
+        print("WARNING: friends=[...] is deprecated, please prefer associates=[...] instead.")
+        if associates:
+            fail("friends= may not be used together with associates=. Use one or the other.")
+        elif ctx.attr.testonly == False:
+            fail("Only testonly targets can use the friends attribute. ")
+        else:
+            associates = friends_legacy
+
+    if not bool(associates):
+        return struct(
+            targets = [],
+            module_name = _utils.derive_module_name(ctx),
+            jars = [],
+        )
+    elif ctx.attr.module_name:
+        fail("if associates have been set then module_name cannot be provided")
+    else:
+        jars = [depset([a], transitive = a[_KtJvmInfo].module_jars) for a in associates]
+        module_names = _sets.copy_of([x[_KtJvmInfo].module_name for x in associates])
+        if len(module_names) > 1:
+            fail("Dependencies from several different kotlin modules cannot be associated. " +
+                 "Associates can see each other's \"internal\" members, and so must only be " +
+                 "used with other targets in the same module: \n%s" % module_names)
+        if len(module_names) < 1:
+            # This should be impossible
+            fail("Error in rules - a KtJvmInfo was found which did not have a module_name")
+        return struct(
+            targets = associates,
+            jars = jars,
+            module_name = list(module_names)[0],
+        )
+
+def _get_jars(associates):
+    # Uses depset() expansion because there is no flatten/flatmap list function.
+    jars = [
+        a[JavaInfo].compile_jars
+        for a in depset(transitive = associates.jars).to_list()
+        if a[JavaInfo].compile_jars
+    ]
+    return depset(transitive = jars).to_list()
+
+associate_utils = struct(
+    get_associates = _get_associates,
+    get_jars = _get_jars,
+)
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -35,8 +35,16 @@ load(
     _plugins_to_options = "plugins_to_options",
 )
 load(
+    "//kotlin/internal/jvm:associates.bzl",
+    _associate_utils = "associate_utils",
+)
+load(
     "//kotlin/internal/utils:utils.bzl",
     _utils = "utils",
+)
+load(
+    "//kotlin/internal/utils:sets.bzl",
+    _sets = "sets",
 )
 load(
     "@bazel_tools//tools/jdk:toolchain_utils.bzl",
@@ -87,36 +95,18 @@ def _compiler_toolchains(ctx):
         java_runtime = find_java_runtime_toolchain(ctx, ctx.attr._host_javabase),
     )
 
-def _compiler_friends(ctx, friends):
-    """Creates a struct of friends meta data"""
-
-    if len(friends) > 0 and ctx.attr.testonly == False:
-        fail("only testonly targets can have friends associated with them")
-
-    # TODO extract and move this into common. Need to make it generic first.
-    if len(friends) == 0:
-        return struct(
-            targets = [],
-            module_name = _utils.derive_module_name(ctx),
-            paths = [],
-        )
-    elif len(friends) == 1:
-        if friends[0][_KtJvmInfo] == None:
-            fail("only kotlin dependencies can be friends")
-        elif ctx.attr.module_name:
-            fail("if friends has been set then module_name cannot be provided")
-        else:
-            return struct(
-                targets = friends,
-                paths = _java_info(friends[0]).compile_jars,
-                module_name = friends[0][_KtJvmInfo].module_name,
-            )
-    else:
-        fail("only one friend is possible")
-
-def _jvm_deps(toolchains, friend, deps, runtime_deps = []):
+def _jvm_deps(toolchains, associated_targets, deps, runtime_deps = []):
     """Encapsulates jvm dependency metadata."""
-    dep_infos = [_java_info(d) for d in friend.targets + deps] + [toolchains.kt.jvm_stdlibs]
+    diff = _sets.intersection(
+        _sets.copy_of([x.label for x in associated_targets]),
+        _sets.copy_of([x.label for x in deps]),
+    )
+    if diff:
+        fail(
+            "\n------\nTargets should only be put in associates= or deps=, not both:\n%s" %
+            ",\n ".join(["    %s" % x for x in list(diff)]),
+        )
+    dep_infos = [_java_info(d) for d in associated_targets + deps] + [toolchains.kt.jvm_stdlibs]
     return struct(
         deps = dep_infos,
         compile_jars = depset(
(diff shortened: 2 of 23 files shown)


kotlin/internal/jvm/associates.bzl
```diff
@@ -0,0 +1,78 @@
...
+)
+
+def _get_associates(ctx):
+    """Creates a struct of associates meta data"""
+
+    friends_legacy = getattr(ctx.attr, "friends", [])
+    associates = getattr(ctx.attr, "associates", [])
+
+    if friends_legacy:
+        print("WARNING: friends=[...] is deprecated, please prefer associates=[...] instead.")
```
Bencodes, line 34: Is the flag to drop support for friends going to make it in before the next alpha cut?

cgruber: Unlikely. For now it's just a warning if you use the old one.

kotlin/internal/jvm/associates.bzl
```diff
@@ -0,0 +1,78 @@
...
+            targets = associates,
+            jars = jars,
+            module_name = list(module_names)[0],
+        )
+
+def _get_jars(associates):
+    # Uses depset() expansion because there is no flatten/flatmap list function.
+    jars = [
+        a[JavaInfo].compile_jars
+        for a in depset(transitive = associates.jars).to_list()
```
restingbull, line 70: Use add_all(map_each), not depset.to_list. This can eat up a lot of memory during analysis phase.

kotlin/internal/jvm/compile.bzl
```diff
@@ -398,8 +389,11 @@ def _run_kt_builder_action(
     args.add_all("--sources", srcs.all_srcs, omit_if_empty = True)
     args.add_all("--source_jars", srcs.src_jars + generated_src_jars, omit_if_empty = True)
     args.add_all("--deps_artifacts", deps_artifacts, omit_if_empty = True)
-
-    args.add_joined("--kotlin_friend_paths", friend.paths, join_with = "\n")
+    args.add_joined(
+        "--kotlin_friend_paths",
+        _associate_utils.get_jars(associates),
```
restingbull, line 394: Use add_all(map_each), not depset.to_list. This can eat up a lot of memory during analysis phase.

kotlin/internal/utils/sets.bzl
```diff
@@ -0,0 +1,79 @@
+#
```
restingbull, line 1: Why not use skylib? https://github.com/bazelbuild/bazel-skylib/blob/master/docs/new_sets_doc.md

restingbull (changes requested):

kotlin/internal/utils/sets.bzl
cgruber: I had some issues when I did so on bazel_maven_repository, partly in terms of being over-wrought for what I needed, having a separate whole struct wrapping the inner dict. But also adding an extra workspace dep is pretty rough in a bazel world. They don't have good ways of handling and identifying diamond dependency versioning problems, so you just "get what you get" and no way to validate that the version of skylib I"m using here, isn't going to mess with the version of skylib pushed by the maven integration, etc. etc.  This is just cleaner.  If we find we need it anyway, then yeah, rip this out, but for now it seemed just much cleaner.

kotlin/internal/jvm/jvm.bzl
```diff
@@ -198,9 +198,17 @@ _common_attr = utils.add_dicts(
         [Attributes common to all build rules](https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes).""",
             allow_files = True,
         ),
+        "associates": attr.label_list(
```
chancila, line 201: any idea how this will interact with the intellij stuff? the aspect has a list of attributes it walks and this is introducing something new

[https://github.com/bazelbuild/intellij/blob/master/aspect/intellij_info_impl.bzl](url)

kotlin/internal/jvm/associates.bzl
cgruber: Done.

kotlin/internal/jvm/compile.bzl
cgruber: Done

kotlin/internal/jvm/jvm.bzl
cgruber: As I understand it, the deps are extracted from the actual providers, which in this context include the associates as deps - so they should just show up. That said, I've done some initial testing (I used the intellij plugin myself) and haven't seen any issues with it figuring things out, but I've not used it (yet) in complicated scenarios, so we'll have to test it more.

In a pinch, we can make the IJ stuff more aware of associates, but it oughtn't need that, since it's using the underlying providers, and those providers include deps+associates in their JavaInfo dependency information.

restingbull: )*$%)(*^&)!

`associates` will not be included in a very precise edge case -- e.g. if the only included target references a set of dependencies by `associates`, it will not be seen due to the aspect only walking certain attributes.

Walked attributes: 
Definitions: https://github.com/bazelbuild/intellij/blob/master/aspect/intellij_info_impl.bzl#L29
Usage: https://github.com/bazelbuild/intellij/blob/master/aspect/intellij_info_impl.bzl#L1090

restingbull: It's the only argument define associates in two places I've seen.

restingbull (approved):

kotlin/internal/jvm/jvm.bzl
restingbull: Approving  -- it's a pretty small edge case (if really annoying). Most usage is to include entire subdirs. I believe we'd need to make a request against the plugin.

kotlin/internal/jvm/associates.bzl
```diff
@@ -63,16 +63,21 @@ def _get_associates(ctx):
...
-        if a[JavaInfo].compile_jars
+def _flatten_jars(nested_jars_depset):
+    """Returns a list of strings containing the compile_jars for depset of targets.
+
+    This ends up unwinding the nesting of depsets, since compile_jars contains depsets inside
+    the nested_jars targets, which themselves are depsets.  This function is intended to be called
+    lazily form within Args.add_all(map_each) as it collapses depsets.
+    """
+    compile_jars_depsets = [
+        target[JavaInfo].compile_jars
```
restingbull, line 74: This probably means we should tweak provider storage to avoid the filtering of double nesting.

kotlin/internal/jvm/jvm.bzl
cgruber: Ugh. I had a whole theory of why it would work, but in reading the code, I see it doesn't do what I thought it did. Yeah, damn.

cgruber: I'm going to merge this, and we can sort it out, since the problem will only occur if one uses the feature (and the status quo is that the feature isn't used), and as you say, it's a small edge-case.

kotlin/internal/jvm/associates.bzl
cgruber: Maybe, but these are small lists, generally, since they're direct.

PR #466, opened 2021-01-27
Make all the child workspaces share the same .bazelversion.

cgruber:
Add in symlinks to make sure we're not running different versions of bazel (by default) on different example workspaces.


cgruber:
Please don't merge until the parent PR is merged.

Bencodes (approved):

PR #473, opened 2021-02-03
Fix build failure when using Java only modules with annotation procesors

jongerrish:
Include annotation processor generated source from Java only targets.

In these cases, Bazel's Java Builder / Turbine handler annotation processors not KAPT.


nkoroste:
Ooops, thank for the fix!

restingbull (approved):

PR #341, opened 2020-06-15
Specify what the compile_jar value should be when building the library

szinn:
Resolves issue [340](https://github.com/bazelbuild/rules_kotlin/issues/340)

PR #479, opened 2021-02-13
Set the compile jar to be the same as the output jar

szinn:
When kt_jvm_library is used to create a library that contains auxiliary files (lie .proto files) it does so correctly.  However, when referred to via deps in the build of another module, bazel creates the -hjar file that strips these needed components from the .jar file used to link against.

This PR sets the compile / ijar resource to be the same as the created .jar file, thus eliminating this intermediate hjar creation.  This is the same as a previous PR (#341)

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -609,7 +609,7 @@ def kt_jvm_produce_jar_actions(ctx, rule_kind):
 
     java_info = JavaInfo(
         output_jar = output_jar,
-        compile_jar = compile_jar,
+        compile_jar = output_jar,
         source_jar = source_jar,
         jdeps = ctx.outputs.jdeps,
         deps = compile_deps.deps,
@@ -633,7 +633,7 @@ def kt_jvm_produce_jar_actions(ctx, rule_kind):
                 jdeps = ctx.outputs.jdeps,
                 jars = [struct(
                     class_jar = output_jar,
-                    ijar = compile_jar,
+                    ijar = output_jar,
                     source_jars = [source_jar],
                 )],
             ),


restingbull:
@jongerrish -- have a look? This would have changed during the large refactoring.

kotlin/internal/jvm/compile.bzl
```diff
@@ -609,7 +609,7 @@ def kt_jvm_produce_jar_actions(ctx, rule_kind):
 
     java_info = JavaInfo(
         output_jar = output_jar,
-        compile_jar = compile_jar,
```
jongerrish, line 612: This is going to make the compile jar the same as the output jar which will break the whole header / ABI jars for compile avoidance optimization.

Are you expecting any auxiliary files to be included in the compile jars? Note that the final binary (e.g: {java,android}_binary) will link in the output jar for each target, not the compile jar. That is intended only for compilation and why we do this optimization.

A test case showing what is broken would be helpful.

szinn: Yes, I'm expecting auxiliary files in the output_jar to be available for use by follow-on targets.

A specific example is doing a wire_proto compile which generates the jar containing the generated .class files as well as the .proto files.

A follow-on target uses takes as input this jar and requires both the .proto and .class files as input.

Without this PR, the .proto files get removed when the ijar / hjar file is generated prior to compiling the follow-on target.

An alternate solution would be to include the srcjar as an input to the follow-on target, but the output jar is really the proper generated target.

szinn: Is there a way to force the output.jar to be used as a dep in a follow-on target build?

jongerrish: Is `wire_proto` a Skylark rule?

I'm guessing that `wire_proto` invokes some tool to compile its local proto source into either Kotlin/Java source code or direct to JVM bytecode?

If the compiler needs access to transitive proto source, i.e: local proto source references protos in other libraries then you can make these protos available via a provider e.g: WireProtoInfo that your rule can then collect transitive instances of.

szinn:
I was able to fix my wire proto bzl file to reference full_compile_jars rather than compile_jars.  No need for this PR.

Issue #453, opened 2021-01-18
Do not issue "worker" warning when compiling with a Remote Build Environment


shs96c:
When running a build using a remote build environment, such as [buildbarn](https://github.com/buildbarn), the error message `HINT: KotlinCompile will compile faster if you run: echo "build --strategy=KotlinCompile=worker" >>~/.bazelrc` isn't actionable. Could this warning be disabled or hidden behind a flag?

restingbull:
I'm thinking the note isn't that useful, tbh.

PR #481, opened 2021-02-16
Add real multiplex support

restingbull:
Replace worker strategy with a more general one relying on coroutines for multiplexing.

Fixes https://github.com/bazelbuild/rules_kotlin/issues/453 as well.


src/main/kotlin/io/bazel/kotlin/builder/KotlinBuilderComponent.java
```diff
@@ -52,15 +52,10 @@
     class Module {
 
         @Provides
-        public PrintStream provideDebugPrintStream() {
+        public PrintStream provideDebugPrintStream(WorkerContext ctx) {
```
jongerrish, line 55: Unused param?

src/main/kotlin/io/bazel/kotlin/builder/KotlinBuilderComponent.java
```diff
@@ -38,10 +38,10 @@
 
     Kotlin2JsTaskExecutor jsTaskExecutor();
 
-    BazelWorker worker();
+    CompileKotlin compileKotlin();
```
jongerrish, line 41: The name suggests to me an action "Compile Kotlin" but I think this is producing a Kotlin Worker Request Handler or something? Could it be named as such?

jongerrish: For example the jdeps merger equivalent is `work()` - perhaps there's a naming you can find that is consistent between them and descriptive about the thing being produced?

jongerrish (approved):
Worth including a link to multiplexed workers in the PR description: https://docs.bazel.build/versions/master/multiplex-worker.html

Also, should we have an option to disable/enable them? I think this can be done simply my making "supports_multiplexed_worker" attribute conditional (on some flag in the tool chain?).... I know they had some attempts at fixing https://docs.bazel.build/versions/master/multiplex-worker.html#warning-about-rare-bug not sure if its still a thing or not but caution maybe?

src/main/kotlin/io/bazel/kotlin/builder/KotlinBuilderComponent.java
restingbull: Sure -- naming is the hardest part.

I would say that it is compiling kotlin (for the most part) and the request handling is an implementation detail. But I'm not married to that.

src/main/kotlin/io/bazel/worker/WorkerContext.kt
```diff
@@ -0,0 +1,162 @@
...
+import java.util.logging.Level
+import java.util.logging.Logger
+import java.util.logging.SimpleFormatter
+import java.util.logging.StreamHandler
+
+/** WorkerContext encapsulates logging, filesystem, and profiling for a task invocation. */
+class WorkerContext private constructor(
+  private val name: String = Companion::class.java.canonicalName,
+  private val verbose: Granularity = INFO
+) : Closeable, ScopeLogging by ContextLogger(name, verbose.level, null) {
```
cgruber, line 38: Are you doing an anonymous delegate? That's... huh. I didn't realize you could do that. I'm used to delegating to a constructor parameter or property object, but not an anonymous instance like this.

src/main/kotlin/io/bazel/worker/WorkingDirectoryContext.kt
```diff
@@ -0,0 +1,41 @@
...
+        wd.forWork()
+      }
+
+    fun newContext(): WorkingDirectoryContext = WorkingDirectoryContext(
+      Files.createTempDirectory("working")
+    )
+  }
+
+  override fun close() {
+    Files.walk(dir).sorted(Comparator.reverseOrder()).forEach(Files::delete)
```
cgruber, line 39: I presume you're just going to accept if it fails here? Best effort?

src/test/kotlin/io/bazel/worker/IOTest.kt
```diff
@@ -0,0 +1,65 @@
...
+    System.setIn(stdIn)
+    System.setOut(stdOut)
+  }
+
+  @Test
+  fun capture() {
+    Truth.assertThat(captured.written()).isEmpty()
+    IO.capture().use { io ->
+      println("foo foo is on the loose")
+      Truth.assertThat(io.captured.written()).isEqualTo("foo foo is on the loose\n")
```
cgruber, line 61: Nit - static import assertThat()

src/test/kotlin/io/bazel/worker/PersistentWorkerTest.kt
```diff
@@ -0,0 +1,118 @@
...
+import com.google.devtools.build.lib.worker.WorkerProtocol.WorkRequest
+import com.google.devtools.build.lib.worker.WorkerProtocol.WorkResponse
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import org.junit.Test
+import java.io.ByteArrayOutputStream
+import java.nio.charset.StandardCharsets.UTF_8
+
+class PersistentWorkerTest {
+
+  @ExperimentalCoroutinesApi
```
cgruber, line 30: Maybe put this on the class.

cgruber (approved):
Some nits and comments, but I like this structure.

restingbull:
> https://docs.bazel.build/versions/master/multiplex-worker.html

Multiplex is controlled on the command line, so enabling it means you know what you are doing.

src/main/kotlin/io/bazel/worker/WorkingDirectoryContext.kt
restingbull: Pretty much -- it's not really actionable. In many respects, this shouldn't be done during in the invocation.

restingbull: Ah, nm. Catch and log.

PR #483, opened 2021-02-17
Pass a few more args to android_local_test

Bencodes:


keith (approved):
How do the shards work? We don't use that in bazel today under the assumption our test bundles are individually small enough

restingbull:
> How do the shards work? We don't use that in bazel today under the assumption our test bundles are individually small enough

They work best wit RBE -- basically, the shard number is the number of actions spawned and the test runner selects the tests based on the number of shard it resides in.

restingbull (approved):

Issue #485, opened 2021-02-18
StackOverflowError during jdeps generation


Bencodes:
Encountered this one while doing some testing with the latest master. Seems to be the result of generic class types that reference themself. Example from Junit:

```java
public abstract class AbstractObjectAssert<SELF extends AbstractObjectAssert<SELF, ACTUAL>, ACTUAL>
    extends AbstractAssert<SELF, ACTUAL> {
...
```

```console
exception: java.lang.StackOverflowError
        at org.jetbrains.kotlin.builtins.KotlinBuiltIns.isTypeConstructorForGivenClass(KotlinBuiltIns.java:724)
        at org.jetbrains.kotlin.builtins.KotlinBuiltIns.isConstructedFromGivenClass(KotlinBuiltIns.java:716)
        at org.jetbrains.kotlin.builtins.KotlinBuiltIns.isNothingOrNullableNothing(KotlinBuiltIns.java:841)
        at org.jetbrains.kotlin.builtins.KotlinBuiltIns.isNothing(KotlinBuiltIns.java:831)
        at org.jetbrains.kotlin.types.TypeSubstitutor.unsafeSubstitute(TypeSubstitutor.java:198)
        at org.jetbrains.kotlin.types.TypeSubstitutor.substituteTypeArguments(TypeSubstitutor.java:329)
        at org.jetbrains.kotlin.types.TypeSubstitutor.substituteCompoundType(TypeSubstitutor.java:308)
        at org.jetbrains.kotlin.types.TypeSubstitutor.unsafeSubstitute(TypeSubstitutor.java:245)
        at org.jetbrains.kotlin.types.TypeSubstitutor.unsafeSubstitute(TypeSubstitutor.java:173)
        at org.jetbrains.kotlin.types.TypeSubstitutor.substituteWithoutApproximation(TypeSubstitutor.java:126)
        at org.jetbrains.kotlin.types.TypeSubstitutor.substitute(TypeSubstitutor.java:111)
        at org.jetbrains.kotlin.types.TypeSubstitutor.substitute(TypeSubstitutor.java:104)
        at org.jetbrains.kotlin.types.StarProjectionImplKt.starProjectionType(StarProjectionImpl.kt:55)
        at org.jetbrains.kotlin.types.StarProjectionImpl$_type$2.invoke(StarProjectionImpl.kt:35)
        at org.jetbrains.kotlin.types.StarProjectionImpl$_type$2.invoke(StarProjectionImpl.kt:26)
[...]

PR #486, opened 2021-02-18
Fix jdeps collectTypeArguments stackoverflow error

Bencodes:
https://github.com/bazelbuild/rules_kotlin/issues/485


Bencodes:
@jongerrish

restingbull:
@jongerrish

jongerrish:
Perfect, you could also validate it with a test see: io.bazel.kotlin.builder.tasks.jvm.KotlinBuilderJvmJdepsTest#function call parameter type nested type parameters should be an explicit dependency otherwise I'll add it later.

jongerrish (approved):
Thanks for the fix @Bencodes

restingbull (approved):

Issue #23, opened 2018-02-14
Passing kotlinc options


jmmk:
I don't see any documentation around hooks into the Kotlin compiler. Specifically, I want to be able to pass `-Xcoroutines=enable`

hsyed:
I decided to hide the compiler switches when I forked from pubref:

* To ensure user flags cannot break the builder --e.g., kapt arguments are managed by the kotlin builder.
*  For future proofing the design across the different aspects. We have to share the an overall rule design with native and js.
* To allow us to have a configuration contract between the intellij plugin and the rules repo. 

Ideally toolchains are the mechanism we should use for configuring the compiler. Currently we don't know what is going to happen to coroutines when it goes live, will the flag still be required / be available, will there be a core coroutines library included in the compiler distribution.  

A quick fix for now might be to look for certain tags --e.g., `kotlinx-coroutines` and add the flag in skylark if its present.

jmmk:
The java rules have `javacopts`. Is that different from allowing `kotlincopts`?

hsyed:
Imo yes, Kotlinc flags should be configured by the rules / builder.  I'm up for making the kotlinc-coroutines configurable just not by opening up the flags. If we open up arbitrary kotlinc opts it would have to  have to pretty much disallow most of the flags. 

Toolchains should be used for this for now the toolchain should allow setting`-language-version`, `-api-version`, `-jvm-target`.  We could add a `feature-flag` attribute that takes strings which translate to experimental kotlinc configurations.

cgruber:
Let's not call it "no" at this point, rather let's add support for well
known flags as we get requests for them.  I'm sympathetic to the fact that
we do, IIUC, allow this for the java builder, but I'd want to see specific
examples that we want to support.

hsyed:
I have added the kotlin toolchain boilerplate for kt_jvm_. This makes some of the flags that are currently hard coded configurable. 

Skydoc is broken now it doesnt recognise the toolchain attribute and hust errors out. I think we should just follow the go repo and write the rule documentation in rst. 

I’m am thinking that the tags attribute should be sufficient to convey some non standard tunables coroutine can go in here as well, so when it becomes redundant we won’t have baggage. 

If you agree on this approach we need a labelling scheme Coroutines is a universal flag afaik/

hsyed:
Coroutines has the production ready stamp from JB. Might as well enable it with the toolchain commit.

hsyed:
[...]

PR #487, opened 2021-02-18
pass the multiple platform flag to kotlin compiler

ivyspirit:
We are building Kotlin multiple platform project, which need bazel to support building kotlin code with KMP features. `-Xmulti-platform` flag is needed to be passed to kotlinc compiler to avoid the compile errors. 

Based on the statue of this [issue](https://github.com/bazelbuild/rules_kotlin/issues/23) there is no generic way of passing kotlinc compiler arguements. This PR add the arg to the `kt_kotlinc_options` so it can be enabled or disabled. 

Test:
Tested with local `local_repository`, kmp code compile successful.
Before got error:
` error: the feature "multi platform projects" is experimental and should be enabled explicitly


ivyspirit:
@googlebot I signed it!

ivyspirit:
@googlebot I fixed it.

restingbull (approved):

PR #491, opened 2021-02-20
Avoid passing kapt args to kotlinc unless outputs are provided

Bencodes:


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
```diff
@@ -51,7 +51,17 @@ class KotlinJvmTaskExecutor @Inject internal constructor(
   fun execute(context: CompilationTaskContext, task: JvmCompilationTask) {
     val preprocessedTask = task
       .preProcessingSteps(context)
-      .runPlugins(context, plugins, compiler)
+      .let {
+        // Avoid passing in the kapt cli args unless we absolutely need to
+        // The jars will be conditionally packed up at the end of the execution if any of these outputs are provided
+        if (task.outputs.generatedJavaSrcJar.isNotEmpty()
```
Bencodes, line 57: Double check my logic here but from what i'm seeing the `KotlinCompile` phase doesn't need to run kapt at all to run successfully.

restingbull:
Apologies! I missed this pr.
@jongerrish Have a gander?

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
jongerrish: Yeah, this looks right. When these outputs are not requested there is no need to run kapt.

I guess when `experimental_use_abi_jars == false` then KAPT is always going to be requested, so this avoids calling runPluginsUsingKapt in the case where `experimental_use_abi_jars == true`.

I think `runPluginsUsingKapt` short circuits though when there are no annotation processors, no stub phase compiler plugins or no sources so I'm not sure we're actually running kapt for the `KotlinCompile` case as we set `annotation_processors = [],` in compile.bzl:813

Ben, are you seeing Kapt being run when there are no annotation processors passed?

cgruber: I'd like to merge this, but it feels like a pending clarification we should have before I do so.

cgruber (changes requested):

PR #493, opened 2021-02-23
Add Jetpack Compose example to CI

Kernald:
This PR shows a few issues with the Jetpack Compose example:
- It's not built on CI
- It actually doesn't build.

But also with the overall example structure. There seem to be a [hook set-up on CI to use a release build of the rules](https://github.com/bazelbuild/rules_kotlin/blob/master/.bazelci/presubmit.yml#L33-L37), which obviously doesn't work when building locally. The problem is, when using the parent repo, specifying a `compiler_release` in `kotlin_repositories` doesn't seem to work (not entirely sure why).

While this _might_ end up building on CI (edit: it doesn't) with the few changes here, it definitely does not build locally as Compose 1.0.0-alpha09 requires kotlinc 1.4.21, which there's no way to force with the current set-up locally - the default version is 1.4.20.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -94,6 +94,16 @@ tasks:
     build_targets:
       - //coroutines-helloworld/...
       - //express/...
+  example-jetpack-compose:
+    name: "Example - Jetpack Compose"
+    platform: ubuntu1804
+    shell_commands:
+      - "cd ../.. && bazel build //:rules_kotlin_release && rm -rf /tmp/rules_kotlin_release && mkdir -p /tmp/rules_kotlin_release && tar -C /tmp/rules_kotlin_release -xvf bazel-bin/rules_kotlin_release.tgz"
+    working_directory: examples/jetpack_compose
+    test_flags:
+      - "--override_repository=io_bazel_rules_kotlin=/tmp/rules_kotlin_release"
+    test_targets:
+      - //...
   stardoc:
     name: Stardoc api documentation
     platform: ubuntu1804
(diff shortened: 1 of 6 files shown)


Kernald:
cc @Bencodes who initially wrote the sample

restingbull:
Ah. We've been release blocked because we all thought we'd mucked up our dev boxes.

Bencodes (approved):

Kernald:
I'm not sure why `jarjar_runner` fails to run on CI, as it runs perfectly on my machine in the same context.

Kernald:
One of the issues raised here is fixed by #530.

cgruber:
Hey @Kernald - mind resolving this or rebasing against master, and we can look again and see what's going wrong?

Kernald:
I'll rebase and update as necessary tomorrow or Monday and ping here once it's done!

Kernald:
@cgruber This is now rebased onto `master`, with a bit of clean-up to make buildifier happy, and fails with the same error in CI (it builds just fine locally on my machine).

I also tried to update Compose to its latest (1.0.0-rc01) version, which starts failing locally (most likely due to https://github.com/bazelbuild/bazel/issues/13553, not related to rules_kotlin afaik).

Bencodes:
@Kernald I opened up a new PR with everything fixed over here https://github.com/bazelbuild/rules_kotlin/pull/657

Kernald:
Thanks, I'll have a look later this weekend/early next week!

restingbull:
Superseded by  #657

Issue #494, opened 2021-02-23
Multiplex worker implementation limits parallelism


jdai8:
The current multiplex worker implementation (at HEAD) sequences reads and writes: https://github.com/bazelbuild/rules_kotlin/blob/51fe50823ced6c01d4fe1d4e9bf37b6a42775052/src/main/kotlin/io/bazel/worker/PersistentWorker.kt#L83-L113

I think this artificially limits performance in certain cases. For example, if we receive a large work request first, all subsequent work requests will have to wait for it to finish before we write their work responses to stdout.

I tested this on a simple reproducer project here: https://github.com/jdai8/rules_kotlin_coroutine_repro. This project has one large source file `Big.kt` and several small `SmallX.kt` files. As you can see from this screenshot, `//src:small` and `//src:small4` have to wait for `//src:big` to finish, even though the other small compilation actions have already finished (in <5s).

![image](https://user-images.githubusercontent.com/13922268/108878099-e6736600-75cd-11eb-9195-2651d87fad2e.png)

Let me know if I'm understanding this correctly. If so, I'm happy to put up a PR.

jongerrish:
Fascinating, great detective work! Want to put up a PR and I can test against our repo (multiplex workers performance regressed so I'm super keen to try your fix)

jdai8:
@jongerrish If enabling multiplex workers was the only change at play for you, I think this shouldn't regress from normal persistent workers - at worst you'd be performing the same (doing everything in serial).

That being said, I'll try to have a PR up in the next few days so you can test it out.

jongerrish:
Well regular persistent workers we give like 8 max instances so 8 processes for the `KotlinCompile` mnemonic so we'd have 8 parallel Kotlinc jobs running at a time without multiplexed-workers.

jdai8:
Yes, I meant within each persistent worker work would happen serially. There might be some other behavior at play - maybe Bazel tries to schedule more work on a worker if it's known to support multiplexing, or something like that.

jongerrish:
Yeah, that makes sense. There should be just a single process for `KotlinCompile` and I'm not sure if there are any upper bounds for max number of concurrent requests Bazel will schedule at a time, but it makes sense that Bazel would schedule more work on that process.

jeffzoch:
@jdai8 how did you generate that graph out of curiosity? Would like to have that tool in my tool belt :)

jdai8:
[...]

PR #495, opened 2021-02-25
Parallelize multiplex work execution

jdai8:
Fixes https://github.com/bazelbuild/rules_kotlin/issues/494.

See the Java compilation implementation: https://github.com/bazelbuild/bazel/blob/d61d394107f46ad27509b339741a45f4c7205965/src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java#L77

### DIFF
--- a/src/main/kotlin/io/bazel/worker/PersistentWorker.kt
+++ b/src/main/kotlin/io/bazel/worker/PersistentWorker.kt
@@ -80,18 +80,19 @@ class PersistentWorker(
   override fun start(execute: Work) = WorkerContext.run {
     captureIO().use { io ->
       BlockableDispatcher.runIn(coroutineContext) {
+        val workExecutor = Executors.newCachedThreadPool()
         blockable {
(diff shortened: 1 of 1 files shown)


jdai8:
cc: @jongerrish

jongerrish:
Thanks for the fix Jack, just testing this out on our repo now.

jongerrish:
I ran a suite of incremental change tests on our repo and saw that performance for each one is at least as good as without multiplex workers enabled. Some targets which trigger a lot of Kotlin actions show pretty significant performance improvements, up to 20% in some cases! Thanks for the quick fix!

restingbull:
I fixed the coroutine usage in https://github.com/bazelbuild/rules_kotlin/pull/496

Thank you for the bugfix and issue!

restingbull:
(My preference is to use coroutines for the moment -- until the profiler tells me it's a bad idea.)

PR #496, opened 2021-02-25
Use flowOn to make multiplexing actually parallel.

restingbull:
Added simple multiplex example from @jdai8.

Issue: 494


jongerrish (approved):

src/main/kotlin/io/bazel/worker/PersistentWorker.kt
```diff
@@ -83,6 +84,7 @@ class PersistentWorker(
         blockable {
           generateSequence { WorkRequest.parseDelimitedFrom(io.input) }
         }.asFlow()
+          .flowOn(coroutineContext)
```
jdai8, line 87: Does this work if we're not actually launching any new coroutines?

src/main/kotlin/io/bazel/worker/PersistentWorker.kt
```diff
@@ -83,6 +84,7 @@ class PersistentWorker(
         blockable {
```
jongerrish, line 84: unblockedContext on line 60 is unused, is that intentional?

restingbull: For now, yeah.

PR #498, opened 2021-03-08
Fix concurrency for multiplex

restingbull:
* Tucked it behind a flag
* Ignore compose at present for releases.


jongerrish (approved):
Patched this is and works on our repo.

Build seems to be about 20% slower than without multiplex workers but since this is behind a flag I think there is value in merging to allow us to profile and improve it.

PR #501, opened 2021-03-11
Parallelize stdout and clean up PersistentWorker

jeffzoch:
For discussion in #494

### DIFF
--- a/src/main/kotlin/io/bazel/worker/PersistentWorker.kt
+++ b/src/main/kotlin/io/bazel/worker/PersistentWorker.kt
@@ -29,7 +29,12 @@ import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.buffer
 import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.flowOn
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
+import java.io.PrintStream
 import java.nio.charset.StandardCharsets.UTF_8
 import java.util.concurrent.Executors
 import kotlin.coroutines.CoroutineContext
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/bazel/worker/PersistentWorker.kt
```diff
@@ -50,81 +60,66 @@ class PersistentWorker(
...
-        return@runBlocking action()
-      }
-    }
-  }
+  val scope = CoroutineScope(Job() + Dispatchers.IO)
 
   @ExperimentalCoroutinesApi
   override fun start(execute: Work) = WorkerContext.run {
+    //Use channel to serialize writing output
+    val writeChannel = Channel<WorkerProtocol.WorkResponse>(UNLIMITED)
```
jeffzoch, line 68: Main idea is to launch the producer in its own coroutine and have the collector run to completion synchronously. The producer does the compilation in parallel and emits results across a channel which is serially read from. Flows are serial by default so we don't need to synchronize access on the consumer side.

src/main/kotlin/io/bazel/worker/PersistentWorker.kt
```diff
@@ -50,81 +60,66 @@ class PersistentWorker(
...
-        blockable {
+      runBlocking {
+        launch {
           generateSequence { WorkRequest.parseDelimitedFrom(io.input) }
-        }.asFlow()
-          .map { request ->
-            info { "received req: ${request.requestId}" }
-            async {
-              doTask("request ${request.requestId}") { ctx ->
+            .asFlow()
```
jdai8, line 73: Is this flow necessary? Would it be simpler to do something like:
```
generateSequence { WorkRequest.parseDelimitedFrom(io.input) }.forEach {
  launch(Dispatchers.DEFAULT) {
    // do the work
    writeChannel.send(response)
  }
}

jeffzoch: @jdai8 yeah that would totally work too - I didn't want to rock the boat too much but I can change it to that as its easier to reason about.

jeffzoch: Just went ahead and changed to just use coroutines (very similar to your thread implementation now just using kotlin primitives I guess. I'd think we would need to be compiling thousands of files to *maybe* even see a perf difference)

restingbull (approved):

restingbull:
Much cleaner, thank you. 

Also educational. 👍

PR #502, opened 2021-03-12
Introduce ktlint_test and ktlint_fix

shs96c:
These rules are the first step in providing a lightweight way of
ensuring projects use consistent formatting.


restingbull:
Add the packaging bits to the lint dir. You will still need to add a BUILD.release.bazel, and the package to the deps.

I test changes locally by running script/release.sh

restingbull (approved):

PR #523, opened 2021-04-10
Support srcjars in mixed Java/Kotlin sources when ABI jars are enabled

arunkumar9t2:
When srcjars are used in a mixed Java/Kotlin target, build fails since `srcjars` are not propagated to `Java` compiler in the JavaBuilder flow. The error only surfaces when any `*.java` file in target contains references to sources inside said `srcjar`.

Fixed by passing srcjars to Java compiler.


arunkumar9t2:
@googlebot  I fixed it.

arunkumar9t2:
Anything I can do to help merge this PR? This is blocking for us at the moment. 

I can add a test case in one of the examples to verify this case. Dagger example might make sense since it already uses `.srcjars`. Thoughts appreciated.

cgruber:
Looking at it now.

cgruber (approved):
LGTM

PR #525, opened 2021-04-13
Pass tags to android sandwich

Bencodes:


kotlin/internal/jvm/android.bzl
```diff
@@ -57,6 +59,7 @@ def _kt_android_artifact(
         visibility = ["//visibility:private"],
         kotlinc_opts = kotlinc_opts,
         javac_opts = javac_opts,
+        tags = tags,
```
Bencodes, line 62: Not passing the tags here prevents people from conditionally disabling Kotlin ABI compilation via the `tags` attribute

PR #528, opened 2021-04-14
Separate loading repositories from things that might use those downloads

shs96c:
The `kotlin_repositories` macro is designed to download the third
party dependencies that rules_kotlin has for itself. Because of the
way that Bazel works, these must first be downloaded before they can
be referred to by build files.


shs96c:
Are there any changes that you'd like to see to this PR?

cgruber (changes requested):
This looks conceptually fine. I'm sorry, merging more recent stuff has now caused a merge conflict. I'm happy to merge this once those are fixed. If you'd like, I can probably just fix them myself right in this PR.

restingbull:
Resolved by https://github.com/bazelbuild/rules_kotlin/pull/564

PR #530, opened 2021-04-16
Fix kotlin compiler_release not being set

Bencodes:
The Kotlin compiler release wasn't taking despite passing in a `compiler_release`

### DIFF
--- a/kotlin/internal/repositories/repositories.bzl
+++ b/kotlin/internal/repositories/repositories.bzl
@@ -29,4 +29,4 @@ def kotlin_repositories(compiler_release = KOTLIN_CURRENT_COMPILER_RELEASE):
         compiler_release: (internal) dict containing "urls" and "sha256" for the Kotlin compiler.
     """
     kt_configure()
-    _release_kotlin_repositories(compiler_release)
+    _release_kotlin_repositories(compiler_release = compiler_release)


Kernald:
Looking at this, it looks like the issue is caused by the first argument here:
https://github.com/bazelbuild/rules_kotlin/blob/2e18e79f0b409c10a8d914dc3fff10e8a7d8756f/kotlin/internal/repositories/release_repositories.bzl#L41

which, on top of having a typo, is never actually used. Should the hard-coded `_KT_COMPILER_REPO` just below:
https://github.com/bazelbuild/rules_kotlin/blob/2e18e79f0b409c10a8d914dc3fff10e8a7d8756f/kotlin/internal/repositories/release_repositories.bzl#L50

be updated to use this `compiler_repository_name`? This would prevent this bug from happening (in some way), as the two arguments are different types (dict vs string).

jeffzoch:
@restingbull does this fix releases? They were blocked on this issue right?

restingbull:
> @restingbull does this fix releases? They were blocked on this issue right?
Unfortunately, no. this only affects the work on bootstrapping kotlin with kotlin.

I pushed an ignore for the compose example (the breaking part), so I can roll up a release.

PR #534, opened 2021-04-25
Add buildifier lint to ci

restingbull:


cgruber (approved):
Heartily approved. 

I assume the plan is to turn on selective warnings as errors as we go? (Stuff like duplicate names seems useful)

cgruber:
Some failing tests in the examples for some reason.

Issue #536, opened 2021-04-30
kt_android_local_test can't depend on testonly libraries


Nikolas-LFDesigns:
Trying to use a newly-introduced kt_android_local_test in my project. Can't use testonly dependencies on it, since the rule is not testonly by itself.
Error message is as follows: 
```
in android_library rule //testproject/src/test/java/com/test:LocalTest_base: non-test target '//testproject/src/test/java/com/test:LocalTest_base' depends on testonly target '@com_google_truth_truth//:com_google_truth_truth' and doesn't have testonly attribute set
in android_library rule //testproject/src/test/java/com/test:LocalTest_base: non-test target '//testproject/src/test/java/com/test:LocalTest_base' depends on testonly target '@org_mockito_mockito_core//:org_mockito_mockito_core' and doesn't have testonly attribute set
in android_library rule //testproject/src/test/java/com/test:LocalTest_base: non-test target '//testproject/src/test/java/com/test:LocalTest_base' depends on testonly target '@junit_junit//:junit_junit' and doesn't have testonly attribute set
```

Minimal example omitting standard stuff like rule definitions and basic trivial stuff:
**WORKSPACE:**
```
java_import_external(
    name = "com_google_truth_truth",
    jar_sha256 = "4868998229b155dcbe68416803975d491704f7cdb6847346cf9a49f65067d6fd",
    jar_urls = [
        "https://repo1.maven.org/maven2/com/google/truth/truth/0.40/truth-0.40.jar",
    ],
    testonly_ = True,
    deps = [
        "@com_google_errorprone_error_prone_annotations",
        "@com_google_guava_guava_android",
        "@com_googlecode_java_diff_utils_diffutils",
    ],
)

java_import_external(
    name = "junit_junit",
    jar_sha256 = "59721f0805e223d84b90677887d9ff567dc534d7c502ca903c0c2b17f05c116a",
    jar_urls = [
        "https://repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar",
    ],
    testonly_ = True,
    deps = [
        "@org_hamcrest_hamcrest_core",
    ],
)
```
**Local.kt:**
```
package com.test

class Local {
  fun doSomething(): Int = 0
}
```
**testproject/src/main/java/com/test/BUILD.bazel:**
```
kt_android_library(
    name = "test",
    srcs = [
        "Local.kt",
    ],
)
```
**LocalTest.kt:**
```
package com.test

@RunWith(JUnit4::class)
[...]

PR #537, opened 2021-04-30
Fix kt_android_local_test can't depend on testonly libraries

Bencodes:
https://github.com/bazelbuild/rules_kotlin/issues/536

### DIFF
--- a/kotlin/internal/jvm/android.bzl
+++ b/kotlin/internal/jvm/android.bzl
@@ -28,6 +28,7 @@ def _kt_android_artifact(
         kotlinc_opts = None,
         javac_opts = None,
         enable_data_binding = False,
+        testonly = False,
         tags = [],
         **kwargs):
     """Delegates Android related build attributes to the native rules but uses the Kotlin builder to compile Java and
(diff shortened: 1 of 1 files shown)


kotlin/internal/jvm/android.bzl
```diff
@@ -28,6 +28,7 @@ def _kt_android_artifact(
         kotlinc_opts = None,
         javac_opts = None,
         enable_data_binding = False,
+        testonly = False,
```
Kernald, line 31: Not a big deal, but I don't think adding this is needed - both the native `android.library` and the `_kt_jvm_library` fetch it from the `kwargs` anyway. The only change really needed here is the one on line 101.

Nikolas-LFDesigns: Tried the suggested fix to line 101, works as intended. Is there any way to speed up the process of applying these changes (like contributing to a subject PR)?

Nikolas-LFDesigns (changes requested):

Nikolas-LFDesigns:
@googlebot I fixed it.

Nikolas-LFDesigns:
@cgruber @jin @restingbull @timpeut 
al checks passed, could you please check out the final solution?

cgruber (approved):
LGTM

Issue #281, opened 2020-02-17
Cannot include generated file as a resource


rockwotj:
Using the latest commit from master, this fails:
```
genrule(
    name = "generated_logging_properties",
    srcs = [":local.logging.properties"],
    outs = ["logging.properties"],
    cmd = "cp $< $@",
)

kt_jvm_library(
    name = "logging",
    srcs = glob(["*.kt"]),
    resources = [":generated_logging_properties"],
    visibility = ["//visibility:public"],
    deps = [
        "@maven//:com_google_flogger_flogger",
        "@maven//:com_google_guava_guava",
        "@maven//:com_google_flogger_flogger_system_backend",
        "@maven//:org_slf4j_slf4j_jdk14",
    ],
)
```

```
ERROR: /Users/rockwotj/Workspace/foo/kotlin/logging/BUILD:10:1: error executing shell command: '/bin/bash -c external/bazel_tools/tools/zip/zipper/zipper c bazel-out/darwin-fastbuild/bin/kotlin/logging/logging-resources.jar @bazel-out/darwin-fastbuild/bin/kotlin/logging/logg...' failed (Exit 255) bash failed: error executing command 
  (cd /private/var/tmp/_bazel_rockwotj/70f6e107f8f90b5b2d301b4029134ae9/sandbox/darwin-sandbox/12/execroot/__main__ && \
  exec env - \
  /bin/bash -c 'external/bazel_tools/tools/zip/zipper/zipper c bazel-out/darwin-fastbuild/bin/kotlin/logging/logging-resources.jar @bazel-out/darwin-fastbuild/bin/kotlin/logging/logging_resources_zipper_args')
Execution platform: @local_config_platform//:host

Use --sandbox_debug to see verbose messages from the sandbox
File kotlin/logging/logging.properties does not seem to exist.Target //kotlin/logging:logging failed to build

```

restingbull:
Try compiling `//kotlin/logging:generated_logging_properties ` directly. It looks like the genrule is not building.

rockwotj:
@restingbull the genrule does build correctly and it's output is in `bazel-genfiles`

restingbull:
Investingating

rockwotj:
If it would help I can create a small sample repo demoing this behavior

rockwotj:
Looks related: https://github.com/bazelbuild/rules_kotlin/pull/187

sugarmanz:
[...]

PR #538, opened 2021-05-13
Fix generated resources

jdai8:
Fixes https://github.com/bazelbuild/rules_kotlin/issues/281.

The issue is that we were using `short_path`, which excludes the configuration root. Using `file.path` instead fixes this. Note that we still want to use `short_path` when deciding where to place the file in the jar. We don't want resources in the jar to look like `bazel-bin/darwin-fastbuild/...`.

Tested this on our repo internally and it seems to work.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -154,14 +154,14 @@ def _adjust_resources_path_by_strip_prefix(path, resource_strip_prefix):
         fail("Resource file %s is not under the specified prefix to strip" % path)
 
     clean_path = path[len(resource_strip_prefix):]
-    return resource_strip_prefix, clean_path
+    return clean_path
 
 def _adjust_resources_path_by_default_prefixes(path):
     for cp in _CONVENTIONAL_RESOURCE_PATHS:
         dir_1, dir_2, rel_path = path.partition(cp)
         if rel_path:
-            return dir_1 + dir_2, rel_path
-    return "", path
+            return rel_path
+    return path
 
 def _adjust_resources_path(path, resource_strip_prefix):
     if resource_strip_prefix:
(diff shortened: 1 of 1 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -266,14 +266,13 @@ def _fold_jars_action(ctx, rule_kind, toolchains, output_jar, input_jars, action
...
-        c_dir, res_path = _adjust_resources_path(f.short_path, ctx.attr.resource_strip_prefix)
+        res_path = _adjust_resources_path(f.short_path, ctx.attr.resource_strip_prefix)
         target_path = res_path
         if target_path[0] == "/":
             target_path = target_path[1:]
-        line = "{target_path}={c_dir}{res_path}\n".format(
-            res_path = res_path,
+        line = "{target_path}={f_path}\n".format(
             target_path = target_path,
-            c_dir = c_dir,
```
jdai8, line 276: This was the only usage of `c_dir`. Since we don't need it anymore I removed it from the return values as well.

kotlin/internal/jvm/compile.bzl
```diff
@@ -266,14 +266,12 @@ def _fold_jars_action(ctx, rule_kind, toolchains, output_jar, input_jars, action
...
     res_cmd = []
     for f in ctx.files.resources:
-        c_dir, res_path = _adjust_resources_path(f.short_path, ctx.attr.resource_strip_prefix)
-        target_path = res_path
+        target_path = _adjust_resources_path(f.short_path, ctx.attr.resource_strip_prefix)
         if target_path[0] == "/":
             target_path = target_path[1:]
-        line = "{target_path}={c_dir}{res_path}\n".format(
-            res_path = res_path,
+        line = "{target_path}={f_path}\n".format(
```
freetheinterns, line 272: Thanks for this change @jdai8! I was so confused why zipping was failing for some resource files generated under bazel-out.

To be super clear this argument line `{target_path}={c_dir}{res_path}` is saying that the file located at `{c_dir}{res_path}` should be zipped under `{target_path}` for the resource jar. So moving to `{target_path}={f_path}` should be strictly just as correct, because `f.path` should always be how input files are located by a bazel action and `{target_path}` remains the same.

cgruber (approved):
This is subtle. Thanks for figuring it out.

PR #541, opened 2021-05-19
Add the plugins examples to run in CI

Kernald:
This is another example that was never built on CI (and didn't actually build, missing a load call for `kt_android_library`). Once they all build properly on CI, it could be interesting to write a script building and testing all examples in the examples directory, rather than having to add an entry in the `presubmit.yml` (which seems to be forgotten quite often).


cgruber (approved):
I did a bit of merge conflict resolution and tidying. This should be fine, assuming we don't have more CI issues (lint or otherwise)

cgruber:
Ugh, sorry - once I fixed merge conflicts, some other issues came loose, so I had to fix those. I roughly followed the WORKSPACE pattern for the android example. Normally I wouldn't edit someone else's PR so much, but doing the conflict resolution backed me into a corner.

Anyway, it's all good, and I'll merge.

Kernald:
No worries, thanks for cleaning all that up!

PR #544, opened 2021-05-24
Kotlin 1.5 support

Bencodes:
This PR includes:
- Support for Kotlin 1.5.x API and language levels
- Compiles `rules_kotlin` using 1.5.10
- Updates `validate_kotlinc_options` to fail fast when the new IR Kotlinc compiler flags that are enabled by default in 1.5 are passed in.

PR #547, opened 2021-06-07
Add support for 1.5 api_version and language_version

nkoroste:
Enables the ability to build with Kotlin compile v1.5


Bencodes:
https://github.com/bazelbuild/rules_kotlin/pull/544

cgruber:
This duplicates #544, which was earlier, so I'll close this in favor of that earlier PR.

jongerrish (commented):
If we suspect that 1.4 compiler is not compatible with multiplex-workers but 1.5 is, then we could consider automatically disabling multiplex-workers if < 1.5 is specified?

It presence/absence is specified in toolchains.bzl here:

"supports-multiplex-workers": "1" if ctx.attr.experimental_multiplex_workers else "0",

PR #551, opened 2021-06-08
Add configuration for stale bot

alexeagle:
Should help with issue/PR triage.

Ideally we would have the resources to address and resolve all issues, and review and merge all pull requests. But we don't.
Rather than leave these open indefinitely, it's better to set user expectations that these will just age out.

This is the same config used in rules_nodejs, rules_docker, and rules_python.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,59 @@
+# See https://github.com/marketplace/actions/close-stale-issues
+
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  # run at 21:30 UTC daily
+  - cron: "30 21 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+        # NB: We start with very long duration while trimming existing issues,
+        # with the hope to reduce when/if we get better at keeping up with user support.
+
+        # The number of days old an issue can be before marking it stale.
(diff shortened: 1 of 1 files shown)


alexeagle:
Note: I think someone with permissions needs to create the "Can Close?" label on this repo before the GH action runs

.github/workflows/stale.yml
```diff
@@ -0,0 +1,59 @@
...
+        # The number of days old an issue can be before marking it stale.
+        days-before-stale: 180
+        # Number of days of inactivity before a stale issue is closed
+        days-before-close: 30
+
+        # If an issue/PR is assigned, trust the assignee to stay involved
+        # Can revisit if these get stale
+        exempt-all-assignees: true
+        # Issues with these labels will never be considered stale
+        exempt-issue-labels: "need: discussion,cleanup"
```
cgruber, line 31: Are these exact, or are they a match? We have labels more like this: `type: cleanup` I don't think these exempt labels would ever trigger on this project.

cgruber (changes requested):
Love it in general, one quick question before merging.

cgruber:
> Note: I think someone with permissions needs to create the "Can Close?" label on this repo before the GH action runs

Assuming we take this PR (and I'm inclined to, once we figure out the other issues with specific labels) I"ll add it.

.github/workflows/stale.yml
alexeagle: I assume they are exact, we could check the specification. I'm on vacation at the moment, feel free to amend or add commits.

cgruber:
Gotcha. I'll take care of it.

PR #558, opened 2021-07-20
Call System.gc between work requests

Bencodes:
These long running Kotlin workers are consuming lots of memory, eventually causing the Linux out-of-memory killer to step in and kill a random java process resulting in this error message:

Example crash output (the out file is empty):
```console
--
  | Server terminated abruptly (error code: 14, error message: 'Socket closed', log file: '/root/.cache/bazel/_bazel_root/7b7747ec045ae606eb720a1222f56098/server/jvm.out')
  |  
```

Calling `System.gc` between these Kotlin work requests has improved stability for us, and it seems to be the standard in Bazel core (seen [here](https://github.com/lyft/bazel/blob/6d1b9725b1e201ca3f25d8ec2a730a20aab62c6e/src/test/java/com/google/devtools/build/lib/worker/ExampleWorkerMultiplexer.java#L148) and [here](https://github.com/bazelbuild/bazel/blob/3835d9b21ad524d06873dfbf465ffd2dfb635ba8/src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java#L468)).

This implementation is intended to closely mirror what Bazel is doing inside it's worker implementation seen [here](https://github.com/Bencodes/bazel/blob/3835d9b21ad524d06873dfbf465ffd2dfb635ba8/src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java#L431-L474).

### DIFF
--- a/src/main/kotlin/io/bazel/worker/CpuTimeBasedGcScheduler.kt
+++ b/src/main/kotlin/io/bazel/worker/CpuTimeBasedGcScheduler.kt
@@ -1,0 +1,41 @@
+package io.bazel.worker
+
+import com.sun.management.OperatingSystemMXBean
+import java.lang.management.ManagementFactory
+import java.time.Duration
+import java.util.concurrent.atomic.AtomicReference
+
+class CpuTimeBasedGcScheduler(
+  /**
+   * After this much CPU time has elapsed, we may force a GC run. Set to [Duration.ZERO] to
+   * disable.
+   */
+  private val cpuUsageBeforeGc: Duration
+) {
+
+  /** The total process CPU time at the last GC run (or from the start of the worker).  */
+  private val cpuTime: Duration
+    get() = if (cpuUsageBeforeGc.isZero) Duration.ZERO else Duration.ofNanos(bean.processCpuTime)
+  private val cpuTimeAtLastGc: AtomicReference<Duration> = AtomicReference(cpuTime)
+
+  /** Call occasionally to perform a GC if enough CPU time has been used.  */
+  fun maybePerformGc() {
+    if (!cpuUsageBeforeGc.isZero) {
+      val currentCpuTime = cpuTime
+      val lastCpuTime = cpuTimeAtLastGc.get()
+      // Do GC when enough CPU time has been used, but only if nobody else beat us to it.
+      if (currentCpuTime.minus(lastCpuTime).compareTo(cpuUsageBeforeGc) > 0
+        && cpuTimeAtLastGc.compareAndSet(lastCpuTime, currentCpuTime)
+      ) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/bazel/worker/PersistentWorker.kt
```diff
@@ -73,6 +73,8 @@ class PersistentWorker(
               .forEach { request ->
                 launch {
                   compileWork(request, io, writeChannel, execute)
+                  //Be a friendly worker by performing a GC between compilation requests
+                  System.gc()
```
Bencodes, line 77: Mirroring what's being done in the generic worker implementation inside of Bazel might actually be a better approach here that doesn't incur too much gc overhead. I can make those changes if folks think this is the right direction.

https://github.com/Bencodes/bazel/blob/3835d9b21ad524d06873dfbf465ffd2dfb635ba8/src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java#L431-L474

rockwotj: I've found in other systems calling System.gc frequently only hurts instead of helping. Calling periodically would certainly be better, but is there other JVM GC settings that can help here instead of manually invoking GC?

jeffzoch: I am curious why bazel has decided to call System.gc as well - id think tuning the GC a bit more worthwhile. That said the approach you linked @Bencodes seems like a decent compromise.

src/main/kotlin/io/bazel/worker/CpuTimeBasedGcScheduler.kt
```diff
@@ -0,0 +1,41 @@
+package io.bazel.worker
+
+import com.sun.management.OperatingSystemMXBean
+import java.lang.management.ManagementFactory
+import java.time.Duration
+import java.util.concurrent.atomic.AtomicReference
+
+class CpuTimeBasedGcScheduler(
```
Bencodes, line 8: Mirroring what's being done inside the generic Bazel worker

https://github.com/Bencodes/bazel/blob/3835d9b21ad524d06873dfbf465ffd2dfb635ba8/src/main/java/com/google/devtools/build/lib/worker/WorkRequestHandler.java#L431-L474

rockwotj: It maybe worthwhile for future readers to inline the link to bazel source and part of the PR description in a class comment :)

rockwotj (approved):

restingbull (approved):

Issue #531, opened 2021-04-20
Targets can get out of date


Kernald:
I'm not 100% sure of when or why it happens, and can't consistently reproduce it, but using the rules at 8d987688eb74d025d24106fbbfba57d8aa5e48dc sometimes results in targets being used without being recompiled. One example I encountered:

Target A depends on B. B contains a sealed class, with a simple list of sub-objects. A has a `when` statement on an instance of this sealed class. I add a subclass to this sealed class hierarchy in B, and update the `when` statement in A. IntelliJ is happily claiming that everything is fine - as intended. Running a bazel build or bazel test, on the other hand, will result in a compilation error in A's `when` statement, as according to the compiler, the new subclass doesn't exist. Changing B's implementation from there doesn't do anything - I have to run `bazel clean` to discard the faulty cache (I assume that's what's happening).

I just encountered it again and sadly ran `bazel clean` before thinking of investigating it a bit more in depth, next time it happens I'll try running a `bazel build //:B` and see if Bazel claims it's up-to-date, despite having changes, and inspect the jar as well, but I've definitely encountered this a few times since updating to 8d987688eb74d025d24106fbbfba57d8aa5e48dc (from the prebuilt v1.5.0-alpha-3).

Kernald:
Update on this, which I just encountered again: `bazel build //:B` claims that the target is up-to-date which, looking at the content of the jar, seems correct. But for some reason, that jar is _not_ used to build A.

Bencodes:
We just encountered something very similar in our repo. Same setup where target A depends on and references Kotlin sources in target B. Renaming a class (say `Foo.kt` to `Bar.kt`) in module B will result in a compilation failure due to target A not being able to reference `Bar.kt` despite that file existing.

Bencodes:
This looks related https://github.com/bazelbuild/bazel/issues/13135

mgenov:
Similar issue on my side. bazel clean needs to be executed to enforce using of the new jar.

rockwotj:
We also hit this only when we're using persistent workers - does this happen without persistent workers? Additionally a bazel shutdown seems to fix things, instead of a bazel clean. It seems like the worker gets into a n invalid state or something... Is this the case for others? If not I can open a new issue to track

cgruber:
@restingbull - I wonder if the multiplex worker is having an issue with cleaning up or isolating state?

cgruber:
For those seeing this, is it happening with different versions of bazel?

[...]

PR #559, opened 2021-07-20
Revert "Isolate Kotlin compiler instances for Multiplex workers."

rockwotj:
Reverts bazelbuild/rules_kotlin#515

This results in targets being able to get out of date https://github.com/bazelbuild/rules_kotlin/issues/531

I'm assuming that this was enabled due to the work around from https://youtrack.jetbrains.com/issue/KT-45007 @jongerrish? Which seems to be fixed in kotlin 1.5... Or maybe we should only enable this system property for multiplexing mode? I wasn't able to easily figure out how to do this, so I figured that reverting this change until then was a better solution

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
@@ -163,18 +163,16 @@ class KotlinToolchain private constructor(
     toolchain: KotlinToolchain,
     clazz: String
   ) {
+    private val compiler: Any
     private val execMethod: Method
     private val getCodeMethod: Method
-    private val compilerClass = toolchain.classLoader.loadClass(clazz)
 
     init {
-      // Disables disposing of the environment which is required when running parallel builds. This
-      // follows the behavior of the Kotlin compiler daemon.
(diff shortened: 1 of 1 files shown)


restingbull:
*sigh* Willing to bet that this is due to the kotlin compiler not checking content hash but date stamp.

We can't completely revert this change, as the initial fix dealt with cross contamination. Thanks for the debugging!
Going to need to sort out what is happening.

rockwotj:
The cross contamination was only for experimental_multiplex_workers right? It seems undesirable to break the defaults to make an experimental feature work.

rockwotj:
Also I found one place in kotlinc where it caches based on timestamp and file length for zip files, and a couple other places that just looked at timestamps. I'll comment on the bug where I found those caches in kotlinc

restingbull:
Wait... I was under the impression this issue only cropped up with multiplex workers enabled?

rockwotj:
@restingbull sorry to not be clear, but this repros with just the default settings (no experimental flags on) with the latest release.

rockwotj:
You can test if you want by following my steps in https://github.com/bazelbuild/rules_kotlin/issues/531#issuecomment-882654313

restingbull (approved):

PR #566, opened 2021-08-04
Add --exclude_build_data arguement to singlejar invocation

skoeld:
Adding --exclude_build_data so that the build-data.properties won't be created.

This file contains contains information about the build host and will therefore break cacheing
for users that are not on the same platform as platform that built the cached artefact.

build-data.properties is not present in java artefacts since this flag is already applied so with this change, kotlin artefacts will have the same behaviour as java artifacts.


skoeld:
@googlebot I signed it!

cgruber:
Thanks a lot!

PR #587, opened 2021-09-18
Use JavaPluginInfo if available in Bazel

comius:
In Bazel@Head, JavaPluginInfo provides the same data as kt_jvm_plugin_aspect collects.

The change uses it where JavaPluginInfo is available (new Bazel versions) and stops propagating the aspect.

Incompatible part is, that the java_plugins in deps are not allowed, following java_library pattern. Otherwise if plugins are exported in java_library which is in the deps, they are still collected.

The change is otherwise backwards compatible.

Fixes https://github.com/bazelbuild/rules_kotlin/issues/592


.bazelversion
```diff
@@ -1 +0,0 @@
-4.2.0
```
Bencodes, line 1: Should this actually be pinned so that CI doesn't spontaneously break?

Bencodes (approved):

.bazelversion
comius: I removed it by accident, while testing.

restingbull (approved):

PR #588, opened 2021-09-18
Update Kotlin to 1.5.30

Bencodes:
Tracking the latest Kotlin 1.5.30 release and removing the deprecated plugin encoding `configuration` CLI usage.


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationArgs.kt
```diff
@@ -172,28 +169,18 @@ class CompilationArgs(
     return this
   }
 
-  fun list(): List<String> = args.toList()
-
-  fun base64Encode(
+  fun flagRepeated(
```
restingbull, line 172: The undocumented format changed? Oh, surprise....

This will need to be pulled into the versioned repos. Lmk if you'd prefer to pass this off to me, as I haven't finished writing up how to make the versioned toolchains work.

Bencodes: Sure you can take over.

Have an ETA on when the versioned toolchains will be ready? People won't be able to update Kotlin and likely Compose until this is merged.

jeffzoch: @Bencodes any idea if this works for 1.6-M1? https://github.com/JetBrains/kotlin/releases/tag/v1.6.0-M1 If we can stay on top of the feature previews we can hopefully get 1.6 support out in a timely manner.

Bencodes: @jeffzoch In theory yes but I haven't tried yet. We are waiting on some important fixes in 1.6.0 so i'll be looking at supporting that soon.

nkoroste:
@Bencodes @restingbull What do you guys think the status of this PR? are we planning to shoot for 1.6 or get this merged in the meantime so that we can start using 1.5.30?

Kernald:
Having this before 1.6 would be quite nice given that Compose now depends on Kotlin 1.5.30.

restingbull:
Hoping to take care of this today. Been insane on my side.

PR #600, opened 2021-10-14
add test for sealed class

thomasbao12:


restingbull:
Repro for https://github.com/bazelbuild/rules_kotlin/issues/638

PR #601, opened 2021-10-15
Pass compilation args using the official way

Bencodes:
Pulling this portion of the Kotlin 1.5.30 support out of the original PR. We should be able to merge this independently so that people can use Kotlin 1.5.30 without rules_kotlin having it as the default.

Original PR here: https://github.com/bazelbuild/rules_kotlin/pull/588

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationArgs.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationArgs.kt
@@ -18,12 +18,9 @@
 package io.bazel.kotlin.builder.tasks.jvm
 
 import io.bazel.kotlin.builder.toolchain.KotlinToolchain
-import java.io.ByteArrayOutputStream
-import java.io.ObjectOutputStream
 import java.nio.file.FileSystem
 import java.nio.file.FileSystems
 import java.nio.file.Path
-import java.util.Base64
 
 /**
  * CompilationArgs collects the arguments for executing the Kotlin compiler.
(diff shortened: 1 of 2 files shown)


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationArgs.kt
```diff
@@ -172,28 +169,18 @@ class CompilationArgs(
...
+        flag(flag, transform(option, it))
       }
     }
-
-    oos.flush()
-    flag(flag, transform(Base64.getEncoder().encodeToString(os.toByteArray())))
     return this
   }
+
+  fun list(): List<String> = args.toList()
```
cgruber, line 185: Probably should stick with the toList() naming pattern.

cgruber (commented):
This seems fine to me, but I want @restingbull to confirm it's not going to clobber or interfere with his other efforts around 1.5.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationArgs.kt
restingbull: Can't just merge it as we needed the base64 encoding for older versions of kotlinc. Currently working on a way to preserve versionsing without making too much of a mess.

restingbull: Added branch to cover pre-1.5

Not the way I would like it, but I don't want to hold up progress while I yak shave with one hand.

PR #607, opened 2021-10-28
Add x_explicit_api_mode to remaining kotlinc_options

pettermahlen:
See https://github.com/Kotlin/KEEP/blob/master/proposals/explicit-api-mode.md for more about the switch itself. 

I'm not sure how to test this - the best test would be to have some test code with default access modifiers and explicit-api=strict set, and then run a compilation and verify that it fails. That seems complex, and like it would require a fair amount of tooling, which I'm not sure if it's available. Please advice if so.

### DIFF
--- a/src/main/starlark/rkt_1_5/kotlin/opts.bzl
+++ b/src/main/starlark/rkt_1_5/kotlin/opts.bzl
@@ -88,6 +88,19 @@ _KOPTS = {
             True: ["-Xno-optimized-callable-reference"],
         },
     ),
+    "x_explicit_api_mode": struct(
+        args = dict(
+            default = "off",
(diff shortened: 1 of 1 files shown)


Bencodes:
Ah so the solution here is to replicate the existing `x_explicit_api_mode` option in the newly added opts files that are missing this explicit api mode attribute. Do you want to update your PR to do that?

pettermahlen:
Done. AFAICT, it was only the 1.5 file that was missing the option.

Bencodes (approved):

Bencodes:
@cgruber or @restingbull will need to merge this

restingbull (approved):

PR #611, opened 2021-11-04
Fix JavaPluginInfo in Kotlin Android and Plugin examples.

comius:
Changes to JavaPlugin handling cause failures with Bazel@HEAD,

https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2237#9b9caec9-2b90-4c31-8f38-d3086cc37e73

PR #614, opened 2021-11-05
Remove overzealous provider requiring.

restingbull:
Signed-off-by: restingbull@cabaretmechanique.com <restingbull@cabaretmechanique.com>


Bencodes:
Another PR was opened up here https://github.com/bazelbuild/rules_kotlin/pull/611/files

restingbull:
> Another PR was opened up here https://github.com/bazelbuild/rules_kotlin/pull/611/files

Won't fix the core problem (both compiler plugin which do *not* have `JavaInfoPlugin` _and_ `java_plugin` use the same attribute, but has some optimizations we should merge.

cgruber (approved):

Issue #582, opened 2021-09-07
Inconsistency with Java rules when laying resources out


utikeev:
Kotlin rules (`kt_jvm_library`) do not put resource folders in the resulting JAR, only files, while Java rules (`java_library`) put all folders on the path to the resource files. The minimal example can be found here:
[kotlin-resources-folder-issue.zip](https://github.com/bazelbuild/rules_kotlin/files/7120952/kotlin-resources-folder-issue.zip)

`bazel run //module/src/main:kotlin_loader_with_java_lib` output:
```
INFO: Analyzed target //module/src/main:kotlin_loader_with_java_lib (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //module/src/main:kotlin_loader_with_java_lib up-to-date:
  bazel-bin/module/src/main/kotlin_loader_with_java_lib.jar
  bazel-bin/module/src/main/kotlin_loader_with_java_lib.jdeps
INFO: Elapsed time: 0.197s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
jar:file:/private/var/tmp/_bazel_stanislav.utikeev/febbbb93c0d677bddc787480695e7c67/execroot/__main__/bazel-out/darwin-fastbuild/bin/module/src/main/libjava_lib.jar!/folder/migration_a.sql
jar:file:/private/var/tmp/_bazel_stanislav.utikeev/febbbb93c0d677bddc787480695e7c67/execroot/__main__/bazel-out/darwin-fastbuild/bin/module/src/main/libjava_lib.jar!/folder/migration_b.sql
```

`bazel run //module/src/main:kotlin_loader_with_kotlin_lib` output:
```
INFO: Analyzed target //module/src/main:kotlin_loader_with_kotlin_lib (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //module/src/main:kotlin_loader_with_kotlin_lib up-to-date:
  bazel-bin/module/src/main/kotlin_loader_with_kotlin_lib.jar
  bazel-bin/module/src/main/kotlin_loader_with_kotlin_lib.jdeps
INFO: Elapsed time: 0.231s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
```

[...]

PR #612, opened 2021-11-04
Allow building from head repo

restingbull:
Fixes https://github.com/bazelbuild/rules_kotlin/issues/570

PR #615, opened 2021-11-07
Ensure resource directories exist in jars

restingbull:
Fixes https://github.com/bazelbuild/rules_kotlin/issues/582

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -175,6 +175,7 @@ def _fold_jars_action(ctx, rule_kind, toolchains, output_jar, input_jars, action
         "--normalize",
         "--compression",
         "--exclude_build_data",
+        "--add_missing_directories",
     ])
     args.add_all([
         "--deploy_manifest_lines",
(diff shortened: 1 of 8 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -175,6 +175,7 @@ def _fold_jars_action(ctx, rule_kind, toolchains, output_jar, input_jars, action
         "--normalize",
         "--compression",
         "--exclude_build_data",
+        "--add_missing_directories",
```
cgruber, line 178: Man, I'm confused about this PR... is the fundamental fix adding this flag, or adding the closeable that forces execution? Or is the closeable just do you can use it more conveniently in tests in a `try-with-resources`/`Closeable.use{}`? 

Anyway, I'd laugh/cry if this flag were the fix...  Sigh.

cgruber (approved):

cgruber:
Note, I'll leave merging to you, since this clearly builds on #612

kotlin/internal/jvm/compile.bzl
restingbull: this flag is the fix, but I was adding tests to all of 'em.

The closeable just makes the code read cleaner.

restingbull:
> Note, I'll leave merging to you, since this clearly builds on #612

Unintentionally.

Issue #591, opened 2021-09-22
JdepsMerger doesn't support worker_sandboxing


Bencodes:
With worker sandboxing enabled `depsMerger#readJarOwnerFromManifest` fails with a `NoSuchFileException` while attempting to read manifest information form JARs that aren't linked to the sandbox.

It can be reproduced by building with `--worker_sandboxing=true` and the below `kt_toolchain`:
```python
define_kt_toolchain(
    name = "kotlin_toolchain",
	...
    experimental_reduce_classpath_mode = "KOTLINBUILDER_REDUCED",
    experimental_report_unused_deps = "error",
    experimental_strict_kotlin_deps = "error",
	...
)
```

```console
SEVERE: ERROR: unexpected exception
java.io.UncheckedIOException
        at io.bazel.kotlin.builder.tasks.jvm.JdepsMerger$Companion.readJarOwnerFromManifest(JdepsMerger.kt:56)
        at io.bazel.kotlin.builder.tasks.jvm.JdepsMerger$Companion.merge(JdepsMerger.kt:107)
        at io.bazel.kotlin.builder.tasks.jvm.JdepsMerger.execute(JdepsMerger.kt:146)
        at io.bazel.kotlin.builder.tasks.MergeJdeps.invoke(MergeJdeps.kt:28)
        at io.bazel.worker.PersistentWorker$compileWork$2$result$1.invoke(PersistentWorker.kt:101)
        at io.bazel.worker.PersistentWorker$compileWork$2$result$1.invoke(PersistentWorker.kt:99)
        at io.bazel.worker.WorkerContext$TaskContext.resultOf(WorkerContext.kt:125)
        at io.bazel.worker.WorkerContext.doTask(WorkerContext.kt:153)
        at io.bazel.worker.PersistentWorker$compileWork$2.invokeSuspend(PersistentWorker.kt:99)
        at io.bazel.worker.PersistentWorker$compileWork$2.invoke(PersistentWorker.kt)
        at io.bazel.worker.PersistentWorker$compileWork$2.invoke(PersistentWorker.kt)
        at kotlinx.coroutines.intrinsics.UndispatchedKt.startUndispatchedOrReturn(Undispatched.kt:89)
        at kotlinx.coroutines.BuildersKt__Builders_commonKt.withContext(Builders.common.kt:165)
        at kotlinx.coroutines.BuildersKt.withContext(Unknown Source)
[...]

PR #620, opened 2021-11-16
Filter on jars inside of createDepsMap

Bencodes:
This prevents crashes when `createDepsMap` attempts to resolve dependencies coming from a local java repository.

https://github.com/bazelbuild/rules_kotlin/issues/591

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt
+++ b/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt
@@ -265,7 +265,10 @@ class JdepsGenExtension(
     val jarsToClasses = mutableMapOf<String, MutableList<String>>()
     classes.forEach {
       val parts = it.split("!/")
-      jarsToClasses.computeIfAbsent(parts[0]) { ArrayList() }.add(parts[1])
+      val jarPath = parts[0]
+      if (jarPath.endsWith(".jar")) {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt
```diff
@@ -265,7 +265,10 @@ class JdepsGenExtension(
     val jarsToClasses = mutableMapOf<String, MutableList<String>>()
     classes.forEach {
       val parts = it.split("!/")
-      jarsToClasses.computeIfAbsent(parts[0]) { ArrayList() }.add(parts[1])
+      val jarPath = parts[0]
+      if (jarPath.endsWith(".jar")) {
+        jarsToClasses.computeIfAbsent(jarPath) { ArrayList() }.add(parts[1])
```
Bencodes, line 270: The logic that relies on this is highly reliant on jar stamping to correctly work, which isn't always enabled by default. Not sure what the best way to handle this is.

restingbull: Is the concern on third party jars being unstamped? We could force it in the rules...

Bencodes: Should have been more specific here. This comment was more of a mental note-to-self as we work on jdeps/classpath reduction.

PR #628, opened 2021-11-20
Repo case for class conflicts with compiler and compiler-embeddable

mauriciogg:
running blaze build //repro/src/main/java/com/repro/lib:lib will produce
the following exception:

java.lang.NoSuchMethodError: 'org.jetbrains.kotlin.com.intellij.psi.PsiElement[] org.jetbrains.kotlin.psi.KtClassOrObject.getChildren()'

NOTE: anvil version is updated in WORKSPACE file. previous versions dont
have this issue since they are selfcontained. Starting with this version
anvil was split into 3 jars (compiler, compiler-utils and compiler-api).
The conflict happens in compiler-utils not compiler.

A naive fix is implemented in this commit https://github.com/mauriciogg/rules_kotlin/commit/bdfcaa510ff7b15576322aba04088b91162ef189
it simply deshades all transitive deps


restingbull:
Thanks for doing this! Looking into solutions.

Issue #639, opened 2021-12-03
Shaded plugin ars do not play well with `android_binary`


restingbull:
In deps attribute of `android_binary rule //platform/analytics/sample:analytics-sample-apk`: Dependencies on .jar artifacts are not allowed in Android binaries, please use a java_import to depend on `external/host_mvn/com_google_dagger_dagger_deshaded_processed_dagger-2.37.jar`. If this is an implicit dependency then the rule that introduces it will need to be fixed to account for it correctly.
notice the _deshaded_
`//platform/analytics/sample:analytics-sample-apk` is an android_binary target which fails in this case. However, the kt_android_library  that actually have this dep compiles just fine.
That dagger deps comes from maven which is downloaded via rules_jvm_external which automatically creates this target:
```
jvm_import(
	name = "com_google_dagger_dagger",
	jars = ["v1/gcs/snapengine-maven-publish/releases/com/google/dagger/dagger/2.37/dagger-2.37.jar"],
	deps = [
		":javax_inject_javax_inject",
	],
	tags = ["maven_coordinates=com.google.dagger:dagger:2.37"],
)
```
essentially deshaded jvm_imports for some reason are being considered vanilla jar files which the android rules don't like although they come from a jvm_import

PR #644, opened 2021-12-10
Don't propagate JavaInfo for compiler plugin deshaded libs

nkoroste:
If JavaInfo is propagated, compiler plugin libs end up in the set of runtime dependencies of android_binary targets, which makes no sense.
Fixes https://github.com/bazelbuild/rules_kotlin/issues/639

### DIFF
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -419,7 +419,6 @@ def kt_compiler_plugin_impl(ctx):
 
     return [
         DefaultInfo(files = classpath),
-        java_common.merge([]),
         _KtCompilerPluginInfo(
             classpath = classpath,
             options = options,
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -419,7 +419,7 @@ def kt_compiler_plugin_impl(ctx):
 
     return [
         DefaultInfo(files = classpath),
-        info,
+        java_common.merge([]),
         _KtCompilerPluginInfo(
             classpath = classpath,
             options = options,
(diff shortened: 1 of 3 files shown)


nkoroste:
cc @mauriciogg

kotlin/internal/jvm/impl.bzl
```diff
@@ -419,7 +419,7 @@ def kt_compiler_plugin_impl(ctx):
 
     return [
         DefaultInfo(files = classpath),
-        info,
+        java_common.merge([]),
```
nkoroste, line 422:
```suggestion
```

Bencodes (approved):

kotlin/internal/jvm/impl.bzl
restingbull, line 422: What consumes this JavaInfo?

nkoroste: I couldn't find anything

restingbull: Let's remove it, then.

nkoroste: sorry, updated

nkoroste: Looking closer into it, I see it's referenced in
https://sourcegraph.com/github.com/bazelbuild/rules_kotlin/-/blob/kotlin/internal/jvm/compile.bzl?L707#tab=history
https://sourcegraph.com/github.com/bazelbuild/rules_kotlin/-/blob/kotlin/internal/jvm/plugins.bzl?L51

Do we still want to remove it?

restingbull: Yes -- providing an empty JavaInfo doesn't help, and increases the chances of something stupid happening.

Speaking of stupid... why does the target have a `KtJvmPluginInfo` and no `JavaInfo`? It isn't due to `KtCompilerPluginInfo`.

Maybe it's an export issue?

restingbull (approved):

PR #650, opened 2022-01-02
Add initial test for multiplex worker corruption

restingbull:


src/test/kotlin/io/bazel/kotlin/integration/MultiplexTest.kt
```diff
@@ -0,0 +1,216 @@
...
+                  "fun one"(!"value:String") {
+                    "println"(!"value")
+                  }
+                }
+              }
+            }
+          }
+          1 -> {
+            one {
+              kotlin("One.kt") {
```
Bencodes, line 190: Just writing `One${round}.kt` and then referencing that in `Two.kt1 like `class Two : One${round}` should be enough to repro.

src/test/kotlin/io/bazel/kotlin/integration/WriteWorkspace.kt
```diff
@@ -0,0 +1,364 @@
+package io.bazel.kotlin.integration
+
```
rockwotj, line 2: There is a lot of stuff here - at what point does it make sense to pull in something like [bazel-integration-testing](https://github.com/bazelbuild/bazel-integration-testing/blob/master/java/README.md) and then just have some lightweight kotlin extensions on those methods so you can write in the style of MultiplexTest...

restingbull: Soon, I think. This was a first pass.

src/test/kotlin/io/bazel/kotlin/integration/MultiplexTest.kt
restingbull: Testing says I need to make it bigger... :/

Bencodes:
@restingbull I ran some tests in our codebase with the worker java implementation and wasn't able to reproduce the issue. I was however able to reproduce it almost instantly switching back to the coroutines implementation.

rockwotj:
There are a bunch of thread local variables in the compiler using the IR backend: https://github.com/JetBrains/kotlin/blob/54cca882797de3d7861fcddae0214ad802937669/compiler/ir/ir.tree/src/org/jetbrains/kotlin/ir/util/SymbolTable.kt#L360-L374

I'm not sure if something like that could be the source of this error, but it seems like it if switching to a stable thread caller vs coroutine that may switch threads - thread local seems like a likely cause of these issues. Maybe if you use a coroutine dispatch that always switches threads at yield points may be a good way to force and see if that is the issue? (although I'm not sure how to do that off the top of my head)

rockwotj:
Looks like you could do that with a dispatcher that always returns true for `isDispatchNeeded` and then have one that for dispatch creates a new thread for every task...

Docs:
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/is-dispatch-needed.html
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-dispatcher/dispatch.html


Alternatively I wonder if you ran on a dispatcher with only a single thread if you wouldn't see corruption...

PR #653, opened 2022-01-07
Add experimental stdlib opt-in option

lorenz:
Since work on the generic opt-in framework seems to be stalled, this adds an explicit option for opting into Kotlin's experimental stdlib.

### DIFF
--- a/src/main/starlark/rkt_1_5/kotlin/opts.bzl
+++ b/src/main/starlark/rkt_1_5/kotlin/opts.bzl
@@ -121,6 +121,16 @@ _KOPTS = {
             True: ["-Xmulti-platform"],
         },
     ),
+    "x_opt_in_experimental_stdlib_api": struct(
+        args = dict(
+            default = False,
+            doc = "Enable experimental standard library features.",
+        ),
+        type = attr.bool,
+        value_to_flag = {
+            True: ["-Xopt-in=kotlin.ExperimentalStdlibApi"],
+        },
+    ),
 }
 
 KotlincOptions = provider(
--- a/src/main/starlark/rkt_1_6/kotlin/opts.bzl
+++ b/src/main/starlark/rkt_1_6/kotlin/opts.bzl
@@ -121,6 +121,16 @@ _KOPTS = {
             True: ["-Xmulti-platform"],
         },
     ),
+    "x_opt_in_experimental_stdlib_api": struct(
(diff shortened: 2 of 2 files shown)


rockwotj (approved):

Bencodes:
@lorenz have you verified that all of the correct `opts.bzl` files have this option specified?

```console
➜ fd opts.bzl
kotlin/internal/opts.bzl
src/main/starlark/legacy/jvm/opts.bzl
src/main/starlark/legacy/kotlin/opts.bzl
src/main/starlark/rkt_1_4/jvm/opts.bzl
src/main/starlark/rkt_1_4/kotlin/opts.bzl
src/main/starlark/rkt_1_5/jvm/opts.bzl
src/main/starlark/rkt_1_5/kotlin/opts.bzl
src/main/starlark/rkt_1_6/jvm/opts.bzl
src/main/starlark/rkt_1_6/kotlin/opts.bzl
```

Bencodes (approved):

lorenz:
@Bencodes I have only added the option to Kotlin versions where I've checked that it works (so 1.5 and 1.6). I'm not sure how far back that argument goes as documentation on experimental features is sparse.

restingbull:
Let's change this to a string list of classnames -- Jetbrains has left the api a bit looser than I'd like, but it's what we have.

Example:
```
kotlins_options(
  name=....
  opt_in = [
    "kotlin.ExperimentalStdlibApi"
  ]
)
```

Implementation will probably look something like this:
```
def _map_optin_class_to_flag(values):
    return ["-Xopt-in=%s" %v for v in values]

...

"x_optin": struct(
        args = dict(
            default = False,
            doc = "",
        ),
        type = attr.string_list,
        value_to_flag =None,
        map_value_to_flag = _map_optin_class_to_flag
    ),

...

def kotlinc_options_to_flags(kotlinc_options):
    """Translate KotlincOptions to worker flags

    Args:
        kotlinc_options maybe containing KotlincOptions
    Returns:
        list of flags to add to the command line.
    """
    if not kotlinc_options:
        return ""

    flags = []
    for n, o in _KOPTS.items():
        value = getattr(kotlinc_options, n, None)
        flag = o.value_to_flag.get(value, None) if o.value_to_flag else o.map_value_to_flag(value)
        if flag:
            flags.extend(flag)
    return flags
```
```

PR #664, opened 2022-01-21
Support Bazel to 5.0.0

Bencodes:


restingbull (approved):
Love a version bump with no changes!

restingbull:
QQ -- Do you what we need to change in the RBE? It's been complaining for sometime that we aren't pulling the right images for our kotlin version.

Issue #665, opened 2022-01-21
Plugins are being incorrectly propagated


mauriciogg:
Plugin deps are being propagated as regular deps [here](https://github.com/bazelbuild/rules_kotlin/blob/686f0f1cf3e1cc8c750688bb082316b3eadb3cb6/kotlin/internal/jvm/compile.bzl#L811). This is causing them to appear in the classpath of the app and crashing the build with the following error:
 
`ERROR: /home/nk/Snapchat/Dev/android/snapchat/platform/opera/sample/BUILD.bazel:188:15: in deps attribute of android_binary rule //platform/opera/sample:opera-sample-apk: Dependencies on .jar artifacts are not allowed in Android binaries, please use a java_import to depend on external/host_mvn/v1/gcs/snapengine-maven-publish/releases/com/google/dagger/dagger-compiler/2.37/processed_dagger-compiler-2.37.jar. If this is an implicit dependency then the rule that introduces it will need to be fixed to account for it correctly.`

If anything only exported plugins should be propagated (https://sourcegraph.com/github.com/bazelbuild/bazel/-/blob/src/main/java/com/google/devtools/build/lib/starlarkbuildapi/java/JavaInfoApi.java?L299&subtree=true) although this is not available in older versions of bazel.

restingbull:
Commented on the PR.

PR #666, opened 2022-01-21
Don't propagate plugins through deps

mauriciogg:
Plugins are not a target dependency. They are a host dep.
If propagated they will cause the build to break at the android_binary
level.
---
 fixes https://github.com/bazelbuild/rules_kotlin/issues/665


restingbull (commented):
A. Nice PR number. \m/
B. good catch
C. Add the plugins to `exported_plugins`

restingbull (approved):

PR #668, opened 2022-01-21
Make the underlying _kt library public.

### LINKED ISSUE #540: Associates attribute doesn't work with kt_android_local_test
I'm working on a test which works with internal class fields. Kotlin supports adding a "friends" module which opens up its internal members to a declaring entities. I found out the following description about special "associates" attribute here https://github.com/bazelbuild/rules_kotlin/pull/465, which suits my needs.

> Android should work, but because kt_android_* is a macro not a rule, the implicit target //my/android/library:mytarget_kt should be friended, since it has a KtJvmInfo. The //my/android/library will macro-resolve into an android_library. Until the android rules get the right kind of love, such that we can make a rule that has KtJvmInfo AND android-whatever providers, this simply is a known limitation we'll have to live with.

Unfortunately, the statement above is not completely realized in a project. 
An error is:
```
in kt_jvm_library rule //testproject/src/test/java/com/test:LocalTest_kt: target '//testproject/src/main/java/com/test:test_kt' is not visible from target '//testproject/src/test/java/com/test:LocalTest_kt'. Check the visibility declaration of the former target if you think the dependency is legitimate
```
This does not work because kt_jvm_library meant to be exposed the way described earlier is actually private by definition and seemingly should remain so (https://github.com/bazelbuild/rules_kotlin/blob/5357ac1a1d25e6ce083c0cb7ee489012d396177d/kotlin/internal/jvm/android.bzl#L57). 

Minimal example omitting standard stuff like rule definitions and basic trivial stuff:
**Local.kt:**
```
package com.test

class Local {
  internal fun doSomething(): Int = 0
}
```
**testproject/src/main/java/com/test/BUILD.bazel:**
```
kt_android_library(
    name = "test",
    srcs = [
        "Local.kt",
    ],
)
```
**LocalTest.kt:**
```
package com.test

@RunWith(JUnit4::class)
class LocalTest {
  @Test
  fun doSomething(): {
    LocalTest().doSomething()
  }
}
```
**testproject/src/test/java/com/test/BUILD.bazel:**
```
kt_android_local_test(
    name = "LocalTest",
    srcs = [
        "LocalTest.kt",
    ],
    associates = [
        "//testproject/src/main/java/com/test:test_kt",
    ]
    deps = [
        "//testproject/src/main/java/com/test:test",
        "@junit_junit",
    ],
)
```

Is there any way to make it work on an android tests without further exposing a library? Or should it be exposed?

@cgruber

cgruber:
Android kotlin libraries are a macro wrapper, and the underlying kotlin library (for target //foo:bar -> //foo:bar_kt) is marked private. Since the associates= (friendly) internal feature of kotlin requires access to a kotlin library, one is forced to often link test targets to the underlying android _kt backing target, which breaks if you're not in hte same package.

This PR makes it public (which sucks, but kind of is necessary until we can eliminate teh macro-based 'android-sandwich' thing we presently do, so that the associates= feature can work at all in kt_android_* targest.

Also adds an example (which fails without the change) to the android example.

Fixes #540

### DIFF
--- a/examples/android/libKtAndroid/src/main/java/examples/android/lib/TestKtValue.kt
+++ b/examples/android/libKtAndroid/src/main/java/examples/android/lib/TestKtValue.kt
@@ -5,11 +5,17 @@ import com.google.auto.value.AutoValue
 @AutoValue
 abstract class TestKtValue {
   abstract fun name(): String
-  fun builder(): Builder = AutoValue_TestKtValue.Builder()
+  internal fun nameInternal() = "${name()}_internal"
 
   @AutoValue.Builder
   abstract class Builder {
     abstract fun setName(name: String): Builder
     abstract fun build(): TestKtValue
   }
+  companion object {
+    fun create(builderFunction: Builder.() -> Unit) = AutoValue_TestKtValue.Builder().also {
+        builderFunction.invoke(it)
+      }.build()
+  }
+
 }
--- a/examples/android/WORKSPACE
+++ b/examples/android/WORKSPACE
@@ -50,6 +50,7 @@ maven_install(
         "com.google.dagger:dagger-producers:2.35.1",
         "com.google.auto.value:auto-value:1.6.5",
         "com.google.auto.value:auto-value-annotations:1.6.5",
+        "org.robolectric:robolectric:4.7.3",
     ],
     repositories = [
         "https://maven.google.com",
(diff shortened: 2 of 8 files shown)


examples/android/libKtAndroid/src/main/java/examples/android/lib/TestKtValue.kt
```diff
@@ -5,11 +5,17 @@ import com.google.auto.value.AutoValue
...
   abstract fun name(): String
-  fun builder(): Builder = AutoValue_TestKtValue.Builder()
+  internal fun nameInternal() = "${name()}_internal"
 
   @AutoValue.Builder
   abstract class Builder {
     abstract fun setName(name: String): Builder
     abstract fun build(): TestKtValue
   }
+  companion object {
```
cgruber, line 15: Not strictly necessary, but builder() wasn't used, and having a very incorrect use of autovalue bugged me.

Bencodes (approved):

restingbull (commented):
LGTM.

PR #671, opened 2022-01-28
Add github action for creating releases

Bencodes:


rockwotj:
I think @cgruber tried this but GitHub actions aren't enabled for the bazel org... https://github.com/bazelbuild/rules_kotlin/commit/6423eb339b2029049c63aa81013889b9b2b97cfa

restingbull:
Let's see if this works now...

PR #675, opened 2022-02-03
Use canonical repo name for android rules: build_bazel_rules_android

dflems:
The [canonical](https://github.com/bazelbuild/rules_android/blob/e8fbc49f913101e846235b9c9a31b3aa9788364a/WORKSPACE#L1) repo name for the [android rules repo](https://github.com/bazelbuild/rules_android) appears to be `build_bazel_rules_android`, not `rules_android`. Our project already had the former, so integrating this ruleset led to us having two copies of the android ruleset in our project which felt a little redundant and unnecessary.

Perhaps there's a perfectly good or historical reason for using `rules_android` in here that I'm not aware of. If this complicates things in any way, feel free to close this PR.

### DIFF
--- a/examples/associates/WORKSPACE
+++ b/examples/associates/WORKSPACE
@@ -20,13 +20,6 @@ kt_register_toolchains()
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
-    name = "rules_android",
-    sha256 = "cd06d15dd8bb59926e4d65f9003bfc20f9da4b2519985c27e190cddc8b7a7806",
-    strip_prefix = "rules_android-0.1.1",
-    urls = ["https://github.com/bazelbuild/rules_android/archive/v0.1.1.zip"],
-)
-
-http_archive(
     name = "build_bazel_rules_android",
     sha256 = "cd06d15dd8bb59926e4d65f9003bfc20f9da4b2519985c27e190cddc8b7a7806",
     strip_prefix = "rules_android-0.1.1",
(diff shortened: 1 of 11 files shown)


examples/associates/WORKSPACE
```diff
@@ -19,13 +19,6 @@ kt_register_toolchains()
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-http_archive(
-    name = "rules_android",
-    sha256 = "cd06d15dd8bb59926e4d65f9003bfc20f9da4b2519985c27e190cddc8b7a7806",
-    strip_prefix = "rules_android-0.1.1",
-    urls = ["https://github.com/bazelbuild/rules_android/archive/v0.1.1.zip"],
-)
-
```
dflems, line 28: this example workspace had both :)

src/main/starlark/core/repositories/download.bzl
```diff
@@ -89,7 +89,7 @@ def kt_download_local_dev_dependencies():
     )
 
     rules_stardoc_repository(
-        name = "rules_android",
+        name = "build_bazel_rules_android",
```
dflems, line 92: not sure of the context of this macro but it seems to basically be `http_archive` but adding some stardoc extras

dflems:
@cgruber @restingbull ?

Bencodes (approved):
I believe we use `rules_android` in our repo, but not against renaming that to align with rules_kotlin and rules_android.

This change lgtm! We should make sure this ends up in the changelog for the next release so that people using `rules_android` know to adapt their projects.

Bencodes (approved):

restingbull (approved):

PR #677, opened 2022-02-04
Update api docs.

cgruber:


restingbull (approved):

cgruber:
Other updates already happened, so this is obsolete.

PR #678, opened 2022-02-04
Fix annotation processing when no kotlin sources are passed

mauriciogg:
The rules assume the kotlin compiler will take care of annotation
processing, but kotlinc does not run if not kotlin sources are
specified. This commit forces annotation processing with java
if no kotlin sources are passed.


cgruber (changes requested):
This looks good, but can you please add some cases to one of the example projects, to force this case, please?

mauriciogg:
> This looks good, but can you please add some cases to one of the example projects, to force this case, please?

Done.

cgruber:
This file has been deleted, and the whole mechanism has changed, so this is no longer a valid change (the purpose of it may not have changed, but the mechanism is no longer here in the JavaCompiler class)

PR #679, opened 2022-02-05
Support KSP annotation processing

Bencodes:
First pass at supporting KSP annotation processing as an alternative to Kapt.

This PR runs KSP as part of the existing `KotlinCompile` action so that Kotlinc only has to be invoked once while compiling a target. We could optionally pull this out into a unique action, like we do for Kapt, at the cost of re-processing Kotlin sources.


RinOrz:
I've been planning on moving my Kotlin Gradle project to Bazel lately, but Bazel's support for Kotlin ecology seems imperfect, so I'm just curious if KSP can't be implemented like Jetpack Compose? Just apply the Kotlin compiler plugin, because essentially, KSP is just a KCP plugin

Bencodes:
> I've been planning on moving my Kotlin Gradle project to Bazel lately, but Bazel's support for Kotlin ecology seems imperfect, so I'm just curious if KSP can't be implemented like Jetpack Compose? Just apply the Kotlin compiler plugin, because essentially, KSP is just a KCP plugin

@RinOrz it should theoretically be possible. I took this approach because I was interested in at a lower level so that we could play around with performance a bit more (Ex: Running KSP in a standalone action like we do for kapt vs running as part of the existing Kotlin compile).

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/compilation_task.kt
```diff
@@ -170,6 +170,47 @@ internal fun JvmCompilationTask.kaptArgs(
...
+        "kotlinOutputDir" to listOf(directories.generatedSources),
+        // Directory where META-INF data is written to. This might not be the most ideal place to
+        // write this. Maybe just directly to the classes directory?
+        "resourceOutputDir" to listOf(directories.generatedSources),
+        // TODO(bencodes) Not sure what this directory is yet.
+        "kspOutputDir" to listOf(directories.incrementalData),
+        // Directory to write KSP caches. Shouldn't matter because incremental is disabled
+        "cachesDir" to listOf(directories.incrementalData),
+        // Set withCompilation to false because we run this as part of the standard kotlinc pass
+        "withCompilation" to listOf("false"),
```
idanakav, line 202: With `withCompilation=false` wouldn't this require an extra compilation pass to compile the generated ksp code?

Bencodes: The outputs get swept up by the existing Kotlinc pass. There are some tests in place that assert that the compiled class files exist. https://github.com/bazelbuild/rules_kotlin/pull/679/files#diff-bdab9cf068d9747c4ade4192a1d414ebbe1e33d1634b6747fa6e3b942899e1e4R108

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/compilation_task.kt
```diff
@@ -170,6 +170,47 @@ internal fun JvmCompilationTask.kaptArgs(
...
+        "projectBaseDir" to listOf(directories.incrementalData),
+        // Disable incremental mode
+        "incremental" to listOf("false"),
+        // Directory where class files are written to. Files written to this directory are class
+        // files being written directly from the annotation processor, not Kotlinc
+        "classOutputDir" to listOf(directories.generatedClasses),
+        // Directory where generated Java sources files are written to
+        "javaOutputDir" to listOf(directories.generatedJavaSources),
+        // Directory where generated Kotlin sources files are written to
+        "kotlinOutputDir" to listOf(directories.generatedSources),
```
restingbull, line 193: `kt_compiler_plugin` provide the following template variables:
    - `{generatedSources}` template provides a generated source folder
    - `{generatedClasses}` for class generation via the `kt_compiler_plugin`
    - `{temp}` for incremental work

`kt_compiler_plugin.stubs_phase` runs before compilation and kapt. We might consider renaming this to `header`/`abi` phase depending on how we structure the actions.

 Do we need another template variable for java sources to be generated? If so, should it be propagated to the java compile?

ref: https://github.com/bazelbuild/rules_kotlin/blob/master/docs/kotlin.md#kt_compiler_plugin

Kernald:
What's the status of this? Is there any way I could help?

Bencodes:
> What's the status of this? Is there any way I could help?

Definitely needs some more testing. We are moving over our annotation processors over the KSP internally and will hopefully have something to report back with in the coming weeks.

My hope here is that we don't actually need to build this into `rules_kotlin` and that we can just lean into the existing Kotlinc plugin infrastructure that exists.

lukaciko:
@Bencodes have you considered whether KSP can be used in a target that also runs KAPT, and if so, if KSP & KAPT can depend on each other's outputs? Looking at https://github.com/google/ksp/issues/863 that decision hasn't been made yet in Gradle-land.

Bencodes:
@lukaciko that's not a case that we've accounted for and not sure if we do. This seems like something that would add a lot of complexity to our annotation processing, and even introduce some clashing between the two.

lukaciko:
@Bencodes I think it's fair enough if it is not supported, as long as that's clear and explicit. I'd like to avoid one of those cases where something is accidentally supported in one version of the rules and gets dropped in another.

Bencodes:
I took another stab at getting a less intrusive version of KSP working yesterday. This approach does work and it comes with the added benefit of us not needing to couple KSP to `rules_kotlin`.

https://github.com/bazelbuild/rules_kotlin/pull/829

PR #680, opened 2022-02-07
Fix --enable_runfiles on Windows

jesseschalken:
This code effectively asserted that `RUNFILES_MANIFEST_FILE` is set on Windows, but this is not true if `--enable_runfiles` is enabled.

So allow `RUNFILES_MANIFEST_FILE` not to be set on Windows.


jesseschalken:
As a side note it isn't clear what the point of any of this code is versus using Bazel's own [Java runfiles library](https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java) directly.

cgruber (approved):

jesseschalken:
@jin @timpeut @restingbull Ping on this PR. It's 2 lines and I'd like to get rid of our internal fork.

restingbull (approved):

PR #684, opened 2022-02-14
Remove Android NDK

Kernald:
It's not used by the rules, but this declaration makes it mandatory to
have an NDK installed to build the rules locally. As far as I know, it
shouldn't be a breaking change as users of the release archive never got
this NDK declaration in the first place.


Kernald:
@Bencodes I can't figure out how to run it with the same arguments than on CI. As far as I can tell, it started failing on `master` because the buildifier version is not pinned - probably something we should address?

Bencodes:
@Kernald just pinned the version in master. Want to give this a rebase?

restingbull (approved):

PR #689, opened 2022-03-01
Support using --jvmopt with kt_jvm_binary

mai93:
Add the JVM options specified via [`--jvmopt`](https://docs.bazel.build/versions/main/user-manual.html#flag--jvmopt) to the `jvm_flags` of `kt_jvm_binary` targets.

Since `--jvmopt` options should apply to the JVM for all java binaries in the Bazel command, they should also affect the Kotlin JVM binaries. This is needed by the IntelliJ Bazel plugin to pass a javaagent from command line to the Kotlin binary targets to enable debugging coroutines.

### DIFF
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -243,11 +243,13 @@ def kt_jvm_library_impl(ctx):
 
 def kt_jvm_binary_impl(ctx):
     providers = _kt_jvm_produce_jar_actions(ctx, "kt_jvm_binary")
+    jvm_flags = ctx.attr.jvm_flags[:]
+    jvm_flags.extend(ctx.fragments.java.default_jvm_opts)
     _write_launcher_action(
         ctx,
         providers.java.transitive_runtime_jars,
         ctx.attr.main_class,
-        ctx.attr.jvm_flags,
+        jvm_flags,
     )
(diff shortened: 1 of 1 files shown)


Bencodes (approved):

kotlin/internal/jvm/impl.bzl
```diff
@@ -243,11 +243,13 @@ def kt_jvm_library_impl(ctx):
 
 def kt_jvm_binary_impl(ctx):
     providers = _kt_jvm_produce_jar_actions(ctx, "kt_jvm_binary")
+    jvm_flags = ctx.attr.jvm_flags[:]
+    jvm_flags.extend(ctx.fragments.java.default_jvm_opts)
```
Bencodes, line 247: Will ordering be an issue here, meaning jvm_flags that are specific to the `kt_jvm_binary` target can be overridden by the default jvm opts provided by the java fragment?

Bencodes, line 247: Will ordering be an issue here, meaning jvm_flags that are specific to the `kt_jvm_binary` target can be overridden by the default jvm opts provided by the java fragment?

Bencodes: @mai93 looking at this more I think we should reverse the ordering here.

mai93: Good point, I reversed the order

Bencodes (approved):

PR #694, opened 2022-03-05
Remove legacy Javac compilation

Bencodes:
Notable changes:
- Removing the legacy compilation so that the new java_common path is the default means for multi-source compilation.
- Reworked the Jdeps tests to read the test java classes from a java_library output instead of compiling the Java source files by directly calling javac.


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
```diff
@@ -90,13 +88,6 @@ class KotlinJvmTaskExecutor @Inject internal constructor(
                 emptyList()
               }
             }
-            context.execute("javac") {
```
nkoroste, line 93: It looks to me that we were compiling java twice, once here and once in https://sourcegraph.com/github.com/bazelbuild/rules_kotlin/-/blob/kotlin/internal/jvm/compile.bzl?L699  

I'm still testing this against our code base but so far it seems like it's safe to remove this -- didn't notice huge perf gains but we dont have that much java in our codebase

nkoroste (approved):
This works with our repo. I think it's safe to submit this and close https://github.com/bazelbuild/rules_kotlin/pull/678 as it will no longer be relevant.

restingbull (approved):
We'll need to cut  1.6 alpha candidate to get more testing.

Bencodes:
@restingbull sounds good. I'm going to merge this one in and we can look into the java_home/rbe issues async https://github.com/bazelbuild/rules_kotlin/pull/694/commits/d0abebe7ee671e97d99f8938f27d9a8a972d358d

PR #703, opened 2022-03-18
Use JavaPersistantWorker, create new kotlinc every invocation

restingbull:
* Remove coroutine worker in favor of stdlib java for speed (and hopefully reliability)
* Create a new `K2JVMCompiler` on each invocation to avoid keeping thread locals
* Set `zip.handler.uses.crc.instead.of.timestamp` as a hail mary to prevent corruption
* Remove `android_ndk` from examples, as no example uses it... and it's broken for later versions of the android ndk.
* Standardize on Bazel 5.


examples/anvil/WORKSPACE
```diff
@@ -28,12 +28,9 @@ http_archive(
 
 load(
     "@build_bazel_rules_android//android:rules.bzl",
-    "android_ndk_repository",
```
cgruber, line 31: yass qween

src/main/kotlin/io/bazel/kotlin/compiler/BazelK2JVMCompiler.kt
```diff
@@ -22,8 +22,10 @@ import org.jetbrains.kotlin.cli.jvm.K2JVMCompiler
 import org.jetbrains.kotlin.config.Services
 
 @Suppress("unused")
-class BazelK2JVMCompiler(private val delegate: K2JVMCompiler = K2JVMCompiler()) {
-  fun exec(errStream: java.io.PrintStream, vararg args: kotlin.String): ExitCode {
+class BazelK2JVMCompiler {
+  fun exec(errStream: java.io.PrintStream, vararg args: String): ExitCode {
+    System.setProperty("zip.handler.uses.crc.instead.of.timestamp", "true")
```
cgruber, line 27: This is painful (that it's set as a global property in an exec() method I mean, nothing to be done about that if we want to ensure it, I suspect)

src/test/kotlin/io/bazel/worker/WorkerEnvironment.kt
```diff
@@ -43,12 +43,17 @@ object WorkerEnvironment {
...
           val responses = ConcurrentLinkedQueue<WorkResponse>()
-          executor.submit {
-            generateSequence { WorkResponse.parseDelimitedFrom(stdOut.input) }.forEach {
-              responses.add(it)
+          val reader = executor.submit {
+            Thread.currentThread().name = "readStdOut"
+            while (true) {
+              WorkResponse.parseDelimitedFrom(stdOut.input)
+                ?.let(responses::add)
+                ?: break
```
cgruber, line 51: Ok, I like this sneaky little pattern.

src/test/kotlin/io/bazel/worker/JavaPersistentWorkerTest.kt
```diff
@@ -73,8 +67,15 @@ class PersistentWorkerTest {
...
       closeStdIn()
-      return@inProcess generateSequence { readStdOut() }
-    }.associateBy { wr -> wr.requestId }
+      waitForStdOut()
+      return@inProcess generateSequence {
+        readStdOut().apply {
+          println("sequence $this")
+        }
+      }
+    }.associateBy { wr ->
```
cgruber, line 76: wr? I know you didn't name it here, but can you drive-by rename with something more clear?

src/test/kotlin/io/bazel/worker/WorkerEnvironment.kt
```diff
@@ -87,6 +92,13 @@ object WorkerEnvironment {
       }
     }
 
+    fun waitForStdOut() {
+      while (stdOut.input.available() != 0) {
+        Thread.onSpinWait()
+        Thread.sleep(1) // don't starve other processes.
```
cgruber, line 98: Doesn't doing a `Thread.sleep(1)` more or less do as much as `onSpinWait()` or more? As in, if you have `sleep(1)` do you still need `onSpinWait()`?

src/main/kotlin/io/bazel/worker/JavaPersistentWorker.kt
```diff
@@ -0,0 +1,122 @@
...
+ * PersistentWorker satisfies Bazel persistent worker protocol for executing work.
+ *
+ * Supports multiplex (https://docs.bazel.build/versions/master/multiplex-worker.html) provided
+ * the work is thread/coroutine safe.
+ *
+ * @param executor thread pool for executing tasks.
+ * @param captureIO to avoid writing stdout and stderr while executing.
+ * @param cpuTimeBasedGcScheduler to trigger gc cleanup.
+ */
+class JavaPersistentWorker(
```
cgruber, line 41: Why the rename here? Is this naming relevant to "java vs. js or llvm/native"? Or java in some other sense?

cgruber (approved):
Nice clean up. Just some non-blocking nits and questions.

src/main/kotlin/io/bazel/worker/JavaPersistentWorker.kt
restingbull: Ah, I had named the other Kotlin for coroutines. Will fix.

src/test/kotlin/io/bazel/worker/WorkerEnvironment.kt
restingbull: It's very jetbrains kotlin. Been reading too much code from there.

src/main/kotlin/io/bazel/kotlin/compiler/BazelK2JVMCompiler.kt
restingbull: It also doesn't work. Will remove.

examples/anvil/WORKSPACE
restingbull: The ndk rules are a mess. Don't work past 27 afaict.

PR #709, opened 2022-04-04
Remove @Singleton annotations.

brettwooldridge:
This is a change that removes all singleton annotations. I thought that this could actually effect performance, but it appears there is no change (see below). But what removing singletons does is reduce the likelihood of "state bleed-over" during compilations.

I ran a series of timing tests with and without singleton annotations. ``bazel clean --expunge`` was executed between each run. Command used was ``time make``.

With Singleton annotations:
```
make  0.05s user 0.05s system 0% cpu 2:54.62 total
make  0.05s user 0.05s system 0% cpu 2:39.98 total
make  0.05s user 0.05s system 0% cpu 2:42.00 total
```
*Without* Singleton annotations:
```
make  0.05s user 0.05s system 0% cpu 2:46.34 total
make  0.05s user 0.05s system 0% cpu 2:41.53 total
make  0.05s user 0.05s system 0% cpu 2:41.74 total
```
As can be seen, there is little to no difference between the two, except the use without Singletons is likely safer.


src/main/kotlin/io/bazel/kotlin/builder/JdepsMergerComponent.java
```diff
@@ -21,10 +21,8 @@
 import io.bazel.kotlin.builder.tasks.MergeJdeps;
 import io.bazel.kotlin.builder.tasks.jvm.JdepsMerger;
 
-import javax.inject.Singleton;
 import java.io.PrintStream;
 
-@Singleton
```
cgruber, line 27: You shouldn't remove it from the component, here it doesn't tell dagger to make it a singleton (component lifecycle is managed in user code, not by dagger) but rather is a scope constraint, saying that the only scopes that should be memoized in this component are labelled `@Singleton`. Remove them from objects in the graph if appropriate, but unless there are no objects which share state within the life of this component, provably, this should remain.

src/main/kotlin/io/bazel/kotlin/builder/KotlinBuilderComponent.java
```diff
@@ -25,9 +25,6 @@
 import io.bazel.kotlin.builder.tasks.jvm.KotlinJvmTaskExecutor;
 import io.bazel.kotlin.builder.toolchain.KotlinToolchain;
 
-import javax.inject.Singleton;
-
-@Singleton
```
cgruber, line 30: Ditto.

src/main/kotlin/io/bazel/kotlin/builder/JdepsMergerComponent.java
cgruber: Especially since toolchain is effectively a singleton scope (but here it's supplied during component construction via a @BindsInstance builder method). So technically this is valid, but I think it's poor documentation. It might be reasonable to use a custom annotation (e.g. `@PerWorkload` or something, even if there aren't any declared scoped (memoized) objects inside the graph, unless you're forbidding shared state, which we aren't)

cgruber: What state bleed-over are you most concerned about, incidentally, or is this more cautionary?

brettwooldridge: Specifically, multiple threads running in a single instance of the K2JVMCompiler class. At least in Buck we found this this lead to unpredictable corruption of outputs. Corruption has also been reported against the Bazel Kotlin rules.

brettwooldridge:
@cgruber As long as I have your ear, is there any documentation regarding how the runtime structure of classes is assembled?

In particular, I am trying to understand how persistent workers actually work. In Buck, we are able to run the K2JVMCompiler "in-process" with Buck, i.e. executing within the same JVM instance that is running Buck itself. However, in Bazel, when I run ``ps`` during compilation, it appears that each "worker" is forked as a separate JVM process:
```
brettw            3788   0.0  2.2 26245504 1452628   ??  S     5:08AM   8:38.28 /private/var/tmp/_bazel_brettw/ba6f1f31eab91f1d52ff2982674ba73f/execroot/__main__/bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/build.runfiles/remotejdk11_macos/bin/java -classpath bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/build.runfiles/__main__/../io_bazel_rules_kotlin/src/main/kotlin/kotlin_worker.jar:bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/build.runfiles/__main__/../remote_java_tools/java_tools/JacocoCoverage_jarjar_deploy.jar -XX:ErrorFile=/dev/stderr io.bazel.kotlin.builder.cmd.Build --persistent_worker
```
If this is so, how is "work" passed to each worker process? We are suspecting that the JIT-compilation of the K2JVMCompiler that must occur within each independent JVM instance is substantially slowing down the build in Bazel compared to our Buck build. Under Buck we were seeing 1-minute total clean build times, compared to 3-minute total clean build times under Bazel.

src/main/kotlin/io/bazel/kotlin/builder/JdepsMergerComponent.java
rockwotj: FWIW after #703 we create a new kotlin compiler for every build. The code here has a couple layers of indirection and this singleton creates many kotlinc instances 


https://github.com/bazelbuild/rules_kotlin/commit/cb3dd0b6fe86b3251a001aef2a711d69addd8704

rockwotj:
@brettwooldridge see these docs: https://docs.bazel.build/versions/main/persistent-workers.html

Essential Protos are passed over stdin/stdout

src/main/kotlin/io/bazel/kotlin/builder/JdepsMergerComponent.java
brettwooldridge: @rockwotj To be honest, I'm having trouble following the indirection. I see that ``BazelK2JVMCompiler`` instantiates a ``K2JVMCompile`` every time that it itself is instantiated, but what instantiates the ``KotlincInvoker`` (that is marked as a Singleton) that instantiates the ``BazelK2JVMCompiler``? If ``KotlinInvoker`` is truly a Singleton, then ultimately there is still a Singleton ``K2JVMCompiler``. Unless, as I said, I am not following the indirection.

restingbull:
> @cgruber As long as I have your ear, is there any documentation regarding how the runtime structure of classes is assembled?
> 
> In particular, I am trying to understand how persistent workers actually work. In Buck, we are able to run the K2JVMCompiler "in-process" with Buck, i.e. executing within the same JVM instance that is running Buck itself. However, in Bazel, when I run `ps` during compilation, it appears that each "worker" is forked as a separate JVM process:
> 
> ```
> brettw           21301 142.4  1.3 26162484 890544   ??  S     5:04PM   0:48.33 /private/var/tmp/_bazel_brettw/2eb6368d09513d6b04f237307bef6bb8/execroot/dev_io_bazel_rules_kotlin/bazel-out/host/bin/src/main/kotlin/io/bazel/kotlin/builder/cmd/build.runfiles/remotejdk11_macos/bin/java -classpath bazel-out/host/bin/src/main/kotlin/io/bazel/kotlin/builder/cmd/build.runfiles/dev_io_bazel_rules_kotlin/src/main/kotlin/io/bazel/kotlin/builder/cmd/build_jarjar.jar -XX:ErrorFile=/dev/stderr -XX:+IgnoreUnrecognizedVMOptions --add-opens=java.base/java.nio=ALL-UNNAMED --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=jdk.jdeps/com.sun.tools.jdeps=ALL-UNNAMED io.bazel.kotlin.builder.cmd.Build --persistent_worker
> ```
> 
> If this is so, how is "work" passed to each worker process? We are suspecting that the JIT-compilation of the K2JVMCompiler that must occur within each independent JVM instance is substantially slowing down the build in Bazel compared to our Buck build. Under Buck we were seeing 1-minute total clean build times, compared to 3-minute total clean build times under Bazel.

I'm currently working on removing the indirection. The initial setup was arranged to allow swapping the compiler implementation via classpaths. This is no longer necessary.

src/main/kotlin/io/bazel/kotlin/builder/JdepsMergerComponent.java
rockwotj: If you look [here](https://github.com/bazelbuild/rules_kotlin/blob/772d973c7750b350062cd449c3cb582f83433591/src/main/kotlin/io/bazel/kotlin/compiler/BazelK2JVMCompiler.kt#L28) a new `K2JVMCompiler` is created every time `BazelK2JVMCompiler::exec` is invoked, so we get a new compiler for each build as `exec` is called for each target.

PR #712, opened 2022-04-07
Allow java 15 and 17 for jvm_target

### LINKED ISSUE #711: Can not use with Java versions newer than 13
I am experimenting with upgrading a mixed-jvm codebase to java 17.

I started by setting `--java_language_version=17` and `--java_runtime_version=17` in our `.bazelrc`.

Our Kotlin targets that depend on java targets started failing to coming with

```
bad_class_file: path/to/some/File.class
  class file has wrong version 61.0, should be 55.0
  Please remove or make sure it appears in the correct subdirectory of the classpath.
```

We are using a custom Kotlin toolchain, so I tried to set `jvm_target = "17"`. However, the rule impl prevents this:

```
BUILD:89:20: //:kotlin_toolchain_impl: invalid value in 'jvm_target' attribute: has to be one of '1.6', '1.8', '9', '10', '11', '12' or '13' instead of '17'
```

Indeed, kotlin_toolchain_impl has an allowlist for options that only goes up to 13: https://github.com/bazelbuild/rules_kotlin/blob/772d973c7750b350062cd449c3cb582f83433591/kotlin/internal/toolchains.bzl#L176-L188

andrewhamon:
This seems to fix #711

I've tested some Kotlin targets at my company that depend on java deps and I can successfully build with java 17, but haven't done anything more thorough than that.

Also, what do y'all think about removing the check altogether? Kotlin compiler will fail if an invalid version is provided anyways, and this Bazel-level check doesn't even spare users from that error if they try to use java 17 with a Kotlin compiler that is too old.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ load("@io_bazel_rules_kotlin//kotlin:core.bzl", "define_kt_toolchain")
 define_kt_toolchain(
     name = "kotlin_toolchain",
     api_version = KOTLIN_LANGUAGE_LEVEL,  # "1.1", "1.2", "1.3", "1.4", "1.5", or "1.6"
-    jvm_target = JAVA_LANGUAGE_LEVEL, # "1.6", "1.8", "9", "10", "11", "12", or "13",
+    jvm_target = JAVA_LANGUAGE_LEVEL, # "1.6", "1.8", "9", "10", "11", "12", "13", "15", or "17"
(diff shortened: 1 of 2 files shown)


nkoroste (approved):

Bencodes:
> Also, what do y'all think aout removing the check altogether?

This would be a better question for @restingbull

Bencodes:
@andrewhamon can you also update the `## Custom toolchain` section in the root README.md to list the newly supported jvm targets?

Bencodes (approved):

andrewhamon:
@Bencodes just circled back to update the readme only to find you did it for me... thanks!

PR #713, opened 2022-04-12
Rename PersistentWorkerTest to JavaPersistentWorkerTest

nkoroste:
File was not remade in https://github.com/bazelbuild/rules_kotlin/pull/703


Bencodes (approved):

src/test/kotlin/io/bazel/worker/JavaPersistentWorkerTest.kt
```diff
@@ -24,7 +24,7 @@ import org.junit.Test
 import java.nio.charset.StandardCharsets.UTF_8
 import java.util.concurrent.Executors
 
-class PersistentWorkerTest {
```
rockwotj, line 27: Shouldn't the file be renamed too?

Issue #618, opened 2021-11-12
Remove tag to disable ABI jar production (as of 1.6)


cgruber:
The bug in kotlinc which caused ABI jars to be produced incorrect jars, without accounting for cross-inlining. 1.6 fixes this, so we should be able to use ABI jars reliably in usages of Kotlin 1.6+

Bencodes:
This is the PR that fixed it https://github.com/JetBrains/kotlin/pull/3986

sfs:
I'm responsible for the `jvm-abi-gen` changes. While testing this internally we ran into some bugs in `kotlinx-metadata` (which `jvm-abi-gen` bundles). The fixes (https://github.com/JetBrains/kotlin/pull/4614, https://github.com/JetBrains/kotlin/pull/4617) for these issues didn't quite make it into the 1.6.0 release, but should be included in the next `kotlinx-metadata` and compiler releases. Until then you may want to use the previous implementation (via `-P plugin:org.jetbrains.kotlin.jvm.abi:useLegacyAbiGen=true`), although that one contains a number of other bugs...

Without the fixes above we currently loose the nullability of the underlying types of value classes and the flags that mark interfaces compiled with `-Xjvm-default=all` or `all-compatibility`.

Sorry for the inconvenience!

nkoroste:
@sfs sounds like the fixes are coming in `1.6.10`?

sfs:
Yes, all of the fixes that are relevant for the compiler made it into the 1.6.10 release and we have been using `jvm-abi-gen` internally with these patches applied for about a week now.

---

We encountered one more problem internally (https://github.com/JetBrains/kotlin/pull/4677), but this only affects ABI jars imported as dependencies into IntelliJ. In an ideal world that shouldn't be a common use case, though, since it breaks some other IntelliJ features (e.g., decompilation).

PR #716, opened 2022-04-16
Enable experimental_use_abi_jars by default

Bencodes:
Incremental step towards removing this flag entirely https://github.com/bazelbuild/rules_kotlin/issues/618


restingbull:
I guess the question I have is "Is it safe?"

PR #717, opened 2022-04-16
Support -Xuse-fir to enable the experimental new Kotlin frontend

Bencodes:
The Kotlin team is working on a new experimental frontend that is expected to improve compile times. This is very much an experimental feature at the moment, but we should support dog fooding this in rules_kotlin so we can report issues. https://github.com/JetBrains/kotlin/blob/1.6.20/compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/CommonCompilerArguments.kt#L303-L308


Bencodes:
`-Xuse-fir-extended-checkers` is another front-end related flag that we could support if needed, but i'm not familiar with when this would need to be enabled.

restingbull (approved):

PR #722, opened 2022-04-27
Update compose example to 1.2.0-alpha08

Bencodes:


restingbull (approved):
Seems like we need another dep bump?

Bencodes:
> Seems like we need another dep bump?

Waiting on Android 31, which explains why this passed locally for me and doesn't build on CI. https://github.com/bazelbuild/continuous-integration/issues/1363

restingbull:
Need to bump the sdk

PR #733, opened 2022-05-04
'experimental_use_abi_jars' toolchain attribute cleanup

### LINKED ISSUE #732: 'kt_abi_plugin_incompatible' should be respected even if 'experimental_use_abi_jars' is False
In kotlin/internal/jvm/compile.bzl :
-        if not "kt_abi_plugin_incompatible" in ctx.attr.tags and toolchains.kt.experimental_use_abi_jars == True:

We should probably do a larger cleanup, and remove all ref to experimental_use_abi_jars toolchain attribute

oliviernotteghem:
Fixes https://github.com/bazelbuild/rules_kotlin/issues/732

### DIFF
--- a/docs/kotlin.md
+++ b/docs/kotlin.md
@@ -416,7 +416,7 @@ kt_kotlinc_options(<a href="#kt_kotlinc_options-name">name</a>, <a href="#kt_kot
 ## define_kt_toolchain
 
 <pre>
-define_kt_toolchain(<a href="#define_kt_toolchain-name">name</a>, <a href="#define_kt_toolchain-language_version">language_version</a>, <a href="#define_kt_toolchain-api_version">api_version</a>, <a href="#define_kt_toolchain-jvm_target">jvm_target</a>, <a href="#define_kt_toolchain-experimental_use_abi_jars">experimental_use_abi_jars</a>,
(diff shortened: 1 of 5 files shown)


Bencodes (approved):
Lgtm as long as we still have the docs for `kt_abi_plugin_incompatible`. I think there's one edge case in our codebase that still might be broken, but it's isolated to a single module that has some creative Kotlin in it.

Bencodes:
@oliviernotteghem we can land this PR after 1.6 has it's first stable release, which would put this in the 1.7 release.

Bencodes:
Alright was chatting with @nkoroste and @cgruber about how to best merge this. The main concern that we still support Kotlin 1.6 which has version of `kotlin-abi-gen` that are completely broken, so enabling this globally is probably going to cause issues for the people that haven't upgraded yet.

The plan that we came up with is:
- Drop the top level Kotlin toolchain option and enable `experimental_use_abi_jars` by default so that more of our users are bucketed into using the abi compiler.
- Keep the `config_setting` so that you can still disable `kotlin-abi-gen` globally using `--define=experimental_use_abi_jars=1|0`.
- Keep the `kt_abi_plugin_incompatible` check so that there are still ways to opt-out at the target level if there are problematic Kotlin files that aren't yet abi-gen compatible.

PR #749, opened 2022-05-24
Run `kapt` on `kt` files passed via `srcjar`s

arunkumar9t2:
Fixes https://github.com/bazelbuild/rules_kotlin/issues/748﻿

The issue is `srcjars` were not passed to `kapt`. `java` files in a `srcjar` works since `.java` files are handled by `java_common`.

An unfortunate side effect is `srcjars` need to be passed to multiple places even if they are not going to be used. For example, if srcjars contain only `kt` files they need not be passed to `java_common`. 

A better approach would be to unpack the srcjars to examine the content and then decide if they passed to relevant compiler but that might be premature optimisation.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -554,9 +554,10 @@ def _run_kt_java_builder_actions(
     compile_jars = []
     output_jars = []
     kt_stubs_for_java = []
+    has_sources = srcs.kt or srcs.src_jars
 
     # Run KAPT
-    if srcs.kt and annotation_processors:
+    if has_sources and annotation_processors:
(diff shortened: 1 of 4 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -554,9 +554,10 @@ def _run_kt_java_builder_actions(
     compile_jars = []
     output_jars = []
     kt_stubs_for_java = []
+    has_sources = srcs.kt or srcs.src_jars
```
Bencodes, line 557: `has_kt_sources` might be a better name for this so that we don't mistakenly reference this for Java specific blocks of code.

arunkumar9t2: Addressed 👍🏼

Bencodes (approved):
Lgtm. Going to test this against our repo before we merge.

Issue #761, opened 2022-05-29
1.6.0-RC-1/2  - JDeps Merge Failure


rob-pomelo:
Tried 1.6.0-RC-2 (coming from 1.5), seeing the following jdeps failure on binary and test targets. 

`SEVERE: ERROR: unexpected exception
java.lang.IllegalArgumentException: --inputs is not optional
	at io.bazel.kotlin.builder.utils.ArgMap.mandatory(ArgMap.kt:63)
	at io.bazel.kotlin.builder.utils.ArgMap.mandatory(ArgMap.kt:76)
	at io.bazel.kotlin.builder.tasks.jvm.JdepsMerger.execute(JdepsMerger.kt:130)
	at io.bazel.kotlin.builder.tasks.MergeJdeps.invoke(MergeJdeps.kt:28)
	at io.bazel.worker.PersistentWorker$workTo$1.invoke(PersistentWorker.kt:97)
	at io.bazel.worker.PersistentWorker$workTo$1.invoke(PersistentWorker.kt:97)
	at io.bazel.worker.WorkerContext$TaskContext.resultOf(WorkerContext.kt:128)
	at io.bazel.worker.WorkerContext.doTask(WorkerContext.kt:156)
	at io.bazel.worker.PersistentWorker$start$1$1$producer$1$1.invoke$lambda-0(PersistentWorker.kt:70)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)`
	
[...]

PR #763, opened 2022-06-02
Symlink empty jdeps as output when there are none to merge

Bencodes:
Symlinking the empty jdeps file in place of the expected output jdeps when there are no jdeps to merge.

Fixes https://github.com/bazelbuild/rules_kotlin/issues/761

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -694,15 +694,19 @@ def _run_kt_java_builder_actions(
         if java_info.outputs.jdeps:
             jdeps.append(java_info.outputs.jdeps)
 
-    _run_merge_jdeps_action(
-        ctx = ctx,
-        toolchains = toolchains,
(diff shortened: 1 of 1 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -694,15 +694,19 @@ def _run_kt_java_builder_actions(
...
-    _run_merge_jdeps_action(
-        ctx = ctx,
-        toolchains = toolchains,
-        jdeps = jdeps,
-        deps = compile_deps.deps,
-        outputs = {
-            "output": output_jdeps,
-        },
-    )
+    if jdeps:
```
Bencodes, line 697: We could alternatively provide empty jdeps and run the merge action anyways, which would help clean up the conditional-nesting here.

I personally prefer this approach though, as it skips spinning up the merge action entirely.

restingbull (approved):

Issue #599, opened 2021-10-14
`kt_test` targets fail on Windows


fmeum:
On Windows and with 1.5.0-beta-4, any `kt_jvm_test` target fails with an error of the following form:
```
Executing tests from //foobar:descriptor_utils_test
-----------------------------------------------------------------------------
ERROR(tools/test/windows/tw.cc:1293) ERROR: src/main/native/windows/process.cc(202): CreateProcessW("C:\users\runneradmin\_bazel_runneradmin\2gbstkcc\execroot\project\bazel-out\x64_windows-opt\bin\foobar\descriptor_utils_test"): %1 is not a valid Win32 application.
 (error: 193)
ERROR(tools/test/windows/tw.cc:1450) Failed to start test process (arg: C:\users\runneradmin\_bazel_runneradmin\2gbstkcc\execroot\project\bazel-out\x64_windows-opt\bin\foobar\descriptor_utils_test)
```

When inspecting the test executable, it turns out to be a shell script generated from Bazel's `java_stub_template.txt`. While this could run on Windows via `bash.exe`, such a script cannot be launched directly via `CreateProcessW`, hence the error.

Does this mean that running Kotlin tests on Windows is not supported or is there a simple workaround for this problem? Wrapping every Kotlin test in a `java_test` may help.

deannagarcia:
+1 I'm seeing this bug here as well: https://buildkite.com/bazel/protobuf/builds/4038#a77575c0-a256-4ddd-9e5b-7410337bc054 with the same error message.

fmeum:
This is my current workaround: https://github.com/CodeIntelligenceTesting/jazzer/blob/main/bazel/kotlin.bzl

hanneskaeufler-bmw:
+1, seeing this as well. I'm not perfectly sure where the point is, where the file is called in the end, but could we fix this by ensuring this is called through `cmd_bash`, as I am aware of `genrule` for example (https://bazel.build/reference/be/general?hl=en#genrule.cmd)? If anyone could point out how to fix this potentially, one could more easily contribute a fix.

fantzorro:
Still a problem (rules_kotlin v2.0.0)

PR #771, opened 2022-06-12
Enable windows CI

hanneskaeufler:
There are a few issues like #599
which in order to fix without regressing
again it is ugely benefitial to have
some build automation verifying the
intended behaviour. This adds windows
platform testing just as described in
https://github.com/bazelbuild/continuous-integration/tree/master/buildkite

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -63,6 +63,15 @@ tasks:
       - test
     build_targets:
       - //...
+  examples-trivial-windows:
+    name: "Example - Trivial (Windows)"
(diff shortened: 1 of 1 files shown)


Bencodes:
@hanneskaeufler Happy to merge this. Are you able to look at the failing test case so that the windows tests are passing here?

hanneskaeufler:
Hi Ben, yeah that's the idea. I was just trying to see if the CI says the same as my local results, which they do. No promises, but I hope to be looking at this and fixing the issues, yes.

PR #772, opened 2022-06-13
Remove repo mapping from "@" in kotlin_repositories

Wyverald:
The repo "@" must always point to the main repo and cannot be remapped. Attempting to do so will result in a hard error from Bazel 6.0 onwards.

cc @meteorcloudy

### DIFF
--- a/src/main/starlark/core/repositories/initialize.release.bzl
+++ b/src/main/starlark/core/repositories/initialize.release.bzl
@@ -99,7 +99,6 @@ def kotlin_repositories(
         parent = KOTLIN_RULES,
         repo_mapping = {
             "@dev_io_bazel_rules_kotlin": "@%s" % KOTLIN_RULES.workspace_name,
-            "@": "@%s" % KOTLIN_RULES.workspace_name,
         },
     )
 


Bencodes (approved):

src/main/starlark/core/repositories/initialize.release.bzl
```diff
@@ -99,7 +99,6 @@ def kotlin_repositories(
         parent = KOTLIN_RULES,
         repo_mapping = {
             "@dev_io_bazel_rules_kotlin": "@%s" % KOTLIN_RULES.workspace_name,
-            "@": "@%s" % KOTLIN_RULES.workspace_name,
```
Bencodes, line 102: @Wyverald how does Bazel surface this error in 6.0? We are using the 6.0 pre releases internally with a repository that's primarily Kotlin and haven't seen issues with this.

Wyverald: It'll be an error that basically says that the empty string is not a valid repository name. This change just landed last week, so you might not have seen it yet; I'm only fixing this because rules_kotlin is currently red on our CI with Bazel@HEAD.

Wyverald:
The issue is with the usages of '@' here: https://github.com/bazelbuild/rules_kotlin/blob/9e36436cbe88350384d2303cc008bf9897e81099/src/main/starlark/rkt_1_7/jvm/opts.bzl#L1

'@' is fixed to refer the the main repository and shouldn't be remapped. Previously '@' was being mapped to the `io_bazel_rules_kotlin` repo and happened to work, but such a mapping isn't fully supported and breaks in other unexpected places. Instead, this line should load from `@dev_io_bazel_rules_kotlin//src/foo`, since "dev_io_bazel_rules_kotlin" is also being mapped to that repo: https://github.com/bazelbuild/rules_kotlin/blob/9e36436cbe88350384d2303cc008bf9897e81099/src/main/starlark/core/repositories/initialize.release.bzl#L101

I'm not sure how this should be fixed since it appears rules_kotlin is using a very idiosyncratic setup with release archives baked into the repo rule. Do I just update the files in the `src/main/starlark/rkt_*` directories and everything would be magically fixed, or do I need to update some template somewhere?

c-parsons:
Ping. This issue is breaking a number of bazel-downstream-project CI targets, as rules_kotlin is a broken dependency for these targets at HEAD.

restingbull:
Will need more restructuring to work properly. Working on a fix.

restingbull:
@Wyverald When you say `main repo` what do you mean?
Is it the current repository that Bazel is resolving in, or is it the repository that the loader resides in?

restingbull:
Ah, wonderful. Stardoc doesn't respect `repository_mapping`.

restingbull:
Fixed in https://github.com/bazelbuild/rules_kotlin/pull/777

PR #774, opened 2022-06-15
Add no-optimize flag

rockwotj:
In an effort to reduce compile times I'd like to play with turning this flag on during `-c fastbuild` to see if it speeds anything up in our codebase.

See: https://github.com/JetBrains/kotlin/blame/master/compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/K2JVMCompilerArguments.kt#L155-L156


rockwotj:
Oops - no I'll add. We are on the latest compiler so I don't have a ton of motivation to backport to all the right versions 😄

PR #776, opened 2022-06-16
Update -Xuse-fir to -Xuse-k2

rockwotj:
In kotlin 1.7 this flag got renamed and there are deprecation warnings using the old name.

Note that this flag doesn't actually work because they don't support plugins in K2 yet.

See: https://cs.github.com/JetBrains/kotlin/blob/6fc27c22f44d8051bf2e89f1657c0d9741c281c0/compiler/cli/cli-common/src/org/jetbrains/kotlin/cli/common/arguments/CommonCompilerArguments.kt#L298


rockwotj:
Looks like the latest release supports jvm-abi-gen

https://kotlinlang.org/docs/whatsnew1720.html#support-for-kotlin-k2-compiler-plugins

PR #777, opened 2022-06-18
Fix repository mapping to be compliant with Bazel 6

restingbull:


src/main/starlark/core/repositories/initialize.release.bzl
```diff
@@ -93,15 +93,15 @@ def kotlin_repositories(
...
-        name = configured_repository_name,
-        archive = Label("//:%s.tgz" % selected_version),
-        parent = KOTLIN_RULES,
-        repo_mapping = {
-            "@dev_io_bazel_rules_kotlin": "@%s" % KOTLIN_RULES.workspace_name,
-            "@": "@%s" % KOTLIN_RULES.workspace_name,
-        },
-    )
+    if configured_repository_name:
+        rules_repository(
```
Bencodes, line 97: Does this need to error out if a valid `configured_repository_name` isn't provided?

src/main/starlark/core/repositories/initialize.bzl
```diff
@@ -33,4 +33,7 @@ def kotlin_repositories(compiler_release = versions.KOTLIN_CURRENT_COMPILER_RELE
         compiler_release: (internal) version provider from versions.bzl.
     """
     kt_configure()
-    _release_kotlin_repositories(compiler_release = compiler_release)
+    _release_kotlin_repositories(
+        compiler_release = compiler_release,
+        configured_repository_name = "",
```
Bencodes, line 38: Should this be removed if it's always defaulting to an empty string?

restingbull: No, because it will be needed to work with bootstrapping.

src/main/starlark/core/repositories/initialize.release.bzl
restingbull: No -- when using these rules without a release binary, we need to skip this.

Bencodes (approved):

Wyverald:
Hello! What's the release cadence of rules_kotlin like? When is it possible to have a release containing this fix?

Protobuf has a dependency on rules_kotlin (https://github.com/protocolbuffers/protobuf/blob/7a07be5ad5e5dfe5af65f8101e99fec9604d15e5/protobuf_deps.bzl#L106-L111) and this is currently breaking the Bazel@HEAD+Downstream pipeline.

Bencodes:
@Wyverald I cut a new release here https://github.com/bazelbuild/rules_kotlin/releases/tag/v1.7.0-RC-1

PR #786, opened 2022-06-24
Allow resources to be passed to kt_android_library

sugarmanz:
Not sure if there will be push back since this technically is only possible due to the underlying implementation of `kt_android_library` delegating to a `kt_jvm_library`. Other than that, it's just a simple change that helps ease the burden of passing `resources` to a `kt_android_library`.


Bencodes (approved):
@cgruber and I looked this over. Thanks for the PR!

Issue #787, opened 2022-06-28
Using more than one android resources library/directory, R not found


Nikolas-LFDesigns:
In our project we use flat resources structure like

```
package:
- ResourceUser.java
- resources_common
- res
```
and BUILD.bazel like this:
```
android_library(
    name = "resource_user",
    resource_files = glob(["res/**"]),
    srcs = ["ResourceUser.java"],
    deps = [
        ":resources_common",
        [...]
    ],
)

android_library(
    name = "resources_common",
    resource_files = glob(["resources_common/**"]),
    [...]
)
```
One even could depend on some common resources from other packages if appropriate.

When trying to replace android_library definition in resource_user from example above with kotlin rules' kt_android_library (even not actually converting anything to kotlin at all and leaving a java file) and building the project, the kotlin compiler seems confused and can't find resource_files' R definitions that are linked in the [resource_user] library (but can find resources from the [resources_common] android library which is in its deps).

An error is as follows:  
```
ERROR: [project_root]/kotlin-android-resources-example/example/src/main/java/examples/android/lib/BUILD.bazel:3:19: KotlinCompile //example/src/main/java/examples/android/lib:example_fragment_kt { kt: 1, java: 0, srcjars: 0 } for k8 failed: (Exit 1): build failed: error executing command bazel-out/host/bin/external/io_bazel_rules_kotlin/src/main/kotlin/build '--flagfile=bazel-out/k8-fastbuild/bin/example/src/main/java/examples/android/lib/example_fragment_kt-kt.jar-0.params'
example/src/main/java/examples/android/lib/MainFragment.kt:16:35: error: unresolved reference: layout
        return inflater.inflate(R.layout.sample_layout, container,  /*attachToRoot=*/false)
                                  ^
example/src/main/java/examples/android/lib/MainFragment.kt:20:39: error: unresolved reference: id
        view.findViewById<TextView>(R.id.text_view).text = getString(R.string.hello_new)
                                      ^
Jun 29, 2022 12:38:07 AM worker request 0
```
Might be related to some internal resource library not properly linked, but .params-0 .params-1 files for an example_fragment_kt-kt look A-OK to me (at least they have all the resources listed here). 

**Rules kotlin:** 1.6.0
**Bazel:** 4.1.0, reproducible even on 5.2.0

**Example which illustrates the case:**  
https://github.com/Nikolas-LFDesigns/kotlin-android-resources-example

**Similar issues:**
[...]

PR #790, opened 2022-06-29
Support multiple `kt_android_library` targets with same package name.

arunkumar9t2:
When two `kt_android_library` or `android_library` has the same package name either provided via `custom_package` or due to same package name in `AndroidManifest.xml`, `R` class fields from current module's resources were not compilable due to order of jars appearing in `--direct_dependencies`. 

The current module's resources jar would contain merged `R` class from all dependencies but because it appears later in the list, the dependencies' `R` class is used which does not contain merged R class. The problem only appears when two targets contains the same package name.

The fix is similar to one [applied](https://github.com/bazelbuild/bazel/blob/525df32b5dc60448e1abcd3d537ed3454a841946/src/main/java/com/google/devtools/build/lib/rules/android/AndroidCommon.java#L512-L513) in Bazel 

This PR ensures the merged resources jar always comes ensuring order in deps attribute

### DIFF
--- a/examples/android/libKtAndroid/BUILD.bazel
+++ b/examples/android/libKtAndroid/BUILD.bazel
@@ -32,8 +32,16 @@ kt_android_library(
     tags = ["trace"],
     visibility = ["//visibility:public"],
     deps = [
+        ":res2",
         "@maven//:androidx_appcompat_appcompat",
         "@maven//:com_google_auto_value_auto_value_annotations",
         "@maven//:org_jetbrains_kotlinx_kotlinx_serialization_runtime",
     ],
+)
+
+android_library(
+    name = "res2",
+    custom_package = "examples.android.lib",
+    manifest = "src/main/AndroidManifest.xml",
+    resource_files = glob(["res2/**"]),
 )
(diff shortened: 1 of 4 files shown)


arunkumar9t2:
Fixes #787

corbinrsmith:
Did you try inverting the base_deps and base_name in https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/internal/jvm/android.bzl#L56?

It would be much less invasive that changing the order inside the rules.

kotlin/internal/jvm/compile.bzl
```diff
@@ -51,6 +51,11 @@ load(
 def _java_info(target):
     return target[JavaInfo] if JavaInfo in target else None
 
+def _is_android_sandwich(current_label_name, target):
+    """Checks if the given target is <target>_base target from android sandwich macro"""
+    is_base_target = current_label_name.removesuffix("_kt") == target.label.name.removesuffix("_base")
```
Nikolas-LFDesigns, line 56: Doesn't work on Bazel versions older than 5.1.0, https://github.com/bazelbuild/bazel/pull/14899

arunkumar9t2: @Nikolas-LFDesigns  thanks, I removed this impl and just switched the order in `deps` attribute. That also worked.

arunkumar9t2:
Thanks @corbinrsmith, that worked as well. Updated PR.

corbinrsmith (approved):

restingbull (approved):

PR #793, opened 2022-07-05
Check if `default_jvm_opts` exists in `ctx.fragments.java`

mai93:
To make the changes in PRs: https://github.com/bazelbuild/rules_kotlin/pull/782 and https://github.com/bazelbuild/rules_kotlin/pull/689 backward compatible, we add a check if `default_jvm_opts` exists in `ctx.fragments.java` before using it.


Wyverald:
Thanks @mai93! @Bencodes if this could be included in the next RC of 1.7.0, that would be great!

PR #797, opened 2022-07-08
Allow make variable substitution in kt_compiler_plugin

Kvarnefalk:
It would be useful being able to use variable substitution in the plugin rule. I had some issues with the `{generatedSources}` and `{generatedClasses}` templates not returning proper directories in the bazel output dirs. To work around this, I added variable substition which allows me to write something like this:

```
kt_compiler_plugin(
    name = "kotlin_semanticdb_plugin",
    id = "my-id",
    options = {
        "targetroot": "$(execpath :some-rule)",
    },
    deps = [
        ":some-rule",
    ],
)
```

which gives me access to the execroot where I can output artifacts produced by the compiler plugin. I couldn't figure out any other ways to do this. The artifacts produced by the plugin will not be used by downstream targets. They will be picked up by a separate tool scraping the bazel-out folder after the build.

Any other ideas of implementing the thing I want or is this a reasonable change?

### DIFF
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -223,7 +223,9 @@ def kt_jvm_library_impl(ctx):
 
 def kt_jvm_binary_impl(ctx):
     providers = _kt_jvm_produce_jar_actions(ctx, "kt_jvm_binary")
-    jvm_flags = ctx.fragments.java.default_jvm_opts
+    jvm_flags = []
+    if hasattr(ctx.fragments.java, "default_jvm_opts"):
+        jvm_flags = ctx.fragments.java.default_jvm_opts
     jvm_flags.extend(ctx.attr.jvm_flags)
     _write_launcher_action(
         ctx,
@@ -275,7 +277,9 @@ def kt_jvm_junit_test_impl(ctx):
                         test_class = elements[1].split(".")[0].replace("/", ".")
                         break
 
-    jvm_flags = ctx.fragments.java.default_jvm_opts
+    jvm_flags = []
+    if hasattr(ctx.fragments.java, "default_jvm_opts"):
+        jvm_flags = ctx.fragments.java.default_jvm_opts
     jvm_flags.extend(ctx.attr.jvm_flags)
     coverage_metadata = _write_launcher_action(
         ctx,
@@ -380,7 +384,7 @@ def kt_compiler_plugin_impl(ctx):
     for (k, v) in ctx.attr.options.items():
         if "=" in k:
             fail("kt_compiler_plugin options keys cannot contain the = symbol")
-        options.append(struct(id = plugin_id, value = "%s=%s" % (k, v)))
+        options.append(struct(id = plugin_id, value = "%s=%s" % (k, ctx.expand_location(v))))
 
     if not (ctx.attr.compile_phase or ctx.attr.stubs_phase):
         fail("Plugin must execute during in one or more phases: stubs_phase, compile_phase")


restingbull:
I'm not against variable substitution on plugins, though the usecase seems a bit non-hermetic.

>  I had some issues with the {generatedSources} and {generatedClasses} templates not returning proper directories in the bazel output dirs. 
Those directories are folded into the rule output jars, and namespaced to work with multiplexing. What was the expectation of their placement?

> They will be picked up by a separate tool scraping the bazel-out folder after the build.
This is non-ideal -- especially for remote builds. If the plugin isn't generating classes or sources, could you provide more information about the compiler plugin usage?

If I understand the example, the compiler plugin is writing to the directory of `execpath :some-rule`?

Kvarnefalk:
Thanks for taking a look. Happy to learn if there's a better approach to solve my problem.

> Those directories are folded into the rule output jars, and namespaced to work with multiplexing. What was the expectation of their placement?

Makes sense. I thought I would be able to access the generatedSources/classes somewhere on the Bazel build output paths. But realising now that folding them into the jar 

> This is non-ideal -- especially for remote builds. If the plugin isn't generating classes or sources, could you provide more information about the compiler plugin usage?
> If I understand the example, the compiler plugin is writing to the directory of execpath :some-rule?

Yeah that's essentially what I'm doing, letting the compiler plugin write to `execpath :some-rule`. 

**My usecase**: I'm looking to index the source code of my project to have it accessible outside of the IDE. Similar to [Kythe](https://kythe.io/), [LSIF](https://lsif.dev/) and [Glean](https://glean.software/). In this particular case, it means running the [lsif-kotlin](https://github.com/sourcegraph/lsif-kotlin) compiler plugin.

That plugin runs in the compile phase and writes [semanticdb](https://scalameta.org/docs/semanticdb/guide.html) files, capturing as much information about the source code as possible in, essentially a simple proto message. There are usually one semanticdb file per source code file, and therefore we need a post-processing task to collect all of these semanticdb files into a single "source code database", where all types and xrefs are resolved.

When using this approach, I thought I could read the build-event file post-build to identify newly written semanticdb files. But it sounds like you think this approach could give me trouble with remote execution and non-hermeticity?

Is my use case clear? Do you have any other ideas on how I could solve this? Could I potentially somehow use aspects to pick up these files on the go as they are generated by the compiler plugin? Happy to hear your thoughts

arunkumar9t2:
@Kvarnefalk Using aspects does seem like a better approach since you don't need to navigate the file system after the build completes. 
I took a shot at this [here](https://github.com/arunkumar9t2/bazel-playground/blob/9792ec4a662f846f494a90e46ab5b4ed7c7ae1f4/tools/generated_sources_collector/generated_sources_collector.bzl) where I use a simple aspect to collect all `java_common.source_jars` and a companion rule to merge them into single jar. Probably directly targeting the `KtJvmInfo` provider might be better.

I used the `anvil` compiler plugin which uses `{generatedSources}` and I can see the files generated by the compiler plugin in the merged source jar. Similarly you could probably filter all the semanticdb files from it.

restingbull:
Sounds like `kt_compiler_plugin` could use a resource generation target directory (e.g. `{resources}`, which is added to the compiled jar resources). That would let you either use an aspect or create a deploy jar to grab all the files.

Writing to another task's execution directory is a very bad idea for a lot of reasons -- It's only possible because sandboxing is pretty inefficient, and when remote it's completely impossible.

Kvarnefalk:
Thanks a lot for the input and different suggestions for paths forward. Let me close this and take a stab at those on my own. Cheers

PR #800, opened 2022-07-11
Correcting the version in README.md 

bhushan-helloFresh:


Bencodes (approved):

Bencodes:
Actually the version already exists in the URL and shouldn't be in the version variable as well, otherwise the URL won't link to a valid download.

Bencodes:
Going to close this because the `v` symbol is already included in the url itself, not `rules_kotlin_version`. Feel free to reopen this if these assumptions are incorrect.

Issue #801, opened 2022-07-12
Add additional ``kt_kotlinc_options``


brettwooldridge:
My company needs support for the following ``kotlinc`` compiler options:
```
-Xno-call-assertions       Don't generate not-null assertions for arguments of platform types
-Xno-param-assertions      Don't generate not-null assertions on parameters of methods accessible from Java
-Xno-receiver-assertions   Don't generate not-null assertion for extension receiver arguments of platform types
```
A pull request implementing them is coming shortly.

brettwooldridge:
@Bencodes Thanks.

PR #802, opened 2022-07-12
Add kt_kotlinc_options for additional bytecode generation options

brettwooldridge:
This is a pull request addressing #801.

It would be great if this can make it into ``v1.7.0``. These are options that affect bytecode generation and, in specific performance-critical code, measurably increases performance.


Bencodes:
@brettwooldridge thanks for the pull request. Can you back fill these options into the older Kotlin `opts.bzl` files?

brettwooldridge:
@Bencodes Sure.

restingbull (approved):

PR #806, opened 2022-08-01
cleanup: don't pull pre-release java templates

manuelnaranjo:
java template is exposed by bazel it self, and should point to whatever
the current release of bazel has, instead of pulling a pre-5.x version.

### DIFF
--- a/kotlin/internal/jvm/jvm.bzl
+++ b/kotlin/internal/jvm/jvm.bzl
@@ -136,7 +136,8 @@ _implicit_deps = {
     ),
     "_java_stub_template": attr.label(
         cfg = "exec",
-        default = Label("@kt_java_stub_template//file"),
+        default = Label("@bazel_tools//tools/java:java_stub_template.txt"),
+        allow_files = True,
     ),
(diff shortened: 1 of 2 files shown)


restingbull:
Silly question -- is that template exposed on all versions of Bazel?

manuelnaranjo:
> 



> Silly question -- is that template exposed on all versions of Bazel?

Versions lower than 5.X don't support it, it was added for rules_kotlin almost a year ago https://github.com/bazelbuild/bazel/pull/13790 https://github.com/bazelbuild/bazel/pull/12994

So this change would break compatibility of rules_kotlin with pre 5 bazel, if that's a requirement I could try to make something that's portable, but how long are you going to keep that compatibility

restingbull:
[December 2023. ](https://blog.bazel.build/2020/11/10/long-term-support-release.html)

The template is pretty stable, as it's mostly just small improvement to the bash sniffing -- which is why I haven't been too worried.

manuelnaranjo:
Alright, let's drop this then

PR #807, opened 2022-08-01
Mark receiver parameter type as referenced.

aeremin:
At the moment, jdeps generator will miss some dependencies for this code:

```kotlin
package com.example

import io.ktor.server.application.*
import io.ktor.server.engine.*
import io.ktor.server.netty.*
import io.ktor.server.response.*
import io.ktor.server.routing.*

fun main() {
    // embeddedServer signature:
    // public fun <TEngine : io.ktor.server.engine.ApplicationEngine, TConfiguration : io.ktor.server.engine.ApplicationEngine.Configuration>
    //   embeddedServer(factory: io.ktor.server.engine.ApplicationEngineFactory<TEngine, TConfiguration>,
    //                  port: kotlin.Int /* = compiled code */,
    //                  host: kotlin.String /* = compiled code */,
    //                  watchPaths: kotlin.collections.List<kotlin.String> /* = compiled code */,
    //                  configure: TConfiguration.() -> kotlin.Unit /* = compiled code */,
    //                  module: io.ktor.server.application.Application.() -> kotlin.Unit): TEngine { /* compiled code */ }
    embeddedServer(Netty, port = 8081, host = "0.0.0.0") {
        // `this` is `io.ktor.server.application.Application` here - we depend on it.
        log.info("Hello world!")
        routing {
            get("/") {
                call.respondText("Hello Cloud World!")
            }
        }
    }.start(wait = true)
}
```

In particular, `io.ktor.server.application.Application` is not referenced anywhere directly, but we still depend on it to understand what are `log` and `routing`.

To do that - we need to collect types representing receiver arguments.

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt
+++ b/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt
@@ -139,7 +139,7 @@ class JdepsGenExtension(
         is FunctionDescriptor -> {
           resultingDescriptor.returnType?.let { addImplicitDep(it) }
           resultingDescriptor.valueParameters.forEach { valueParameter ->
-            addImplicitDep(valueParameter.type)
+            collectTypeReferences(valueParameter.type, isExplicit = false)
           }
           val virtualFileClass = resultingDescriptor.getContainingKotlinJvmBinaryClass() as? VirtualFileKotlinClass
             ?: return
@@ -217,8 +217,14 @@ class JdepsGenExtension(
      * are other types required for compilation such as supertypes and interfaces of those explicit
      * types.
      */
-    private fun collectTypeReferences(kotlinType: KotlinType, collectSuperTypes: Boolean = true) {
-      addExplicitDep(kotlinType)
+    private fun collectTypeReferences(kotlinType: KotlinType,
+                                      isExplicit: Boolean = true,
+                                      collectSuperTypes: Boolean = true) {
+      if (isExplicit) {
+        addExplicitDep(kotlinType)
+      } else {
+        addImplicitDep(kotlinType)
+      }
 
       if (collectSuperTypes) {
         kotlinType.supertypes().forEach {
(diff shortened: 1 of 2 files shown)


Bencodes:
Thanks for the PR @aeremin! Are you able to add a test case to `KotlinBuilderJvmJdepsTest` for receiver parameter type?

aeremin:
@Bencodes I have added a test, but while working on it understood that my initial fix was nonviable - it will only work if receiver parameter's `this` is explicitly referenced (by literally writing `this` which nobody really does).
In reality, we need to do a better type expansion by calling `collectTypeArguments` instead of just `addImplicitDep`. I've updated the PR accordingly.

BTW, I've noticed that this file can benefit from some cleanups (e.g. `collectSuperTypes` is unused, many of casts aren't needed - e.g. `(resolvedCall.resultingDescriptor as PropertyImportedFromObject)`). Are you interested in clean-up PR's?

Bencodes:
> it will only work if receiver parameter's this is explicitly referenced

This might be a test case worth adding as well if it potentially changes how jdeps are interpreted.

> Are you interested in clean-up PR's?

Absolutely! The person that was originally working on jdeps support inside of rules_kotlin isn't able to work on this anymore, so this code has largely been unmaintained since then. We'd be happy to see PRs for any other cases that aren't yet supported by jdeps as well.

aeremin:
> This might be a test case worth adding as well if it potentially changes how jdeps are interpreted.
Now that I reverted the addition of `ParameterDescriptor` special case it actually doesn't really matter - we will get all dependencies via `FunctionDescriptor` anyway.

>We'd be happy to see PRs for any other cases that aren't yet supported by jdeps as well.
Sounds good! Let's finish with this one then and then I'll do some cleanups and try to find other cases where jdeps are currently insufficient (I can think of some potential cases, but not yet 100% sure).

restingbull (approved):

Bencodes (approved):

Bencodes:
@aeremin just landed this. Thanks for the contribution!

aeremin:
Thanks folks! https://github.com/bazelbuild/rules_kotlin/pull/819 is a cleanup I promised above.

aryeh-looker:
I was bummed to see that I'm still having issues with `ktor` symbols resolving when using the latest release candidate, `1.7.0-RC-3`.

e.g. this compiles and works fine, but reports errors in the IDE

image

Any thoughts? Can file another ticket.

aeremin:
@aryeh-looker could you file a separate ticket (with a full code snippet) and ping me there? I can take a look.

Bencodes:
@aryeh-looker what's the actual error being shown here when you hover over the red? Is it just complaining that the types aren't available?

PR #810, opened 2022-08-03
pass exec_properties to underlying android_local_test macro

Augustyniak:
`exec_properties` is supported by `android_local_test`, it should be supported by `kt_android_local_test` too.

### DIFF
--- a/kotlin/internal/jvm/android.bzl
+++ b/kotlin/internal/jvm/android.bzl
@@ -94,6 +94,7 @@ def kt_android_local_test(
         shard_count = None,
         visibility = None,
         testonly = True,
+        exec_properties = None,
         **kwargs):
     """Creates a testable Android sandwich library.
 
@@ -119,4 +120,5 @@ def kt_android_local_test(
         manifest_values = manifest_values,
         tags = kwargs.get("tags", default = None),
         testonly = testonly,
+        exec_properties = exec_properties,
     )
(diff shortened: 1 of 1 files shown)


kotlin/internal/jvm/android.bzl
```diff
@@ -119,4 +120,5 @@ def kt_android_local_test(
         manifest_values = manifest_values,
         tags = kwargs.get("tags", default = None),
         testonly = testonly,
+        exec_properties = exec_properties,
```
Bencodes, line 123: Wouldn't we want to also pass this to the rest of the targets that are created here so that the `exec_properties` are consistent for all targets?

Augustyniak: Makes sense 👍 I think that your comment should be addressed in https://github.com/bazelbuild/rules_kotlin/pull/810/commits/9c922b529e2de2fb25ede00d8c280444d2487e32

Bencodes: https://github.com/bazelbuild/rules_kotlin/pull/810/files#diff-686482fb7b1546f7ba4e615a0e5f45354cf1f74934c3a31165571540ff96796dR68 the `kt_android_library` macro could use this property as well so that the two macros provide consistent attributes.

Augustyniak: OK, this time it should _really_ be addressed in https://github.com/bazelbuild/rules_kotlin/pull/810/commits/9601771bab57aa03c34d9d1fff2266f8275d47d0

PR #826, opened 2022-08-20
[KtLint] Add CI job for enforcing Kotlin formatting

Bencodes:
Adding a dedicated CI step for running KtLint to dog food our own tooling and to keep all of our Kotlin sources aligned in the codebase.


Bencodes:
This will need to get merged after https://github.com/bazelbuild/rules_kotlin/pull/828 assuming we are all in agreement that the `.editorconfig` looks consistent with what we expect our Kotlin sources to look like.

restingbull (approved):

PR #829, opened 2022-08-24
Run KSP using a kt_compiler_plugin

Bencodes:
This is a second attempt at getting a less intrusive version of KSP working with rules_kotlin based on some feedback in the previous PR.

This approach relies on a kt_compiler_plugin to pass the KSP compiler into builds.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -657,7 +657,7 @@ def _run_kt_java_builder_actions(
             associates = associates,
             compile_deps = compile_deps,
             deps_artifacts = deps_artifacts,
-            annotation_processors = [],
+            annotation_processors = annotation_processors,
             transitive_runtime_jars = transitive_runtime_jars,
             plugins = plugins,
             outputs = outputs,
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -100,6 +100,12 @@ tasks:
     working_directory: examples/jetpack_compose
     test_targets:
       - //...
+  example-ksp:
+    name: "Example - KSP"
+    platform: ubuntu1804
+    working_directory: examples/ksp
+    test_targets:
+      - //...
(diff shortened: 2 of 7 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -653,7 +653,7 @@ def _run_kt_java_builder_actions(
             associates = associates,
             compile_deps = compile_deps,
             deps_artifacts = deps_artifacts,
-            annotation_processors = [],
+            annotation_processors = annotation_processors,
```
Bencodes, line 656: Pure hacks to run KSP under the compilation pass of Kotlinc instead of the Kapt pass.

restingbull (approved):

jbarr21:
@Bencodes will this KSP integration support mixed src sets? two questions that come to mind:

1) are all *.kt and *.java sources passed to KSP processors?
2) will KSP processors still be invoked for Kotlin targets if they only consist of Java sources?

Bencodes:
> are all *.kt and *.java sources passed to KSP processors?

They should be. `_run_kt_builder_action` doesn't do any filtering on specific source types and just passes `srcs.all_srcs` into the `KotlinCompile` action.

> will KSP processors still be invoked for Kotlin targets if they only consist of Java sources?

No it will not. The `_run_kt_builder_action` call that creates the KotlinCompile action is gated by the `has_kt_sources` check. So you will need at least one Kotlin source file in your target to trigger KotlinCompile with KSP.

Kernald:
Any update on this? More and more libraries stop shipping KAPT processors in favour of KSP, this is becoming increasingly important to have.

ajanuar:
Any update, @Bencodes ?

ajanuar:
I have using this PR as patch for supporting KSP and it works well. 
But I have one issue when I do integration with ASwB, generated codes from KSP is not recognized. Any idea what should I check? thanks!

PR #837, opened 2022-09-20
Add an env attribute to kt_jvm_test

meastham:
This is consistent with the `env` attribute supported by native test rules:
https://docs.bazel.build/versions/main/be/common-definitions.html#common-attributes-tests

It's possible to support env_inherit, but only for Bazel versions >= 5.2.0. I'm not sure of the rules_kotlin bazel support policy so I did not go ahead and do this. It would also be possible but a little complicated to support env_inherit for bazel >= 5.2.0 while not breaking users on older versions with a dynamic check as in [this rules_go PR](https://github.com/bazelbuild/rules_go/pull/3256)

### DIFF
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -300,7 +300,7 @@ def kt_jvm_junit_test_impl(ctx):
             direct = ctx.files._java_runtime,
         ),
         # adds common test variables, including TEST_WORKSPACE.
-        testing.TestEnvironment(environment = ctx.attr.env)
+        testing.TestEnvironment(environment = ctx.attr.env),
     )
 
 _KtCompilerPluginClasspathInfo = provider(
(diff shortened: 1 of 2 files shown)


restingbull (approved):

kotlin/internal/jvm/impl.bzl
```diff
@@ -300,7 +300,7 @@ def kt_jvm_junit_test_impl(ctx):
             direct = ctx.files._java_runtime,
         ),
         # adds common test variables, including TEST_WORKSPACE.
-        testing.TestEnvironment({}),
+        testing.TestEnvironment(environment = ctx.attr.env),
```
cheister, line 303: I think this change is only compatible with Bazel 6.0.  Earlier versions of Bazel fail for me with
```
	File "external/io_bazel_rules_kotlin/kotlin/internal/jvm/impl.bzl", line 306, column 32, in kt_jvm_junit_test_impl
		testing.TestEnvironment(environment = ctx.attr.env),
Error in TestEnvironment: TestEnvironment() got named argument for positional-only parameter 'environment'
```

PR #838, opened 2022-09-23
Move versioned repositories to top level `src` dir

restingbull:
A little deck chair moving/yak shaving.

Previously, each repository that contained Kotlin version specific code was placed under `main/<language>`. This turns out to be painful when more languages are introduced. The proposed fix is invert the language/kotlin version ordering, allowing the version specific code in the appropriate language folder.

_Unrelated_.
Also, this PR includes a fix for `archive_repository` not rebuilding -- this is QOL for `rules_kotlin` developers. It's clever. But so is `archive_repository`.


src/main/starlark/release_archive/repository.bzl
```diff
@@ -96,6 +97,16 @@ def archive_repository_implementation(repository_ctx):
             archive = release_artifact,
         )
 
+        # update the timestamp on the repository rule source to ensure it's always re-evaluated.
+        self = workspace
+        for segment in "src/main/starlark/release_archive/repository.bzl".split("/"):
+            self = self.get_child(segment)
+
+        repository_ctx.execute([
+            "touch",
```
Bencodes, line 106: 😨 does't this force you to always rebuild rules_kotlin locally, even if there isn't a change?

restingbull: Yes, but in practice it's not noticeable. It's all cached.

Bencodes: Something to consider - there are projects/developers that use rules_kotlin at HEAD, so this change might introduce overhead in their builds. Dagger is one example that comes to mind https://github.com/google/dagger/blob/master/WORKSPACE#L123-L142

restingbull: Try it out -- in my testing it's basically a noop. This effectively re-instates the classic, non-packaged, release behaviour of building from source. As for the concerns of memory pressure (e.g. two bazel servers running), that was already there due to the first build. `archive_repository` doesn't do a `shutdown`.

PR #842, opened 2022-10-07
Enables compiling against direct dependencies only - kotlin

nkoroste:
Defines `--experimental_prune_transitive_deps` that will allow you to make java and android to only use direct dependancies when compiling a given target. This significantly speeds up incremental compilations at the cost of adding explicit deps to all targets. 

Relates to https://github.com/bazelbuild/bazel/pull/16426


kotlin/internal/jvm/compile.bzl
```diff
@@ -101,16 +102,26 @@ def _jvm_deps(toolchains, associated_targets, deps, runtime_deps = []):
             ",\n ".join(["    %s" % x for x in list(diff)]),
         )
     dep_infos = [_java_info(d) for d in associated_targets + deps] + [toolchains.kt.jvm_stdlibs]
+
+    # Reduced classpath, exclude transitive deps from compilation
+    if (ctx.attr._experimental_prune_transitive_deps[BuildSettingInfo].value):
```
Bencodes, line 107: Would it be better to expose this at the `toolchains.kt` level so that there's a unified place to look up these high level build settings?

kotlin/internal/jvm/BUILD
```diff
@@ -33,3 +33,13 @@ bzl_library(
         "@build_bazel_rules_android//android",
     ],
 )
+
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
```
Bencodes, line 37: @nkoroste I think we need to include `bazel_skylib` in the release archive as a release dependency.

kotlin/internal/jvm/compile.bzl
nkoroste: You're talking about `kotlin/internal/toolchains.bzl` right? defining this in toolchains will prevent you from passing this flag on CLI, no?
Right now you're able to do something like this:
```
bazel build --@io_bazel_rules_kotlin//kotlin/internal/jvm:experimental_prune_transitive_deps=True //foo
```

Bencodes: Ah yes you are actually right.

kotlin/internal/jvm/BUILD
nkoroste: Not sure how releases work to be honest so I would differ to your expertise

lukaciko:
@nkoroste does this replace `experimental_reduce_classpath_mode`? What is the difference between the two?

kotlin/internal/jvm/android.bzl
```diff
@@ -40,18 +41,33 @@ def _kt_android_artifact(
...
-        visibility = ["//visibility:private"],
-        exports = base_deps,
-        deps = deps if enable_data_binding else [],
-        enable_data_binding = enable_data_binding,
-        tags = tags,
-        exec_properties = exec_properties,
-        **kwargs
-    )
+    exported_target_labels = [kt_name]
+    if resource_files:
```
Bencodes, line 45: We should also feature flag this, otherwise everyones repos are going to break randomly when they pull in the new rules_kotlin release.

kotlin/internal/jvm/android.bzl
```diff
@@ -40,18 +41,47 @@ def _kt_android_artifact(
...
-        visibility = ["//visibility:private"],
-        exports = base_deps,
-        deps = deps if enable_data_binding else [],
-        enable_data_binding = enable_data_binding,
-        tags = tags,
-        exec_properties = exec_properties,
-        **kwargs
-    )
+    exported_target_labels = [kt_name]
+    if "kt_prune_transitive_deps_incompatible" in tags:
```
nkoroste, line 45:
```suggestion
    if "kt_prune_android_transitive_deps_incompatible" in tags:
```

kotlin/internal/jvm/compile.bzl
rockwotj: That isn't strictly true right? You can always have a config setting that enables the feature in the tool chain.

We do this for conditional multiplex workers 

```
config_setting( 
     name = "kotlinc_multiplex_workers", 
     define_values = { 
         "kotlinc_multiplex_workers": "enabled", 
     }, 
 ) 
  
 define_kt_toolchain( 
     name = "toolchain", 
     api_version = "1.7", 
     experimental_multiplex_workers = select({ 
         ":kotlinc_multiplex_workers": True, 
         "//conditions:default": False, 
     }),
```

Then enable it via `--define=kotlinc_multiplex_workers=enabled` one could also do this as a proper flag. The benefits of this is that we don't need to ship skylib in the release archive and downstream users get full control over how it's toggled and set.

rockwotj:
> @nkoroste does this replace `experimental_reduce_classpath_mode`? What is the difference between the two?


@lukaciko the difference with this PR is that the transitive jars aren't even passed into the action, so you get caching benefits instead of just giving kotlinc a smaller classpath. As far as I can tell, this PR should make `experimental_reduce_classpath_mode` obsolete. Please someone correct me if I'm wrong

See: https://github.com/bazelbuild/rules_kotlin/blob/33c31659a47dd940a836e135c055f456a08d8bf5/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt#L72

arunkumar9t2:
I would not say this makes the reduced claspath mode redundant yet since this forces every target to list their deps which can be problematic for many `@maven` targets which do not define their impl vs api correctly. Eg. Room uses LiveData but does not declare it. 

It is true that not even passing jars as inputs is the best option since we get caching benefits. To do that in starlark however we need few changes from Bazel. 

Bazel already has experimental class path mode called `bazel` which prunes transitive classpath based on jdeps before action registration. But this is not exposed to starlark yet. Rules kotlin already has jdeps and if this [function](https://github.com/bazelbuild/bazel/blob/4f5e325e337957ebea139dc52a00027acbbb572f/src/main/java/com/google/devtools/build/lib/rules/java/JavaCompileAction.java#L253) is available via `java_common` like 

```
jars = java_common.reduce_classpath(transitive_classpath, jdeps)
```

Then we can also do classpath reduction before action registration which will help with compile avoidance. 

Lastly bazel has automatic recovery in default `java_builder` classpath mode. It attempts compilation with direct jars first and if it fails it falls back to full transitive classpath. 

Side note: 

What we are doing internally is we filter classpath only for project targets and for external targets like @maven we don't filter and pass as it is. This for now keeps build.bazel not too verbose. We pass this information to bazel and rules_kotlin via tags.

kotlin/internal/jvm/compile.bzl
nkoroste: You're right, but that requires a lot of setup on the user side that can potentially be error prone. I personally think that having it via skylib is more user friendly and aligns with all other first class citizen flags we support. Though I don't have super strong feeling either way.

nkoroste:
> As far as I can tell, this PR should make experimental_reduce_classpath_mode obsolete. Please someone correct me if I'm wrong

I agree with @arunkumar9t2, it doesn't make it obsolete as we can do reduction on both the `class` level (https://github.com/bazelbuild/bazel/pull/16457) and on the `jar/deps` level which is what this PR is trying to achieve. 

>It is true that not passing jars as inputs is the best option since we get caching benefits. To do that in starlark for rules_kotlin however we need few changes from Bazel.

Are you talking about this https://github.com/bazelbuild/bazel/pull/16426? 

>Side note:
What we are doing internally is we filter classpath only for project targets and for external targets like @maven we don't filter and pass as it is. This for now keeps build.bazel not too verbose. We pass this information to bazel and rules_kotlin via tags.

Do you mind sharing what you have? I guess my PR will take care of both which does have pretty large performance gain for us

kotlin/internal/jvm/compile.bzl
chancila: you can attach the build flag on the toolchain rule, there's no need for this build flag to be an attr for pretty much every kotlin rule...

nkoroste: As Cris recommended, I've moved the flag under `kotlin/settings` and attached it to the toolchain as it's cleaner that way. 
I'll do the same for the jdeps, `kotlin_deps` flag for consistency in a follow up pr.

kotlin/internal/jvm/android.bzl
```diff
@@ -45,16 +46,47 @@ def _kt_android_artifact(
...
+            resource_files = resource_files,
+            exports = base_deps,
+            deps = deps if enable_data_binding else [],
+            enable_data_binding = enable_data_binding,
+            tags = tags,
+            visibility = ["//visibility:private"],
+            **kwargs
+        )
+        exported_target_labels.append(base_name)
+    elif resource_files:
```
Bencodes, line 64: This conditional also needs to check if there is a `manifest`, otherwise manifests might get left behind.

nkoroste: done

Bencodes (approved):

PR #844, opened 2022-10-10
Update rules_kotlin to Kotlin 1.7.20

rockwotj:
See: https://github.com/JetBrains/kotlin/releases/tag/v1.7.20


Bencodes (approved):
Thanks for doing this! Would you also be able to update the jetpack compose example?

rockwotj:
Yes, do I just need to bump versions?

Bencodes:
Merged. Thanks @rockwotj!

PR #849, opened 2022-10-11
JVM build time improvements via tracking unused deps or classes [Kotlin part]

oliviernotteghem:
This is the Kotlin part of deps/classes tracking to achieve compilation avoidance. 
Refer to Java part (TODO) for more info.

### DIFF
--- a/kotlin/internal/toolchains.bzl
+++ b/kotlin/internal/toolchains.bzl
@@ -87,6 +87,7 @@ def _kotlin_toolchain_impl(ctx):
         experimental_strict_kotlin_deps = ctx.attr.experimental_strict_kotlin_deps,
         experimental_report_unused_deps = ctx.attr.experimental_report_unused_deps,
         experimental_reduce_classpath_mode = ctx.attr.experimental_reduce_classpath_mode,
+        experimental_track_class_usage = ctx.attr.experimental_track_class_usage,
         javac_options = ctx.attr.javac_options[JavacOptions] if ctx.attr.javac_options else None,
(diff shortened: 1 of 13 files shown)


src/main/kotlin/io/bazel/kotlin/plugin/jdeps/BUILD.bazel
```diff
@@ -25,6 +25,7 @@ kt_bootstrap_library(
         "//src/main/kotlin/io/bazel/kotlin/builder/utils/jars",
         "//src/main/protobuf:deps_java_proto",
         "@com_github_jetbrains_kotlin//:kotlin-compiler",
+        "@kotlin_rules_maven//:com_google_guava_guava",
```
Bencodes, line 28: How much does this increase the release binary size?

oliviernotteghem: Final `rules_kotlin_release.tgz` went 19Mb->21.8Mb

kotlin/internal/toolchains.bzl
```diff
@@ -236,6 +237,14 @@ _kt_toolchain = rule(
                 "KOTLINBUILDER_REDUCED",
             ],
         ),
+        "experimental_track_class_usage": attr.string(
+            doc = "Whether to track used classes",
+            default = "off",
+            values = [
```
nkoroste, line 243: why not use booleans here?

oliviernotteghem: yeah.. I should change that. a boolean should be fine, not planning to use a third value here anymore..

PR #855, opened 2022-10-28
Example to repro jdeps issue 

nkoroste:
Fails with jdeps:
```
cd examples/jdesp_issue
bazel build //libKtAndroid1:src_main_kt --@io_bazel_rules_kotlin//kotlin/internal/jvm:experimental_prune_transitive_deps=True --@io_bazel_rules_kotlin//kotlin/internal/jvm:kotlin_deps=True
```

Passes without jdeps:
```
cd examples/jdesp_issue
bazel build //libKtAndroid1:src_main_kt --@io_bazel_rules_kotlin//kotlin/internal/jvm:experimental_prune_transitive_deps=True --@io_bazel_rules_kotlin//kotlin/internal/jvm:kotlin_deps=False
```

Error:
```
ERROR: /Users/nkorostelev/Snapchat/Dev/rules_kotlin/examples/jdesp_issue/libKtAndroid1/BUILD.bazel:3:19: KotlinCompile //libKtAndroid1:src_main_kt { kt: 1, java: 0, srcjars: 0 } for darwin_arm64 failed: (Exit 1): build failed: error executing command (from target //libKtAndroid1:src_main_kt)
  (cd /Users/nkorostelev/Snapchat/Dev/.cache/bazel/arm64/f7a5653e44abf3a0528c8c383a926394/execroot/__main__ && \
  exec env - \
    LC_CTYPE=en_US.UTF-8 \
    REPOSITORY_NAME=io_bazel_rules_kotlin \
  bazel-out/darwin_arm64-opt-exec-2B5CBBC6/bin/external/io_bazel_rules_kotlin/src/main/kotlin/build '--flagfile=bazel-out/darwin_arm64-fastbuild/bin/libKtAndroid1/src_main_kt-kt.jar-0.params')
# Configuration: 3222cb75a37bee8592598dc8e5437da815dc34e08dd4cee307b00b4cc8e60a77
# Execution platform: @local_config_platform//:host
    class examples.deps.libktandroid2.KtDummy2, unresolved supertypes: examples.deps.libktandroid3.KtDummy3
```

Probably caused by https://github.com/bazelbuild/rules_kotlin/pull/807 

In theory, I think if the build doesn't fail without jdeps then it should also not fail with jdeps - @aeremin  thoughts?

### DIFF
--- a/examples/jdesp_issue/WORKSPACE
+++ b/examples/jdesp_issue/WORKSPACE
@@ -1,0 +1,45 @@
+local_repository(
+    name = "release_archive",
+    path = "../../src/main/starlark/release_archive",
+)
+
+load("@release_archive//:repository.bzl", "archive_repository")
+
+archive_repository(
+    name = "io_bazel_rules_kotlin",
+    local_path = "../..",
+)
+
+load("@io_bazel_rules_kotlin//kotlin:repositories.bzl", "kotlin_repositories", "versions")
+
+kotlin_repositories()
+
+register_toolchains("//bzl:experimental_toolchain")
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "bazel_skylib",
+    sha256 = versions.SKYLIB_SHA,
+    urls = ["https://github.com/bazelbuild/bazel-skylib/releases/download/%s/bazel-skylib-%s.tar.gz" % (
+        versions.SKYLIB_VERSION,
+        versions.SKYLIB_VERSION,
+    )],
+)
+
+http_archive(
+    name = "build_bazel_rules_android",
+    sha256 = versions.ANDROID.SHA,
+    strip_prefix = "rules_android-%s" % versions.ANDROID.VERSION,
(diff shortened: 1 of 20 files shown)


nkoroste:
> Probably caused by https://github.com/bazelbuild/rules_kotlin/pull/807

I no longer think that this PR is an issue. As it fails just the same when I revert it.

oliviernotteghem:
Correct. Issue is not caused by https://github.com/bazelbuild/rules_kotlin/pull/807, it's caused by JdepsGenExtension invoking supertypes() kotlintype API ([here](https://github.com/bazelbuild/rules_kotlin/blob/48cc40371b071f0c11da4493cfe46617eac4da9b/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt#L246)). This call will cause (lazy) resolution of the type, and will cause an error since corresponding class is missing on classpath.

Why is compilation succeed then? See Pavel's comment on [this issue](https://youtrack.jetbrains.com/issue/KT-44583) 

> having a type with missing supertypes doesn't always mean that the old error will be reported

PR #858, opened 2022-11-02
add KAPT arguments to define_kt_toolchain

mohamadk:
add KAPT arguments to define_kt_toolchain


kotlin/internal/defs.bzl
```diff
@@ -36,6 +36,12 @@ KtJvmInfo = provider(
     },
 )
 
+KaptInfo = provider(
+    fields = {
+        "correctErrorTypes": "correct Error Types https://kotlinlang.org/docs/kapt.html#non-existent-type-correction",
```
arunkumar9t2, line 41: Can we follow snake case i.e `correct_error_types` here to stay consistent with rest of the args?

arunkumar9t2:
Notes:

All available options: https://kotlinlang.org/docs/kapt.html#using-in-cli

Kapt is in maintenance mode at the moment hence we could consider adding options statically as it is not expected to change much between versions. The structure outlined in this PR is similar to what is already done in Kotlin Gradle Plugin for example, the `kapt {}`

```
kapt {
   showProcessorStats = true
   correctErrorTypes = true
}
```

If this structure is accepted we could look into adding support for `showProcessorStats` as well via `-Kapt-dump-processor-timings` https://github.com/JetBrains/kotlin/pull/4280

nkoroste:
Sounds reasonable to me. Also we might want to bring in all the other defaults from Gradle as that the source of truth we should follow at this point. 
cc @restingbull @Bencodes

kotlin/internal/jvm/compile.bzl
```diff
@@ -395,6 +396,7 @@ def _run_kt_builder_action(
     args.add_all("--deps_artifacts", deps_artifacts, omit_if_empty = True)
     args.add_all("--kotlin_friend_paths", associates.jars, map_each = _associate_utils.flatten_jars)
     args.add("--instrument_coverage", ctx.coverage_instrumented())
+    args.add("--kapt_correct_error_types", toolchains.kt.kaptInfo[_KaptInfo].correct_error_types)
```
Bencodes, line 399: We should just pass all the options in here like we do for Kotlinc/Javac/Ksp, otherwise we'd end up having to create new proto entries for every single kapt option that we want to support.

restingbull:
I'm against adding these to the toolchain -- we want to move away from specific plugin arguments being embedded in the core infrastructure.

KSP has shown that the plugin ecosystem is robust enough to foster an entire new ecosystem of plugins on top of the old one.

It would be better to move KAPT (and KSP) into an extendable subsystem, as we've previously discussed.

pswaminathan:
> I'm against adding these to the toolchain -- we want to move away from specific plugin arguments being embedded in the core infrastructure.

Would an alternative be to provide a generic options setup, rather than attributes for specific options? Similar to [`java_library.javacopts`](https://bazel.build/reference/be/java#java_library.javacopts) or [`cc_library.copts`](https://bazel.build/reference/be/c-cpp#cc_library.copts)?

FWIW I would find this setup preferable—I have other flags I'd like to pass to kapt. Requiring separate attributes in the core rules feels like a lot of overhead—new PR each time a new flag is introduced, clutter, etc.

pswaminathan:
@restingbull can you elaborate on the preferred direction? Would having something like `cc_library.copts` or `java_library.javacopts` be workable? Maybe something like `kapt_copts`/`ksp_copts` and passing them directly in the rules?

PR #860, opened 2022-11-04
Pass runfiles paths more robustly

comius:
Pass jvm-abi-gen and kotlin-compiler with system property

This passes runfiles path returned from rootpath function, which is correct in setups with and without using bzlmod.

Use Dep.fromLabel in tests to find dependencies 

The `Dep.fromLabel` works well in combination with `kt_rules_test`. The same
mangling and demangling is used to find runfiles.

This is a slight simplification, which as well prepares the repo for
bzlmod.

In addition workaround for autovalue is replaced with a filegroup trick,
which works with newer rules_jvm_external repository (where visibility
of .jar targets is limited).


src/test/kotlin/io/bazel/kotlin/defs.bzl
```diff
@@ -35,6 +35,7 @@ def kt_rules_test(name, **kwargs):
         "//src/main/kotlin:jdeps-gen",
         "@com_github_jetbrains_kotlin//:annotations",
         "@com_github_jetbrains_kotlin//:jvm-abi-gen",
+        "@com_github_jetbrains_kotlin//:kotlin-compiler",
         "@com_github_jetbrains_kotlin//:kotlin-stdlib",
         "@com_github_jetbrains_kotlin//:kotlin-stdlib-jdk7",
         "@com_github_jetbrains_kotlin//:kotlin-stdlib-jdk8",
```
comius, line 41: cc @fmeum line 45 is where `rootpath` is used

src/main/starlark/core/repositories/bzlmod_setup.bzl
```diff
@@ -0,0 +1,67 @@
...
+    for segment in path.split("/"):
+        root = root.get_child(segment)
+    repository_ctx.symlink(root, ".")
+
+_local_repository = repository_rule(
+    implementation = _local_repository_impl,
+    attrs = {
+        "path": attr.string(mandatory = True),
+    },
+)
```
comius, line 63: cc @meteorcloudy @Wyverald 
Is there a better way to do local repos with bzlmod?

Wyverald: unfortunately, still not -- https://github.com/bazelbuild/bazel/issues/15412 is surprisingly hard to fix, involving a rather large refactor.

comius: Uf, this is also a blocker for Android, with android_sdk_repository being native.

restingbull:
Please split the target refactorings from the bzlmod changes -- it'll be easier to keep track of what moves where.

src/main/starlark/core/repositories/bzlmod_setup.bzl
meteorcloudy: How hard is it to starlarkfiy andorid_sdk_repository rule? /cc @ahumesky @ted-xie

PR #859, opened 2022-11-03
Fix AbstractMethodError in JdepsGenCommandLineProcessor

Bencodes:
```
SEVERE: Compilation failure: compile phase failed:
exception: java.lang.AbstractMethodError: Receiver class io.bazel.kotlin.plugin.jdeps.JdepsGenCommandLineProcessor does not define or inherit an implementation of the resolved method 'abstract void appendList(org.jetbrains.kotlin.config.CompilerConfiguration, org.jetbrains.kotlin.config.CompilerConfigurationKey, java.lang.Object)' of interface org.jetbrains.kotlin.compiler.plugin.CommandLineProcessor.
at io.bazel.kotlin.plugin.jdeps.JdepsGenCommandLineProcessor.processOption(JdepsGenCommandLineProcessor.kt:47)
at org.jetbrains.kotlin.cli.jvm.plugins.PluginCliParserKt.processCompilerPluginsOptions(PluginCliParser.kt:125)
at org.jetbrains.kotlin.cli.jvm.plugins.PluginCliParser.processPluginOptions(PluginCliParser.kt:83)
at org.jetbrains.kotlin.cli.jvm.plugins.PluginCliParser.loadPlugins(PluginCliParser.kt:72)
at org.jetbrains.kotlin.cli.jvm.plugins.PluginCliParser.loadPluginsSafe(PluginCliParser.kt:44)
at org.jetbrains.kotlin.cli.common.CLICompiler.loadPlugins(CLICompiler.kt:183)
at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:69)
at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:52)
at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:92)
at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:44)
at org.jetbrains.kotlin.cli.common.CLITool.exec(CLITool.kt:98)
at io.bazel.kotlin.compiler.BazelK2JVMCompiler.exec(BazelK2JVMCompiler.kt:32)
at jdk.internal.reflect.GeneratedMethodAccessor92.invoke(Unknown Source)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #864, opened 2022-11-04
Pin to language 1.6 version

hanneskaeufler:


restingbull:
Fixed by https://github.com/bazelbuild/rules_kotlin/pull/859

PR #866, opened 2022-11-05
Update rules_jvm_external from 4.2 -> 4.4.2

Bencodes:
This will need some additional work done so that we can access files directly.


Bencodes:
This should be safe to merge now that we are using Bazel 6 where the classpath hacks aren't needed to make dexing happy.

PR #870, opened 2022-11-08
Add verbose tracing information about performance.

restingbull:
This is forked off the 1.6 tag, as that is the current release.


src/main/kotlin/io/bazel/kotlin/builder/toolchain/CompilationTaskContext.kt
```diff
@@ -18,11 +18,13 @@ package io.bazel.kotlin.builder.toolchain
 
 import com.google.protobuf.MessageOrBuilder
 import com.google.protobuf.TextFormat
+import com.sun.management.OperatingSystemMXBean
```
Bencodes, line 21: Are these imports needed still?

Bencodes (approved):

PR #878, opened 2022-12-05
Add bzlmod configuration

comius:
This prepares rules_kotlin for bzlmod.

The versions in setup with/without bzlmod are synchronised.


comius:
cc @meteorcloudy @Wyverald

MODULE.bazel
```diff
@@ -0,0 +1,72 @@
...
+
+bazel_dep(name = "platforms", version = "0.0.4")
+bazel_dep(name = "rules_pkg", version = "0.7.0")
+bazel_dep(name = "bazel_skylib", version = "1.2.1")
+bazel_dep(name = "stardoc", repo_name = "io_bazel_stardoc", version = "0.5.3")
+bazel_dep(name = "rules_java", version = "5.1.0")
+
+kt = use_extension("//src/main/starlark/core/repositories:bzlmod_setup.bzl", "kt_bzlmod_ext")
+
+use_repo(kt, "rkt_1_7")
```
Wyverald, line 16: maybe combine these into one `use_repo` statement?

MODULE.bazel
```diff
@@ -0,0 +1,72 @@
...
+use_repo(kt, "com_github_pinterest_ktlint")
+
+bazel_dep(name = "rules_proto", version = "4.0.0")
+bazel_dep(name = "rules_android", repo_name = "build_bazel_rules_android", version = "0.1.1")
+bazel_dep(name = "rules_jvm_external", version = "4.4.2")
+
+maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
+
+maven.install(
+    name = "kotlin_rules_maven",
```
meteorcloudy, line 39: Are those java dependencies considered "internal dependencies" of rules_kotlin?

MODULE.bazel
```diff
@@ -0,0 +1,72 @@
...
+
+use_repo(kt, "legacy")
+
+use_repo(kt, "com_github_jetbrains_kotlin")
+
+use_repo(kt, "kt_java_stub_template")
+
+use_repo(kt, "com_github_pinterest_ktlint")
+
+bazel_dep(name = "rules_proto", version = "4.0.0")
```
Wyverald, line 32: nit: why are these `bazel_dep`s down here, instead of with the group above?

MODULE.bazel
```diff
@@ -0,0 +1,72 @@
...
+    repositories = [
+        "https://maven-central.storage.googleapis.com/repos/central/data/",
+        "https://maven.google.com",
+        "https://repo1.maven.org/maven2",
+    ],
+)
+
+use_repo(maven, "kotlin_rules_maven")
+
+bazel_dep(name = "rules_python", version = "0.12.0")
```
Wyverald, line 70: is it possible to mark this with `dev_dependency = True`?

src/main/starlark/core/repositories/bzlmod_setup.bzl
```diff
@@ -0,0 +1,71 @@
...
+    http_file(
+        name = "com_github_pinterest_ktlint",
+        sha256 = versions.PINTEREST_KTLINT.sha256,
+        urls = [url.format(version = versions.PINTEREST_KTLINT.version) for url in versions.PINTEREST_KTLINT.url_templates],
+        executable = True,
+    )
+
+def _local_repository_impl(repository_ctx):
+    path = repository_ctx.attr.path
+    root = repository_ctx.path(Label("@//:MODULE.bazel")).dirname
```
meteorcloudy, line 57: Are you trying to get the source root of the main repo here? /cc @Wyverald Is`Label("@//:MODULE.bazel")` going to work? Maybe should be `Label("@@_main//:MODULE.bazel")`?

meteorcloudy (changes requested):
Nice! 
nit: I would also recommend to add an empty`WORKSPACE.bzlmod` file to avoid accidentally depending on any dependencies introduced in WORKSPACE prefix / suffix.

MODULE.bazel
comius: No, because on `load("@rules_python//python:defs.bzl", "py_binary")` is used in release.

MODULE.bazel
comius: This should actually be a dev dependency. Repository doesn't appear in the release. How do I tag that?

MODULE.bazel
comius: Fixed.

MODULE.bazel
comius: Done.

src/main/starlark/core/repositories/bzlmod_setup.bzl
Wyverald: ah right, it should be `Label("@@//:MODULE.bazel")`, since `rules_kotlin` does not have visibility into the root module

ultimately we should probably have a starlarkified `local_repostiory`, despite `new_local_repository` allegedly being impossible to starlarkify.

comius:
> nit: I would also recommend to add an empty`WORKSPACE.bzlmod` file to avoid accidentally depending on any dependencies introduced in WORKSPACE prefix / suffix.

Done

src/main/starlark/core/repositories/bzlmod_setup.bzl
comius: Done.

MODULE.bazel
Wyverald: you can set `dev_dependency = True` on the `use_extension` call

meteorcloudy: Ah, then `use_extension("@rules_jvm_external//:extensions.bzl", "maven", dev_dependency=False)` should just work.

See https://bazel.build/rules/lib/globals#use_extension

comius: Thanks! Done!

meteorcloudy (approved):
Thanks, I have also granted rules_kotlin access to "Publish to BCR" Github app, you can probably also setup that in a later PR. ;)

comius:
@restingbull friendly ping

comius:
> @restingbull friendly ping

unping. I figured out released .tgz doesn't work.

Clement-Jean:
Is there any way I can help with this ? 😊

PR #880, opened 2022-12-13
Enable dumping KAPT processor performance stats

pettermahlen:
This enables passing the [dumpProcessorStats](https://kotlinlang.org/docs/kapt.html#measuring-performance-of-annotation-processors) flag to the KAPT invocation. Note that the official documentation uses the new name `dumpProcessorTimings`, which appears to be the name from version 1.8 onwards.

The `kotlin.md` table was auto-reformatted when I added the documentation - I thought that was a bit easier to read, but I'm happy to undo that change if you prefer the old version/making the change smaller.


arunkumar9t2:
A related change [here](https://github.com/bazelbuild/rules_kotlin/pull/858) which defines a `kapt_options` rule to collect all params for kapt.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
```diff
@@ -198,6 +198,9 @@ internal fun JvmCompilationTask.runPlugins(
           .plus(
             kaptArgs(context, plugins, "stubsAndApt"),
           )
+          .plus(
+            CompilationArgs(info.passthroughFlagsList),
```
restingbull, line 202: I'm a little afraid of this one. Need to go look at the docs and make sure it can't get into arguments.

PR #888, opened 2022-12-16
Enable new jarfs for multiplex workers.

restingbull:
This is another attempt to get the multiplex to work without completely reloading the compiler classpath on every invocation.


Bencodes (approved):

nkoroste:
Let's add the tests from the other PR into this PR and merge it

PR #895, opened 2022-12-17
Use placeholders in progress messages

fmeum:
This allows Bazel to render the labels in progress messages in a more user-friendly way by unmapping mapped repository names to the names used by the main repository, which is especially relevant with Bzlmod.

### DIFF
--- a/kotlin/internal/js/impl.bzl
+++ b/kotlin/internal/js/impl.bzl
@@ -89,7 +89,7 @@ def kt_js_library_impl(ctx):
         executable = toolchain.kotlinbuilder.files_to_run.executable,
         execution_requirements = {"supports-workers": "1"},
         arguments = [args],
-        progress_message = "Compiling Kotlin to JS %s { kt: %d }" % (ctx.label, len(ctx.files.srcs)),
+        progress_message = "Compiling Kotlin to JS %%{label} { kt: %d }" % len(ctx.files.srcs),
         input_manifests = input_manifests,
         env = {
             "REPOSITORY_NAME": _utils.builder_workspace_name(ctx),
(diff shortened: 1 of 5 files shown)


restingbull:
Where is the documentation on this?

fmeum:
It's documented in https://bazel.build/rules/lib/actions#run (see the `progress_message` parameter) and has been added in https://github.com/bazelbuild/bazel/commit/bfa364346890eb7950edb002877d569695544d48, which means that it is available in Bazel 5.0.0 and later.

Bencodes:
> which means that it is available in Bazel 5.0.0 and later.

@fmeum what happens on versions prior to 5.x? Would we need to bump the minimum supported version of rules_kotlin up to 5.x, dropping 4.x?

fmeum:
@Bencodes With Bazel 4, users would see the literal string `%{label}` instead of the label. Given that Bazel 6 is out now, it may be a good time to raise the minimum version, but not if this PR is the only reason to do so. It isn't fixing an important issue, just making progress messages a little nicer with Bazel 6 as well as Bzlmod.

Bencodes:
> @Bencodes With Bazel 4, users would see the literal string `%{label}` instead of the label. Given that Bazel 6 is out now, it may be a good time to raise the minimum version, but not if this PR is the only reason to do so. It isn't fixing an important issue, just making progress messages a little nicer with Bazel 6 as well as Bzlmod.

I'm not against dropping support for 4.x after we cut a final 1.7.0 release of `rules_kotlin`, which we can do this week.

restingbull:
Will need to announce bzl 4 removal.

Bencodes:
https://github.com/bazelbuild/rules_kotlin/issues/903#issuecomment-1374144105

Issue #660, opened 2022-01-15
Release rules_kotlin as a Bazel module


fmeum:
Bazel 5 will bring experimental support for a [new system for external dependency management codenamed "bzlmod"](https://docs.bazel.build/versions/main/bzlmod.html). Making rules_kotlin available within this system requires packaging it as a module and submitting it to the [Bazel Central Registry](https://github.com/bazelbuild/bazel-central-registry).
Is this already being worked on? I would be willing to help, but do not understand the dependency structure well enough to take this on on my own.

cgruber:
We are looking at it, but we have to look at how this affects those who aren't using 5, and aren't in the modules ecosystem. We're open to it, and ultimately will likely implement this, but if the migration path doesn't support some sort of way out for <5.0 bazel, we'll have to wait a release (of bazel) or two before we really get it done.

In the mean-time, we're looking into the mechanism and setting up a plan. I wouldn't count on this before 1.7 of the rules.

comius:
Main problems:
- release tgz contains rkt_1_{4-7}.tgz which additionally describe Kotlin/Java compiler flags
- compiler repository depends back on rules_kotlin, to export .jars with `kt_jvm_import` rules (circular deps are possible in bzlmod, however they are potentially problematic)

Goal: Make rules_kotlin work with and without bzlmod

I believe there should be just two repositories: rules_kotlin and kotlin_compiler. Bzlmod should handle kotlin_compiler as a dependency as naturally as possible. This then makes it easy for the users to select kotlin_compiler version and rules_kotlin version they want to use.

Proposal:
1. Move `kt_jvm_import` (and `kt_js_import`) targets from the kotlin_complier repository into the rules_kotlin repository (just export files instead from kotlin_complier repo) to break dependency edge from compiler repo to rules_kotlin
  - alternative, bet that cyclic deps work and just depend on rules_kotlin (we'd need to release rules_kotlin and kotlin_compiler on bazel-central-registry at the same time)
3. Release kotlin_compiler 1_4 to 1_7 on bazel-central-registry including the additional bzl files contained in rkt_1_{4,..7} , `kt_{kotlinc,javac}_options` rules and `{kotlinc,javac}_options_to_flags`.
   - or a more lightweight .bzl file, containing only a string list of available javac and kotlinc flags
4. Make rules_kotlin/MODULE.bazel that has a dependency on kotlin_compiler (no repos for rkt_1_x)

[...]

PR #896, opened 2022-12-20
Convert strings to label with the correct repository

comius:
Remove `absolute_target` and use `Label` in the top-level bzl scope. This makes `Label` correctly evaluate to point to the repository where bzl file resides.

Bzlmod is more stringent about labels and `absolute_target` implementation doesn't work with it - the created labels are de-funct.

This works towards https://github.com/bazelbuild/rules_kotlin/issues/660
It's one of the minor issues.


kotlin/internal/toolchains.bzl
```diff
@@ -264,9 +260,24 @@ _kt_toolchain = rule(
...
+_KT_DEFAULT_TOOLCHAIN = Label("//kotlin/internal:default_toolchain")
+
 def kt_register_toolchains():
     """This macro registers the kotlin toolchain."""
-    native.register_toolchains(absolute_target("//kotlin/internal:default_toolchain"))
+    native.register_toolchains(str(_KT_DEFAULT_TOOLCHAIN))
+
+# Evaluating the select in the context of bzl file to get its repository
+_DEBUG_SELECT = select({
+    Label("//kotlin/internal:builder_debug_trace"): ["trace"],
```
fmeum, line 271: Note that this raises the minimum required Bazel version to 5.1 since only that version [supports `Label`s as keys in `select` dicts](https://github.com/bazelbuild/bazel/pull/14755).

@comius I am not 100% certain, but thanks to canonical label literals `str(Label(...))` should work here as well while remaining compatible with Bazel < 5.1.

comius: Done.

restingbull (approved):

PR #897, opened 2022-12-20
Remove repository mapping

comius:
Rename all `dev_io_bazel_rules_kotlin` to `io_bazel_rules_kotlin` and remove `repo_mapping`.

bzlmod takes over full control of `repo_mapping`, uses it to control "strict repo dependencies" and thus removes the parameter from repository rule

This undo-es some of the previous work done to prevent mixing of development repository and released repos.

### DIFF
--- a/WORKSPACE.dev.bazel
+++ b/WORKSPACE.dev.bazel
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-workspace(name = "dev_io_bazel_rules_kotlin")
+workspace(name = "io_bazel_rules_kotlin")
 
 load("//src/main/starlark/core/repositories:download.bzl", "kt_download_local_dev_dependencies")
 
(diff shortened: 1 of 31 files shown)


Bencodes (approved):
Lgtm but might want to have @restingbull give this a once over.

restingbull:
I'd rather this wasn't merged. Besides changing unnecessary strings, it can be more cleanly handled with the `release_archive`.

Aside from that, this change is several steps back from proper separation of dev and prod runtimes, which in turn makes it harder to move to a proper bootstrap, and away from the `kt_bootstrap` rules (which don't cache, or work nicely with libraries, etc. etc.) One of the major reasons for the development repository naming is to shake out all the places that the standard `io_bazel_rules_kotlin` was hard coded (this later caused strange bugs)

restingbull:
A cursory glance over the docs, indicates that opts repos may best be mapped as extensions.

comius:
>  several steps back from proper separation of dev and prod runtimes
Sorry, I had no intention to do that. Thanks for providing some more insight on why you have this separation.

I think I can make progress to support bzlmod in a different way, so I'll close this PR for now.

PR #904, opened 2022-12-22
Add missing rules_proto calls for deps, etc.

restingbull:
# Changes
* Download `rules_proto` dependencies
* Fix shas in `rules_proto` dependencies as the new version is broken
* Update `rules_python` for the new `rules_pkg`
* Some steps to simplify version management.

# Notes
* We can probably remove the python dependency entirely -- it appears to do some small archive manipulation. that toolchain needs some love.
* Hey, ho, `rules_proto` has got to go! We should move to json for the worker chatter and away from the using it as a compilation model.

### DIFF
--- a/src/main/starlark/core/repositories/download.bzl
+++ b/src/main/starlark/core/repositories/download.bzl
@@ -22,16 +22,6 @@ def kt_download_local_dev_dependencies():
 
     Must be called before setup_dependencies in the versions.WORKSPACE.
     """
-    maybe(
-        http_archive,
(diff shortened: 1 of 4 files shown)


Bencodes:
> Hey, ho, rules_proto has got to go! We should move to json for the worker chatter and away from the using it as a compilation model.

I can probably take a look at this. It's not super hard to migrate over. We'll just need to have people flip the flag on Bazel 4 (unless we drop it).

Bencodes (approved):

PR #906, opened 2022-12-28
Revert "Enable new jarfs for multiplex workers."

Bencodes:
When `-Werror` is enabled in the Kotlin compiler, the new JAR FS system causes every single Kotlin compilation action to fail because it emits the "this is an experimental feature" as a Kotlin compiler warning.

More context here: https://bazelbuild.slack.com/archives/CE6AZ22NR/p1671837440186329

Reverts bazelbuild/rules_kotlin#888

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -345,15 +345,8 @@ def _run_kt_builder_action(
     for f, path in outputs.items():
         args.add("--" + f, path)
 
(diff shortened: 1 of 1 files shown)


Bencodes:
@restingbull we should probably disable this until we can figure out how to have the Kotlin compiler not emit these messages as warnings.

restingbull:
I'm sorta split on it, TBH. Multiplex is a bit broken.

restingbull (approved):

PR #910, opened 2023-01-05
Support Kotlin 1.8

Bencodes:


src/main/kotlin/io/bazel/kotlin/plugin/SkipCodeGen.kt
```diff
@@ -32,6 +32,7 @@ import org.jetbrains.kotlin.resolve.jvm.extensions.AnalysisHandlerExtension
 /**
  *  SkipCodeGen registers an extension to skip code generation. Must be the last compiler plugin.
  */
+@OptIn(org.jetbrains.kotlin.compiler.plugin.ExperimentalCompilerApi::class)
 class SkipCodeGen : ComponentRegistrar {
```
Bencodes, line 36: This needs to be switched over to the new implementation that's in 1.8

src/test/kotlin/io/bazel/kotlin/builder/tasks/js/KotlinBuilderJsTest.java
```diff
@@ -23,6 +23,6 @@ public void testJsErrorRendering() {
                 () ->
                         builder.runCompilationTask(
                                 it -> it.addSource("AClass.kt", "package something", "class AClass{")),
-                lines -> assertThat(lines.get(0)).startsWith(builder.toPlatform("sources/AClass.kt")));
+                lines -> assertThat(lines).contains(builder.toPlatform("sources/AClass.kt:2:14: error: missing '}")));
```
Bencodes, line 26: There's some additional logging that happens because we aren't using the new JS IR backend. Probably worth making the switch for Javascript compilation in this PR so that we can revert this test back.

dpoluyanov:
Shouldn't be language_version values updated too?

```patch
diff --git a/kotlin/internal/toolchains.bzl b/kotlin/internal/toolchains.bzl
index c7bd8b2..bd69f4c 100644
--- a/kotlin/internal/toolchains.bzl
+++ b/kotlin/internal/toolchains.bzl
@@ -128,6 +128,7 @@ _kt_toolchain = rule(
                 "1.5",
                 "1.6",
                 "1.7",
+                "1.8",
             ],
         ),
         "api_version": attr.string(
```

Bencodes:
@dpoluyanov nice catch https://github.com/bazelbuild/rules_kotlin/pull/923

PR #915, opened 2023-01-13
Fix reference to java_infos in compile.bzl

### LINKED ISSUE #877: `_run_kt_java_builder_actions` references unassigned `java_info` variable
Found this issue while attempting to upgrade from 1.5 to 1.6. If there are annotation processors, we try to access `java_info` which is not assigned: 

```starlark
Traceback (most recent call last):
	File "/private/var/tmp/_bazel_brian/543db2c15cef85941ff9c139e2f8f864/external/io_bazel_rules_kotlin/kotlin/internal/jvm/impl.bzl", line 259, column 44, in kt_jvm_junit_test_impl
		providers = _kt_jvm_produce_jar_actions(ctx, "kt_jvm_test")
	File "/private/var/tmp/_bazel_brian/543db2c15cef85941ff9c139e2f8f864/external/io_bazel_rules_kotlin/kotlin/internal/jvm/compile.bzl", line 491, column 50, in kt_jvm_produce_jar_actions
		outputs_struct = _run_kt_java_builder_actions(
	File "/private/var/tmp/_bazel_brian/543db2c15cef85941ff9c139e2f8f864/external/io_bazel_rules_kotlin/kotlin/internal/jvm/compile.bzl", line 746, column 56, in _run_kt_java_builder_actions
		ap_class_jar = [jars.class_jar for jars in java_info.outputs.jars][0],
Error: local variable 'java_info' is referenced before assignment.
```

avarun42:
`compile.bzl` was incorrectly referencing the variable `java_info` in cases where annotation processors were found, causing compilation to fail as this variable had not been assigned yet. We must first get the `outputs` from each `java_info` inside `java_infos`, and then use a nested list comprehension to get the `jars` list from each `outputs`, flatten the resulting list, and get the `class_jar` field of each `jars` object inside the flattened list.

Fixes #877

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -748,7 +748,7 @@ def _run_kt_java_builder_actions(
     if annotation_processors:
         annotation_processing = _create_annotation_processing(
             annotation_processors = annotation_processors,
-            ap_class_jar = [jars.class_jar for jars in java_info.outputs.jars][0],
+            ap_class_jar = [jars.class_jar for jars in java_infos.outputs.jars][0],
             ap_source_jar = ap_generated_src_jar,
         )
 
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -746,9 +746,10 @@ def _run_kt_java_builder_actions(
 
     annotation_processing = None
     if annotation_processors:
+        jars_list = [java_info.outputs.jars for java_info in java_infos]
         annotation_processing = _create_annotation_processing(
             annotation_processors = annotation_processors,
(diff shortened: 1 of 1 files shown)


restingbull:
Oh, *facepalm*.

It didn't fail more often because of https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/internal/jvm/compile.bzl#L694.

I hate that about python scoping.

restingbull (approved):

avarun42:
Ah yes I was wondering why the tests only started failing after changing the reference. Python scoping is... not the greatest.

Any ETA on when a 1.7.2 can be cut with the fix?

avarun42:
bump on getting this fix released — we're currently unable to use any version of this package starting from 1.6.0 due to this bug

@restingbull

Issue #883, opened 2022-12-14
invalid user-provided repo name ''


Clement-Jean:
I'm trying to use this rule and I simply want to use the default compiler and toolchains. So I added this in my WORKSPACE:

```
rules_kotlin_version = "1.6.0"
rules_kotlin_sha = "a57591404423a52bd6b18ebba7979e8cd2243534736c5c94d35c89718ea38f94"
http_archive(
    name = "io_bazel_rules_kotlin",
    urls = ["https://github.com/bazelbuild/rules_kotlin/releases/download/v%s/rules_kotlin_release.tgz" % rules_kotlin_version],
    sha256 = rules_kotlin_sha,
)

load("@io_bazel_rules_kotlin//kotlin:repositories.bzl", "kotlin_repositories")
kotlin_repositories()

load("@io_bazel_rules_kotlin//kotlin:core.bzl", "kt_register_toolchains")
kt_register_toolchains()
```

but once I build, I have the following error:

```
ERROR: Traceback (most recent call last):
	File "/Users/clement/Git/protobuf/WORKSPACE.bzlmod", line 13, column 20, in <toplevel>
		kotlin_repositories()
	File "/private/var/tmp/_bazel_clement/5892ceef4629d9b2aad305f593ba52a0/external/io_bazel_rules_kotlin/src/main/starlark/core/repositories/initialize.bzl", line 96, column 21, in kotlin_repositories
		rules_repository(
Error in repository_rule: invalid user-provided repo name '': valid names may contain only A-Z, a-z, 0-9, '-', '_', '.', and must start with a letter
```

This is seems weird to me because I check [intialize.bzl](https://github.com/bazelbuild/rules_kotlin/blob/16445ede99d4056e4a9c464e2b8133b2d8d0fa4a/src/main/starlark/core/repositories/initialize.release.bzl#L108) where it uses `rules_repository` and I can't figure out where the repo name could be wrong.

## Edit

This is in the context of bzlmod. Apprently bzlmod has stricter repository name validation ([ref](https://github.com/bazelbuild/bazel-central-registry/issues/320#issuecomment-1350754084))

restingbull:
[bzlmod](https://github.com/bazelbuild/rules_kotlin/pull/878) support is still in progress. Will ping back when it's working.

vadym-sedna:
any progress on this?

gregorycollins:
[...]

PR #917, opened 2023-01-16
Fix configured_repository_name check for Bazel 6

jmthvt:
Attempt to fix https://github.com/bazelbuild/rules_kotlin/issues/883

### DIFF
--- a/src/main/starlark/core/repositories/initialize.release.bzl
+++ b/src/main/starlark/core/repositories/initialize.release.bzl
@@ -96,7 +96,7 @@ def kotlin_repositories(
         if (criteria and compiler_release.version.startswith(criteria.prefix)) or (not selected_version and not criteria):
             selected_version = version
 
-    if configured_repository_name:  # without a repository name, no default kt_* rules repository is created.
+    if configured_repository_name != "":  # without a repository name, no default kt_* rules repository is created.
         rules_repository(
             name = configured_repository_name,
             archive = Label("//:%s.tgz" % selected_version),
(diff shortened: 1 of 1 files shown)


src/main/starlark/core/repositories/initialize.release.bzl
```diff
@@ -96,7 +96,7 @@ def kotlin_repositories(
         if (criteria and compiler_release.version.startswith(criteria.prefix)) or (not selected_version and not criteria):
             selected_version = version
 
-    if configured_repository_name:  # without a repository name, no default kt_* rules repository is created.
+    if configured_repository_name != "":  # without a repository name, no default kt_* rules repository is created.
```
restingbull, line 99: Uuuuuh.... is this change in starlark?

SrodriguezO: Hmm the build failed with:
```
Caused by: java.lang.ClassCastException: class net.starlark.java.eval.NoneType cannot be cast to class java.lang.String (net.starlark.java.eval.NoneType is in unnamed module of loader 'app'; java.lang.String is in module java.base of loader 'bootstrap')
```

I think we need this instead?
```suggestion
    if configured_repository_name and configured_repository_name != "":  # without a repository name, no default kt_* rules repository is created.
```

Clement-Jean:
Any update on this?

Bencodes:
@Clement-Jean this shouldn't be needed anymore as long as you are using one of the newer releases.

PR #920, opened 2023-01-20
[KSP] Support disabling Kapt in _run_kt_builder_action

Bencodes:
Makes invoking Kapt using `_run_kt_builder_action` more explicit so that we aren't unintentionally running Kapt in other actions.


Bencodes:
@restingbull @nkoroste mind taking a look?

PR #922, opened 2023-01-21
[KSP] Expose additional attributes to JvmCompilationTask plugins

Bencodes:
These arguments need to be exposed to the Kotlin compiler plugin in order for tools like KSP to work.

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
@@ -99,8 +99,11 @@ internal fun JvmCompilationTask.plugins(
 
     val dirTokens = mapOf(
       "{generatedClasses}" to directories.generatedClasses,
-      "{stubs}" to directories.stubs,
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
```diff
@@ -99,8 +99,11 @@ internal fun JvmCompilationTask.plugins(
 
     val dirTokens = mapOf(
       "{generatedClasses}" to directories.generatedClasses,
-      "{stubs}" to directories.stubs,
       "{generatedSources}" to directories.generatedSources,
+      "{incrementalData}" to directories.incrementalData,
+      "{stubs}" to directories.stubs,
+      "{temp}" to directories.temp,
+      "{apclasspath}" to inputs.processorpathsList.joinToString(File.pathSeparator),
```
restingbull, line 106: Maybe `process_classpath`? 

`apclasspath` is annotation processor specific

restingbull (approved):

Bencodes:
Closing because there's probably a better direction to go with this.

PR #932, opened 2023-02-03
Update Kotlin compiler to 1.8.10

Bencodes:
https://github.com/JetBrains/kotlin/releases/tag/v1.8.10


restingbull (approved):

brettwooldridge:
@Bencodes @restingbull Thanks for this! When is v1.8 (rules_kotlin) targeted for? I see there is an RC now.

PR #940, opened 2023-03-06
Get rid of duplicates in the --processors and --processorpath arguments.

tjgq:
Since annotation_processors is (roughly) a depset<struct(depset<string>)>, calling `map_each` on it will not deduplicate strings across structs. We need `uniquify = True` to remove duplicates across the entire `add_all`.

It's possible that other arguments suffer from the same issue; I'm only trying to fix a specific problem that's currently preventing http://github.com/oppia/oppia-android.git from upgrading to a newer rules_kotlin.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -363,12 +363,14 @@ def _run_kt_builder_action(
         annotation_processors,
         map_each = _plugin_mappers.kt_plugin_to_processor,
         omit_if_empty = True,
+        uniquify = True,
     )
     args.add_all(
         "--processorpath",
         annotation_processors,
         map_each = _plugin_mappers.kt_plugin_to_processorpath,
         omit_if_empty = True,
+        uniquify = True,
     )
 
     compiler_plugins = [


Bencodes (approved):

BenHenning:
@Bencodes might it be possible to get a new 1.7.x release that includes this change? We'll need to maintain our own build of rules_kotlin in the interim until an official release can get out since this change is required for us to move to 1.7.x (which is needed when moving to Bazel 5.x+).

Bencodes:
> this change is required for us to move to 1.7.x

Did you mean to say 1.8.x here? Or are you looking to have this backported to the 1.7.x release?

BenHenning:
> > this change is required for us to move to 1.7.x
> 
> Did you mean to say 1.8.x here? Or are you looking to have this backported to the 1.7.x release?

I was hoping for a backport to 1.7.x since 1.8 is still in pre-release. That being said, if the stable release of 1.8 is soon planned, I think we could wait for that (I've tested my team's build on patched versions of both and either will work, I just don't want to move us to a pre-release version if we can help it).

PR #942, opened 2023-03-21
Add support for env_inherit

gibfahn:
#### Commits _(oldest to newest)_

6302065 Add support for env_inherit

<br/>

### DIFF
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -315,7 +315,7 @@ def kt_jvm_junit_test_impl(ctx):
             direct = ctx.files._java_runtime,
         ),
         # adds common test variables, including TEST_WORKSPACE.
-        testing.TestEnvironment(environment = ctx.attr.env),
+        testing.TestEnvironment(environment = ctx.attr.env, inherited_environment = ctx.attr.env_inherit),
     )
 
 _KtCompilerPluginClasspathInfo = provider(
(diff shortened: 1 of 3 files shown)


kotlin/internal/jvm/impl.bzl
```diff
@@ -303,7 +303,7 @@ def kt_jvm_junit_test_impl(ctx):
             direct = ctx.files._java_runtime,
         ),
         # adds common test variables, including TEST_WORKSPACE.
-        testing.TestEnvironment(environment = ctx.attr.env),
+        testing.TestEnvironment(environment = ctx.attr.env, inherited_environment = ctx.attr.env_inherit),
```
Bencodes, line 306: `inherited_environment` wasn't added until Bazel 5.2.0 and will likely cause issues for anyone using Bazel 5.0.0 to 5.1.1.

Bencodes: Actually thinking about this more - this should be safe to merge as people should be using the latest LTS release anyways or have a clear upgrade path by doing a minor `5.x` version bump. What do you think @restingbull @nkoroste?

restingbull: DTM

restingbull (approved):

kotlin/internal/jvm/impl.bzl
restingbull: I'm a little bit on: 7 was just pushed, supporting for is reaching.

Folks who need to keep Bazel at for are most likely not updating `rules_kotlin`

PR #950, opened 2023-04-11
Refactor kt_kotlinc_options and kt_javac_options to accept array of opts

Bencodes:
Tl;dr: Refactors `kt_kotlinc_options` and `kt_javac_options` to now accept a generic string array of compiler opts.

Background: While reviewing the existing typed Kotlin and Java compiler opts, I encountered a few cases where the default compiler opts did not mirror the default values of Javac/Kotlinc. Defaulting our users to non-standard values isn't ideal, and easily falls out of date as Jetbrains releases new versions of the Kotlin compiler.

There is still a great deal of value for repo maintainers to have some amount of typing behind the Kotlinc and Javac opts. To preserve this typing while still allowing some flexibility in compiler opts without making assumptions about default values, the `kt_kotlinc_options` and `kt_javac_options` can accept a more generic `kotinc_opts` and `jvm_opts`.

This change does introduce a breaking change that will roll out with `rules_kotlin` 1.8.

Example:
```python
kt_kotlinc_options(
    name = "my_custom_kotlinc_opt",
    jvm_target = "1.8",
    kotlinc_opts = [
        "-no-stdlib",
        "-opt-in=kotlin.Experimental",
        "-opt-in=kotlin.ExperimentalStdlibApi",
        "-Xsam-conversions=indy",
        "-Xlambdas=indy",
    ],
)
```


Bencodes:
We decided to go with another approach here. We can reopen this PR if we decide to revisit in the future.

PR #951, opened 2023-04-11
Update the Jetpack Compose compiler to 1.8.10

Bencodes:
There isn't a compose compiler release yet that is compatible with 1.8.20. Will follow up with another PR once that's released.


restingbull (approved):
I kinda hate the compose release cycle. It's all over the place.

PR #953, opened 2023-04-11
Remove kt_jvm_plugin_aspect now that all versions of Bazel support JavaPluginInfo

Bencodes:
`JavaPluginInfo` was added in Bazel 5.x which means that it exists by default for all versions of Bazel that we officially support in `rules_kotlin` which means that we should no longer be applying the `kt_jvm_plugin_aspect` aspect. It should be safe to remove now.

This also defaults the rest plugins.bzl over to using `JavaPluginInfo` directly.

### DIFF
--- a/kotlin/internal/jvm/plugins.bzl
+++ b/kotlin/internal/jvm/plugins.bzl
@@ -13,14 +13,6 @@
 # limitations under the License.
 _JavaPluginInfo = getattr(java_common, "JavaPluginInfo")
 
-KtJvmPluginInfo = provider(
-    doc = "This provider contains the plugin info for the JVM aspect",
-    fields = {
-        "annotation_processors": "depset of structs containing annotation processor definitions",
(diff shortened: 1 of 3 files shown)


kotlin/internal/jvm/plugins.bzl
```diff
@@ -11,67 +11,40 @@
...
-            transitive = [
-                (t[_JavaPluginInfo] if _JavaPluginInfo in t else t[JavaInfo]).plugins.processor_jars
-                for t in targets
-                if _JavaPluginInfo in t or JavaInfo in t
-            ],
-        )
-    return depset(transitive = [t[KtJvmPluginInfo].transitive_runtime_jars for t in targets if KtJvmPluginInfo in t])
+    return depset(
+        transitive = [
+            (t[_JavaPluginInfo] if _JavaPluginInfo in t else t[JavaInfo]).plugins.processor_jars
```
restingbull, line 43: let me guess -- `None` in `t` is `False`?

restingbull (approved):

PR #954, opened 2023-04-12
Support passing data attribute to android_local_test

Bencodes:
The data field currently is not being passed to the android_local_test and will cause runtime exceptions if you try to reference files being passed in with this attribute.


restingbull (approved):
🤦

PR #955, opened 2023-04-12
Add support for KSP annotation processing via KspCompile

Bencodes:
Adding proper support for KSP annotation processors.

KSP plugins can be defined and given to targets using the `kt_ksp_plugin` rule and can be passed to Kotlin targets in the `plugins` attribute.

With this approach KSP and Kapt can be used in the same module. 

Example:
```
kt_ksp_plugin(
    name = "moshi-kotlin-codegen",
    processor_class = "com.squareup.moshi.kotlin.codegen.ksp.JsonClassSymbolProcessorProvider",
    deps = [
        "@maven//:com_squareup_moshi_moshi",
        "@maven//:com_squareup_moshi_moshi_kotlin",
        "@maven//:com_squareup_moshi_moshi_kotlin_codegen",
    ],
)

kt_jvm_library(
    name = "lib",
    srcs = glob(["*.kt"]),
    plugins = ["//:moshi-kotlin-codegen"],
)
```


Bencodes:
Waiting for this to land so that we can properly find runfiles https://github.com/bazelbuild/rules_kotlin/pull/956

kotlin/internal/jvm/compile.bzl
```diff
@@ -321,6 +321,45 @@ def _run_kapt_builder_actions(
...
+        srcs = srcs,
+        generated_src_jars = [],
+        associates = associates,
+        compile_deps = compile_deps,
+        deps_artifacts = deps_artifacts,
+        annotation_processors = annotation_processors,
+        transitive_runtime_jars = transitive_runtime_jars,
+        plugins = plugins,
+        outputs = {
+            "ksp_generated_java_srcjar": ksp_generated_java_srcjar,
```
Bencodes, line 355: This action currently only supports collecting generated Kotlin source files and resources. If a KSP plugin outputs Java or compiled outputs, this action will not collect them up to be used in the downstream actions.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
```diff
@@ -165,6 +165,53 @@ internal fun JvmCompilationTask.kaptArgs(
...
+
+      val values = arrayOf(
+        "apclasspath" to listOf(inputs.processorpathsList.joinToString(File.pathSeparator)),
+        // projectBaseDir shouldn't matter because incremental is disabled
+        "projectBaseDir" to listOf(directories.incrementalData),
+        // Disable incremental mode
+        "incremental" to listOf("false"),
+        // Directory where class files are written to. Files written to this directory are class
+        // files being written directly from the annotation processor, not Kotlinc
+        "classOutputDir" to listOf(directories.generatedClasses),
```
Bencodes, line 184: `classOutputDir`, `javaOutputDir`, and `javaOutputDir` are currently being ignored when written to. We'll need to look into adding support for this incrementally.

examples/android/libKtAndroid/BUILD.bazel
```diff
@@ -19,12 +19,23 @@ java_plugin(
     deps = ["@maven//:com_google_auto_value_auto_value"],
 )
 
+kt_ksp_plugin_info(
```
restingbull, line 22: Why not `kt_ksp_plugin`?

src/main/starlark/core/repositories/ksp.bzl
```diff
@@ -0,0 +1,62 @@
+_JARS_INSIDE_REPO = [
+    "symbol-processing",
+    "symbol-processing-api",
+    "symbol-processing-cmdline",
+]
+
+def _ksp_compiler_plugin_repository_impl(repository_ctx):
```
restingbull, line 7: Is this due to not using the mvn kotlinc artifacts?

restingbull (approved):

examples/android/libKtAndroid/BUILD.bazel
Bencodes: Ah good catch. Renamed this to just `kt_ksp_plugin`

PR #967, opened 2023-05-10
Check all dependency jars before marking unused

mauriciogg:
A dependency might produce more than one jar (e.g. android targets produce a library jar and a resources jar). If one of those jars is unused, the current logic will mark the dependency as unused.  This change only marks the dependency unused if none of the jars are used.

- Also ran ktlint

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
@@ -72,8 +72,21 @@ class JdepsMerger {
           val deps: Deps.Dependencies = Deps.Dependencies.parseFrom(it)
           deps.getDependencyList().forEach {
             val dependency = dependencyMap.get(it.path)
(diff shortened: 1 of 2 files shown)


Bencodes:
@mauriciogg can you fix the KtLint issues?

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
```diff
@@ -90,9 +90,23 @@ class JdepsMerger {
       }
 
       if (reportUnusedDeps != "off") {
-        val unusedLabels = dependencyMap.values
-          .filter { it.kind == Deps.Dependency.Kind.UNUSED }
-          .mapNotNull { readJarOwnerFromManifest(Paths.get(it.path)).label }
+        val kindMap = mutableMapOf<String, Deps.Dependency.Kind>()
```
Bencodes, line 93: @mauriciogg is there any test coverage for this behavior in `JdepsMergerTest` that can be added?

Bencodes (approved):

PR #968, opened 2023-05-12
Normalize label

mauriciogg:
Normalize the label from the jar the same way the unused_deps tool. See https://github.com/bazelbuild/buildtools/blob/master/unused_deps/unused_deps.go#L181

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
@@ -99,10 +99,10 @@ class JdepsMerger {
           var label = readJarOwnerFromManifest(Paths.get(it.path)).label
           if (label != null) {
             if (label.startsWith("@@") || label.startsWith("@/")) {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
```diff
@@ -96,11 +96,14 @@ class JdepsMerger {
         // so we need to make sure wedon't mart the dependency as unused
         // unless all of the jars are unused.
         dependencyMap.values.forEach {
-          val label = readJarOwnerFromManifest(Paths.get(it.path)).label
-          if (label != null &&
-            kindMap.getOrDefault(label, Deps.Dependency.Kind.UNUSED) >= it.kind
-          ) {
-            kindMap.put(label, it.kind)
+          var label = readJarOwnerFromManifest(Paths.get(it.path)).label
```
Bencodes, line 99: Want to add a quick test case for this as well to `JdepsMergerTest`?

Bencodes (approved):

PR #975, opened 2023-05-16
Do not pass default values of kotlinc options.

restingbull:


src/main/starlark/core/options/opts.kotlinc.test.bzl
```diff
@@ -0,0 +1,34 @@
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "unittest")
+load(":opts.kotlinc.bzl", "KotlincOptions", "kotlinc_options_to_flags")
+
+def _kt_kotlinc_options_test_impl(ctx):
```
Bencodes, line 4: Shouldn't this be in the test directory?

src/main/starlark/core/options/convert.bzl
```diff
@@ -10,11 +10,13 @@ def _to_flags(opts, attr_provider):
         list of flags to add to the command line.
     """
     if not attr_provider:
-        return ""
+        return []
 
     flags = []
     for n, o in opts.items():
         value = getattr(attr_provider, n, None)
+        if value == o.args["default"]:
```
Bencodes, line 18: Is this logic comparing the input value on the user defined `kt_kotlinc_options` target to the default `attr` value specified in the opts value, and skipping if true?

PR #989, opened 2023-07-17
Expose ksp_version

zalewskise:
### What has changed?
- Exposed `ksp_version` so it can be loaded and used to override `ksp_compiler_release` in `kotlin_repositories`

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,20 @@ kt_jvm_library(
 )
 ```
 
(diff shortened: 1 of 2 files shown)


Bencodes (approved):

Bencodes (approved):

Bencodes:
Thanks for the contribution @zalewskise!

PR #996, opened 2023-07-26
Add support for Kotlin 1.9

### LINKED ISSUE #995: Support kotlin 1.9
Currently if I try to register a custom toolchain with the language level set to 1.9 the build fails with the following message
```
ERROR: package contains errors: : //:kotlin_toolchain_impl: invalid value in 'language_version' attribute: has to be one of '1.1', '1.2', '1.3', '1.4', '1.5', '1.6', '1.7' or '1.8' instead of '1.9'
```

agluszak:
Fixes #995


Bencodes:
@agluszak Are you able to take a look at the failing tests here? I'd love to merge this in for you once they are passing.

Bencodes:
@agluszak the NodeJS tests and examples seem to be broken still. Are you able to take a look at these so we can land this and cut a new rules_kotlin release with 1.9.x support baked in?

agluszak:
@Bencodes I already spent several hours trying to debug what's going on to no avail. KotlinJS is way beyond my area of expertise. I asked around what the reason might be for this `Built-in class kotlin.Any is not found` error. It turns out that it's because of missing stdlib. Does the new IR compiler expect a different stdlib format? I don't know. I'm afraid that without some further help I won't be able to solve this.

agluszak:
Ilya Goncharov from JB:

```
Earlier (in legacy compiler), we had jar file (ala kotlinx-coroutines.jar) which had contained *.meta.js file, *.js and *.js.map . It was library layout for legacy artifact
Now we have klib layout for artifacts.
It is archive with extension *.klib and contains mainly binary files (no js files).
And for IR you need klib file
For example [this](https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-core-js/1.7.3/) is kotlinx-coroutines-js module, and it has jar archive and klib
So for IR you need to pick klib extension. Unfortunately I do not know how dependency resolution works in your case, so if it is possble to resolve dependency to klib artifact, it should work
```

agluszak:
Apparently there's no klib in https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-stdlib-js/1.9.10/ ??

restingbull:
> Apparently there's no klib in https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-stdlib-js/1.9.10/ ??

afaict, the stdlib jar contains the klib directory. I'm validating that we can stop using `.meta` and move to klib.

Looks like 1.9 will no longer produce the `.meta` file -- I'm stuck somewhere between "nice undocumented breaking change" and "we should have gone to klib sooner."

agluszak:
> afaict, the stdlib jar contains the klib directory

True, my mistake

zalewskise:
> Apparently there's no klib in https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-stdlib-js/1.9.10/ ??

Seems like it's included from [1.9.20-Beta](https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-stdlib-js/1.9.20-Beta/) though 👀

restingbull:
> Seems like it's included from [1.9.20-Beta](https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-stdlib-js/1.9.20-Beta/) though 👀

I feel a headache coming on.

General update:

I've got it building in the integration test, but it's failing in the example (it produces klib directories, but no js). I'm digging into it.

kotlin/internal/toolchains.bzl
```diff
@@ -131,6 +131,7 @@ _kt_toolchain = rule(
                 "1.6",
                 "1.7",
                 "1.8",
+                "1.9",
```
Bencodes, line 134: We technically need to include 2.0 here now.

Kotlin 1.9 has deprecated a lot of the K2 related flags and are now asking developers to use `language_version = 2.0` to enable the K2 compiler.

Bencodes: Which means that this opt is a no-op now https://github.com/bazelbuild/rules_kotlin/blob/f229f1ce273f63b61f072c1304d2f5efb677bdd4/src/main/starlark/core/options/opts.kotlinc.bzl#L245-L256

agluszak:
@restingbull any update here?

restingbull:
> @restingbull I reformatted again with newer buildifier

Need to upgrade. Whoops.

PR #997, opened 2023-07-28
Ensure that KSP-generated META-INF services included in output jar

### LINKED ISSUE #991: META-INF files generated by KSP processor are missing from output
Issue was found when trying to integrate with [auto-service-ksp](https://github.com/ZacSweers/auto-service-ksp) library. In this case, KSP processor is invoked but META-INF service file is missing from out jar artifact.

REPRO:
1. Clone this branch : [https://github.com/bazelbuild/rules_kotlin/compare/master...oliviernotteghem:rules_kotlin:inner](https://github.com/bazelbuild/rules_kotlin/compare/master...oliviernotteghem:rules_kotlin:inner)
2. Change dir to: `examples/ksp`
3. Run command: `bazel build coffee_lib && tar -tf bazel-bin/coffee_lib.jar`

EXPECTED:
There should be a file under `META-INF/services/`.

ACTUAL:
File is missing.

Notes:
KSP processor is invoked, since, for example, specifying a bogus value for AutoService() param (in CoffeAppService.java) causes a crash in KSP auto-value processor. Also, the following is observed in logs:

`info: [ksp] loaded provider(s): [com.squareup.moshi.kotlin.codegen.ksp.JsonClassSymbolProcessorProvider, dev.zacsweers.autoservice.ksp.AutoServiceSymbolProcessor$Provider]`

jbarr21:
Fixes #991

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
@@ -26,6 +26,7 @@ import io.bazel.kotlin.builder.toolchain.KotlinToolchain
 import io.bazel.kotlin.builder.utils.IS_JVM_SOURCE_FILE
 import io.bazel.kotlin.builder.utils.bazelRuleKind
 import io.bazel.kotlin.builder.utils.jars.JarCreator
+import io.bazel.kotlin.builder.utils.jars.JarHelper.Companion.SERVICES_DIR
 import io.bazel.kotlin.builder.utils.jars.SourceJarExtractor
 import io.bazel.kotlin.builder.utils.partitionJvmSources
 import io.bazel.kotlin.model.JvmCompilationTask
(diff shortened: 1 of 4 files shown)


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
```diff
@@ -499,6 +505,27 @@ private fun JvmCompilationTask.updateBuilder(
     it.build()
   }
 
+/**
+ * Copy generated resources from KSP task into generated folder
+ */
+private fun Iterator<String>.copyServicesToGenClasses(directories: Directories): Iterator<String> {
```
Bencodes, line 511: Thanks for putting this pull request up! Are you able to add a test case or two so that this doesn’t regress in the future?

zalewskise: Proguard files are also missing (added https://github.com/bazelbuild/rules_kotlin/issues/991#issuecomment-1657249340), would it make sense to include everything from `MANIFEST_DIR` instead of just `SERVICES_DIR`?

restingbull: Yes, it would. 

It's hard enough to figure out what the plugins are doing, otherwise.

jvgelder: would be an easy change then right? just remove the `if ("/$SERVICES_DIR" in it)`  ? or will it contain files we don't want ?

restingbull:
superseded by https://github.com/bazelbuild/rules_kotlin/pull/1094

PR #1000, opened 2023-08-01
Remove unnecessary Java runtime dependencies

fmeum:
`kt_jvm_library` no longer depends on a Java runtime for the target platform, which ensures that cross-platform builds targeting platforms without standalone Java runtimes (e.g. Android) can succeed without hacks (currently, Bazel's local_jdk is marked as compatible with any target platform, but that will change in the future).

Work towards https://github.com/bazelbuild/bazel/issues/17085

### DIFF
--- a/kotlin/internal/jvm/jvm.bzl
+++ b/kotlin/internal/jvm/jvm.bzl
@@ -149,6 +149,9 @@ _implicit_deps = {
         default = Label("@bazel_tools//tools/jdk:current_java_runtime"),
         cfg = "exec",
     ),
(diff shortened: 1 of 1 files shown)


fmeum:
@Bencodes Could you take a look? This is needed to finish my work https://github.com/bazelbuild/bazel/pull/18262 with the aim of removing a lot of footguns from Java cross-compilation.

fmeum:
I learned that `@Bencodes` is OOO, @restingbull could you take a look instead?

PR #1006, opened 2023-08-11
Add support for Bzlmod

Bencodes:
Adding support for Bzlmod now that Android is unblocked. What's included in this PR:
- The initial first pass at Bzlmod support
- Updated the `examples/trivial` example to use Bzlmod with a  
- We've aligned the repository namespace with the Github repository name. Once this is released to the Bazel central registry you can import this as a bzlmod dep by doing: `bazel_dep(name = "rules_java", version = "...")` with a `local_path_override`. It's worth noting that this does not yet exercise the release binary that we distribute for releases.
- Enables Bzlmod by default for `rules_kotlin` (excluding the examples)

This PR is largely based on the work that @comius did in https://github.com/bazelbuild/rules_kotlin/pull/902.

https://github.com/bazelbuild/rules_kotlin/issues/660


.bazelversion
```diff
@@ -1 +1 @@
-6.1.1
+7.0.0-pre.20230724.1
```
Bencodes, line 1: Note: I don't think we actually need to be using the latest rolling release here.

dkthezero:
@Bencodes This means after merging this PR I can set up a Kotlin project with Bzlmod like?
```
bazel_dep(name = "rules_kotlin", version = "x")
```
Currently, I finding a way to work around with `io_bazel_rules_kotlin` in MODULE.bzl but not work

Bencodes:
@dkthezero yes it would.

MODULE.bazel
```diff
@@ -0,0 +1,71 @@
+module(
+    name = "io_bazel_rules_kotlin",
```
comius, line 2: Can you call this just "rules_kotlin"? This is what gets exported to BCR: https://github.com/bazelbuild/bazel-central-registry/tree/main/modules

So "io_bazel_" prefix would be a precedent.

"repo_name = " attribute controls how this can be used, that is what you need in the `load` statements and as dependencies.

examples/trivial/MODULE.bazel
```diff
@@ -0,0 +1,30 @@
+module(
+    name = "trival-example",
+)
+
+bazel_dep(name = "rules_java", version = "6.4.0")
+bazel_dep(name = "io_bazel_rules_kotlin")
```
comius, line 6: Should be `rules_kotlin` with optional `repo_name="io_bazel_rules_kotlin`".

comius (changes requested):

comius:
Thanks for your effort!

.bazelrc
```diff
@@ -1,3 +1,5 @@
+common --enable_bzlmod=true
```
Bencodes, line 1: Enabling Bzlmod by default for `rules_kotlin`

MODULE.bazel
```diff
@@ -0,0 +1,67 @@
...
+    version = "1.8.1",
+    compatibility_level = 1,
+    repo_name = "io_bazel_rules_kotlin",
+)
+
+bazel_dep(name = "platforms", version = "0.0.6")
+bazel_dep(name = "bazel_skylib", version = "1.3.0")
+bazel_dep(name = "rules_java", version = "6.4.0")
+bazel_dep(name = "rules_python", version = "0.24.0")
+bazel_dep(name = "rules_cc", version = "0.0.8")
```
Bencodes, line 12: `rules_cc` is needed for the remote execution configs.

restingbull (approved):

src/main/starlark/core/repositories/bzlmod_setup.bzl
```diff
@@ -0,0 +1,62 @@
...
+
+_version_tag = tag_class(
+    attrs = {
+        "version": attr.string(mandatory = True),
+        "sha256": attr.string(mandatory = True),
+    },
+)
+
+def _extra_repositories():
+    # TODO(bencodes) This really needs to be a development dependency
```
Bencodes, line 26: Left a TODO here to move this out into a development dependency. I'll do another pass once this is merged to get these dependencies into better shape so that we can do a production-ready release of rules_kotlin.

PR #1011, opened 2023-08-13
Version kotlinc plugins by major version to handle API changes.

restingbull:
Part one of two.

@nkoroste discovered we had been breaking backwards compatibility by updating the plugins as the compiler plugin api changes.

These separates and compiles each major version of the compiler plugins. The follow up pr will choose the correct plugin for the toolchain -- separating the efforts made sense form a complexity and limiting the scope of the PR point of view.


src/main/starlark/core/repositories/versions.bzl
```diff
@@ -61,13 +83,24 @@ versions = struct(
...
+        kotlinc_version(
+            release = "1.7.22",
+            sha256 = "9db4b467743c1aea8a21c08e1c286bc2aeb93f14c7ba2037dbd8f48adc357d83",
+        ),
+        kotlinc_version(
+            release = "1.6.21",
+            sha256 = "632166fed89f3f430482f5aa07f2e20b923b72ef688c8f5a7df3aa1502c6d8ba",
+        ),
+    ],
     KSP_CURRENT_COMPILER_PLUGIN_RELEASE = version(
```
Bencodes, line 104: KSP also needs to be versioned to the Kotlin compiler.

restingbull: Well, this is going to be fun.

restingbull:
@Bencodes -- Looks like we do have folks working with older versions: https://github.com/bazelbuild/rules_kotlin/issues/1019

PR #1013, opened 2023-08-18
Disable strict deps for Android example

Bencodes:
This experimental feature doesn't work super well for Android projects at the moment and should just be disabled until someone has more time to invest in getting things in better shape.


restingbull (approved):

restingbull:
Removes flake due to jdeps strict deps check.

PR #1014, opened 2023-08-18
Move the KSP and compiler target definitions into rules_kotlin

Bencodes:
Some organizational changes to allow easier support for Bzlmod. This may introduce breaking changes for those that are depending on the targets declared directly inside of the `com_github_jetbrains_kotlin`/`com_github_google_ksp` repositories.

The new compiler target references can be found under the `@io_bazel_rules_kotlin//kotlin/compiler:...` package. We may start to reduce the visibility of these and ask users to eventually start pulling their standard libraries from maven.

Details: With Bzlmod we are unable to easily import `kt_` macros from within a Bazel repository. To work around this we can shuffle the actual `kt_` target creation into `io_bazel_rules_kotlin` instead of inside the declared `com_github_jetbrains_kotlin`/`com_github_google_ksp` repositories.


restingbull:
This is pretty big breaking change, as is. Given that the repository hasn't provided a deprecation warning, I'm not in favor of it as is -- which was why I didn't approve it previously.

I think I see a less invasive way to fix the cyclic dependency.

restingbull (approved):

PR #1017, opened 2023-08-22
Pass through kotlinc options to JS compile

### LINKED ISSUE #1016: Kotlinc options don't pass through to JS Compile
Repro [here](https://github.com/pswaminathan/kotlinc-js-passthrough-example)

* Action: Build src:jvm and src:js
* Expected: Both src:jvm and src:js get -Xmulti-platform specified by x_multi_platform argument to :kt_kotlinc_options
* Actual: src:jvm succeeds, src:js doesn't get this option

pswaminathan:
Fixes #1016

I haven't yet considered what options may need to be e.g. stripped out before sending along. But if there are other glaring holes in this, or reasons to avoid this kind of approach altogether, LMK.

### DIFF
--- a/kotlin/internal/js/impl.bzl
+++ b/kotlin/internal/js/impl.bzl
@@ -17,6 +17,11 @@ load(
     _TOOLCHAIN_TYPE = "TOOLCHAIN_TYPE",
 )
 load(
+    "//kotlin/internal:opts.bzl",
+    "KotlincOptions",
+    "kotlinc_options_to_flags",
+)
+load(
     "//kotlin/internal/utils:utils.bzl",
     _utils = "utils",
 )
@@ -42,9 +47,11 @@ def kt_js_library_impl(ctx):
         _utils.derive_module_name(ctx),
     )
 
+    kotlinc_options = ctx.attr.kotlinc_opts[KotlincOptions] if ctx.attr.kotlinc_opts else toolchain.kotlinc_options
+
     args.add_all(
         "--kotlin_js_passthrough_flags",
-        [
+        kotlinc_options_to_flags(kotlinc_options) + [
             "-source-map",
             "-Xir-produce-klib-dir",
             "-no-stdlib",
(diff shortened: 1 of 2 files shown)


agluszak:
@pswaminathan it fails on CI: https://buildkite.com/bazel/rules-kotlin-kotlin/builds/3519#018a1b24-c95a-4722-bc75-59b655a3e724

Bencodes:
@pswaminathan are you able to look into these failures?

pswaminathan:
Yep! Will take a look.

pswaminathan:
I know the tests don't have any JS examples running, so those aren't representative of this working. But the example fail-case I gave in the linked issue does work with this build. I can add that under e.g. `examples/kmp/trivial` here as an stdlib-only working test case for the JS rules. WDYT?

kotlin/internal/js/impl.bzl
```diff
@@ -42,9 +47,11 @@ def kt_js_library_impl(ctx):
...
     )
 
+    kotlinc_options = ctx.attr.kotlinc_opts[KotlincOptions] if ctx.attr.kotlinc_opts else toolchain.kotlinc_options
+
     args.add_all(
         "--kotlin_js_passthrough_flags",
-        [
+        kotlinc_options_to_flags(kotlinc_options) + [
             "-source-map",
             "-Xir-produce-klib-dir",
```
Bencodes, line 56: Do you know if there will be any issues with duplicate flags being passed here. For example, `kotlinc_options_to_flags` could return `"-Xir-produce-klib-dir"` which is already being set here in impl.bzl.

pswaminathan: If I undestand the implementation of `kotlinc_options_to_flags` correctly, most of these aren't available as kotlinc options, meaning that can't happen. But `-no-sdtlib` could be conflicting if the options include `{"include_stdlibs": {"all","stdlib"}` (and duplicated if `"none"`). So probably worth handling more generally anyway.

Bencodes (approved):

PR #1022, opened 2023-09-07
Add transitive_runtime_jars to KtKvmInfo

scosenza:
Add transitive_runtime_jars to KtKvmInfo so we can store the java_info.transitive_runtime_jars for access in the IntelliJ Bazel Plugin's aspect


kotlin/internal/defs.bzl
```diff
@@ -37,6 +37,7 @@ KtJvmInfo = provider(
         "outputs": "output jars produced by this rule. [intelij-aspect]",
         "language_version": "version of kotlin used. [intellij-aspect]",
         "transitive_compile_time_jars": "Returns the transitive set of Jars required to build the target. [intellij-aspect]",
+        "transitive_runtime_jars": "Returns the transitive set of Jars required to run the target. [intellij-aspect]",
```
Bencodes, line 40: I believe you can already get this from the `JavaInfo` provider that will exist alongside the `KtJvmInfo` provider. Is that not the case?

PR #1023, opened 2023-09-07
[bzlmod] Move providers to live in the kotlinc repo

restingbull:
* Introduce kt_repo_import, a specialized internal rule to import kotlinc binaries and source
* Reading the providers from the kotlinc repo ensures that the provider is pinned to a kotlin version. 

Notes:
* This doesn't replace the kt_jvm_import which does more and interesting things.
* It does expand the js from the jar, which turned out to be the hardest part.


restingbull:
In favor of removing repo in bzlmod.

PR #1036, opened 2023-09-29
Add support for passing ksp options

### LINKED ISSUE #1035: Missing support for passing options to KSP processors
KSP offers a way to pass options to processors, its done with the -P flag, for example:
```
-P plugin:com.google.devtools.ksp.symbol-processing:apoption=key=value
```

Since we don't have a dedicated attr for that and the existing kotlinc opts are not supporting -P it's impossible to pass those options to the KSP plugin.

We probably would want to allow declaring the options on both target and processor level.

idanakav:
Resolves #1035 

Sanity tested with:
```
cd examples/ksp ; bazel aquery 'mnemonic(KotlinKsp, //:coffee_lib)' | grep ksp_opts -A 2

    --ksp_opts \
    'arg1=value1' \
    'arg2=value2' \
```


restingbull:
Does KSP fail if flags for absent plugins are passed?

idanakav:
> Does KSP fail if flags for absent plugins are passed?

No, those are generic KSP options that are going to be forwarded to any KSP processor.
In other words, any KSP processor listed in the `plugins` will receive the same options.

restingbull:
> > Does KSP fail if flags for absent plugins are passed?
> 
> No, those are generic KSP options that are going to be forwarded to any KSP processor. In other words, any KSP processor listed in the `plugins` will receive the same options.

Shoot. Going to need to track the plugin that needs the arguments.

PR #1044, opened 2023-10-06
WIP: Fix missing jdeps by consistently using `collectTypeReferences`

scosenza:
# Problem
Kotlin jdeps are missing in some cases which is preventing Airbnb from enabling certain compile-time pruning optimizations in our fork of the IntelliJ Bazel Plugin. For example, in the below repro case, Airbnb's IntelliJ Bazel Plugin compile time pruning drops the jar containing `Base` which results in the following compilation error:
```
 Kotlin: Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:
        class something.ReferencesClassWithSuperClass, unresolved supertypes: something.Base
```

Without the change in this PR, `ReferencesClassWithSuperClass` has a missing jdep on superclass `Base`:
```kotlin
open class Base

class Derived : Base() {
  fun hi(): String {
    return "Hello"
  }
}

class ReferencesClassWithSuperClass {
    fun stuff(): String {
        return Derived().hi()
    }
}
```

# Solution
Use `collectTypeReferences` instead of directly calling `addImplicitDep` or `addExplicitDep` as `collectTypeReferences` additionally always adds superclasses as implicit deps.

Note that the initial fix originally caused two existing test cases to fail:
```
1) function call return type[enableK2Compiler=false](io.bazel.kotlin.builder.tasks.jvm.KotlinBuilderJvmJdepsTest)
io.bazel.kotlin.builder.toolchain.CompilationStatusException: compile phase failed:

        at io.bazel.kotlin.builder.toolchain.CompilationTaskContext.executeCompilerTask(CompilationTaskContext.kt:135)
```
This existing test failure seems related to both:
- #855
- https://youtrack.jetbrains.com/issue/KT-44583

To fix this test, I added `c.addTransitiveDependencies(depWithReturnTypesSuperType)` as `SomeSuperType` is required to compile `ReferencesClassWithSuperClass`.

### DIFF
--- a/src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
+++ b/src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
@@ -1262,10 +1262,6 @@ class KotlinBuilderJvmJdepsTest(private val enableK2Compiler: Boolean) {
       )
       c.addDirectDependencies(depWithFunction)
       c.addTransitiveDependencies(depWithReturnType)
-      //TODO: Adding a dep on depWithReturnTypesSuperType "fixes" this test which started failing after this jdeps bugfix
-      //      Instead, should our jdeps bug fix only be enabled when "-Xextended-compiler-checks" is set, such that this test would continue passing
-      //      And then we'd add another test which sets "-Xextended-compiler-checks" and has the proper transitive dependency on "depWithReturnTypesSuperType"?
-      //      For more info on "-Xextended-compiler-checks" see: https://youtrack.jetbrains.com/issue/KT-44583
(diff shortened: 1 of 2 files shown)


src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
```diff
@@ -444,6 +444,62 @@ class KotlinBuilderJvmJdepsTest(private val enableK2Compiler: Boolean) {
     assertIncomplete(jdeps).isEmpty()
   }
 
+  @Test
+  fun `kotlin indirect property reference on object 2`() {
```
scosenza, line 448: TODO: Rename test case and add additional tests

scosenza:
Replaced this PR with https://github.com/bazelbuild/rules_kotlin/pull/1045

PR #1045, opened 2023-10-08
Fix missing jdeps by consistently calling `collectTypeReferences`

scosenza:
# Problem
Kotlin jdeps are missing in some cases which is preventing Airbnb from enabling certain compile-time pruning optimizations in our optimized IntelliJ Bazel Plugin. For example, for the following repro case:
```kotlin
open class Base

class Derived : Base() {
  fun hi(): String {
    return "Hello"
  }
}

class ReferencesClassWithSuperClass {
    fun stuff(): String {
        return Derived().hi()
    }
}
```
Without this PR fix, Airbnb's compile time pruning (which relies on Kotlin jdeps) drops the jar containing `Base` which results in the following compilation error:
```
Kotlin: Supertypes of the following classes cannot be resolved. Please make sure you have the required dependencies in the classpath:
        class something.ReferencesClassWithSuperClass, unresolved supertypes: something.Base
```

# Solution
Combine `collectTypeReferences` and `collectTypeArguments` and inline `addExplicitDep` and `addImplicitDep`, to prevent future missing jdep bugs where code paths forgot to traverse superclasses and type-params.

Note that the fix in this PR initially caused the following existing test case to fail:
```
1) function call return type[enableK2Compiler=false](io.bazel.kotlin.builder.tasks.jvm.KotlinBuilderJvmJdepsTest)
io.bazel.kotlin.builder.toolchain.CompilationStatusException: compile phase failed:

        at io.bazel.kotlin.builder.toolchain.CompilationTaskContext.executeCompilerTask(CompilationTaskContext.kt:135)
```
This existing test failure seems related to both:
- #855
- https://youtrack.jetbrains.com/issue/KT-44583

To fix this test, I added `c.addTransitiveDependencies(depWithReturnTypesSuperType)` as `SomeSuperType` is required to compile `ReferencesClassWithSuperClass`.


nkoroste:
Hey @scosenza should we promote this to a proper PR and try to merge it in if it fixes a legit jdeps issue?

scosenza:
Sorry for the delay @nkoroste! I'll take a look now into the Buildkite failures (the tests pass on an Airbnb Linux machine) and then I'll remove the draft status.

restingbull (approved):

PR #1048, opened 2023-10-10
Fix corner cases in handling of android lib when compiling against di…

oliviernotteghem:
…rect dependencies only

https://github.com/oliviernotteghem/rules_kotlin/commit/004189ea53792cad74e0b3b8559d4889707cafbb came a slightly outdated branch. Syncing changes with what we're using internally at Uber.

### DIFF
--- a/kotlin/internal/jvm/android.bzl
+++ b/kotlin/internal/jvm/android.bzl
@@ -1,9 +1,3 @@
-load(
-    "@build_bazel_rules_android//android:rules.bzl",
-    _android_library = "android_library",
-    _android_local_test = "android_local_test",
-)
-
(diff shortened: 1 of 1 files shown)


kotlin/internal/jvm/android.bzl
```diff
@@ -103,7 +95,7 @@ def _kt_android_artifact(
         tags = tags,
         exec_properties = exec_properties,
     )
-    return [base_name, kt_name]
+    return [kt_name, base_name]
```
nkoroste, line 98: why is this needed? that sounds like something is not hermetic

Bencodes:

PR #1054, opened 2023-10-13
Add Android Platforms build flags.

katre:
Part of https://github.com/bazelbuild/bazel/issues/16285.

### DIFF
--- a/examples/android/WORKSPACE
+++ b/examples/android/WORKSPACE
@@ -88,3 +88,5 @@ http_archive(
 load("@robolectric//bazel:robolectric.bzl", "robolectric_repositories")
 
 robolectric_repositories()
+
+android_ndk_repository(name = "androidndk")
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -46,20 +46,24 @@ tasks:
     name: "Example - Android"
     platform: ubuntu1804
     working_directory: examples/android
-    bazel: 7.0.0-pre.20230906.2
+    bazel: last_green
     test_flags:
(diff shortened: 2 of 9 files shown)


examples/android/WORKSPACE
```diff
@@ -88,3 +88,5 @@ http_archive(
 load("@robolectric//bazel:robolectric.bzl", "robolectric_repositories")
 
 robolectric_repositories()
+
+android_ndk_repository(name = "androidndk")
```
Bencodes, line 92: Is the NDK a requirement now?

katre: Before enabling `--incompatible_enable_android_toolchain_resolution`, this was using the default (and useless) `cc_toolchain` from [bazel's CC autodetection](https://cs.opensource.google/bazel/bazel/+/master:tools/cpp/BUILD.toolchains.tpl;drc=46e0be4ef67901cab7896348c89de06396f737b4;l=12).

Since that's no longer appropriate (it claims armeabi-v7a, I don't know if the cc compiler in this image can actually do this), it seemed better to at least accurately declare that every Android binary needs an NDK for analysis. If you don't have any actual `cc_library` dependencies it isn't used (which is why the old one worked).

Ideally the Android rules would not require a CC toolchain if it's not being used, but currently [they are](https://github.com/bazelbuild/rules_android/blob/90d45694c3d98b5702f64911b7b46937fe782ca9/rules/android_binary_internal/attrs.bzl#L135).

katre:
@Bencodes Any further comments?

Bencodes (approved):

PR #1082, opened 2023-12-05
Use launcher stub template from bazel

illicitonion:
This was added in https://github.com/bazelbuild/bazel/pull/12994 (and fixed in https://github.com/bazelbuild/bazel/pull/13790).

This has been included in all Bazel releases since 5.0.0, which is the minimum supported Bazel version of rules_kotlin according to https://github.com/bazelbuild/rules_kotlin/blob/master/README.md

This prevents having a build-time dependency on access to raw.githubusercontent.com.


restingbull (approved):

restingbull:
```

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
--
  | java.lang.NullPointerException: attempted to use --release, but JAVA_HOME is not set
  | at java.base@20.0.2/java.util.Objects.requireNonNull(Objects.java:259)
  | at com.google.turbine.binder.CtSymClassBinder.bind(CtSymClassBinder.java:52)
  | at com.google.turbine.main.Main.bootclasspath(Main.java:309)
  | at com.google.turbine.main.Main.compile(Main.java:142)
  | at com.google.turbine.main.Main.compile(Main.java:133)
  | at com.google.turbine.main.Main.main(Main.java:89)
  | Target //:rules_kotlin_release failed to build


```

I think this is unrelated, but we should sort it out before merging.

illicitonion:
Yep, looks like something changed in Bazel CI? https://github.com/bazelbuild/rules_kotlin/pull/1083 is similarly red.

@meteorcloudy any ideas what changed about Bazel CI that may have prompted these kind of issues?

meteorcloudy:
https://buildkite.com/bazel/rules-kotlin-kotlin

Yes, the postsubmit has been red for a while. I guess it's some incompatible change in Bazel@HEAD. The failing job is using Bazel@HEAD, try to use `bazelisk --bisect` to identify which change breaks you.

Bencodes (approved):

PR #1089, opened 2023-12-10
Add missing KotlinJvmTaskExecutorTest to the test suite

raghulvelusamy:
Add missing KotlinJvmTaskExecutorTest to the test suite

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
@@ -472,7 +472,7 @@ private val Directories.incrementalData
  * Create a new [JvmCompilationTask] with sources found in the generatedSources directory. This should be run after
  * annotation processors have been run.
  */
-internal fun JvmCompilationTask.expandWithGeneratedSources(): JvmCompilationTask =
+fun JvmCompilationTask.expandWithGeneratedSources(): JvmCompilationTask =
   expandWithSources(
     Stream.of(directories.generatedSources, directories.generatedJavaSources)
       .map { s -> Paths.get(s) }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
```diff
@@ -472,7 +472,7 @@ private val Directories.incrementalData
  * Create a new [JvmCompilationTask] with sources found in the generatedSources directory. This should be run after
  * annotation processors have been run.
  */
-internal fun JvmCompilationTask.expandWithGeneratedSources(): JvmCompilationTask =
```
Bencodes, line 475: Internal will be visible to tests if you pass this target to the test target as an associate dependency.

src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutorTest.kt
```diff
@@ -44,11 +44,11 @@ class KotlinJvmTaskExecutorTest {
 
     assertTrue(expandedCompileTask.hasInputs())
     assertNotNull(expandedCompileTask.inputs.javaSourcesList.find { path ->
-      path.endsWith("a_test_1/generated_sources/AnotherGenClass.java")
+      path.endsWith("generated_sources/AnotherGenClass.java")
```
raghulvelusamy, line 47: This always generates not `1` but a random number `a-test-115507972808607999568`. Removing the prefix here.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
raghulvelusamy: going to make it public to avoid using 4f0b72806f3603b977138116549762c119ef632c

Bencodes: Forgot about the `kt_bootstrap` layer that these sources are wrapped in.

Bencodes (approved):

PR #1101, opened 2024-01-09
Update name of experimental_prune_transitive_deps flag.

plobsing:
After the rename to `rules_kotlin`, the flag
`--@io_bazel_rules_kotlin//kotlin/settings:experimental_prune_transitive_deps=True` no longer exists.


Bencodes (approved):
Thanks for catching this and fixing it!

PR #1105, opened 2024-01-22
[kt_compiler_plugin] Add kt_plugin_cfg for reusable and composable options

restingbull:
* This allows plugins like KSP and KAPT to be composed with different configuration sets. Will be addressed in followup prs.
* Has starlark tests! (It's almost not painful)

Require bootstrap, will focus on fixing that next.


examples/ksp/BUILD
```diff
@@ -81,3 +81,73 @@ build_test(
         "//:coffee_app_deploy.jar",
     ],
 )
+
+kt_compiler_plugin(
+    name = "ksp",
```
Bencodes, line 86: This seems like something to burry inside of `rules_kotlin`, otherwise people might assume it's how they are expected to do things.

Bencodes (approved):

examples/ksp/BUILD
bqv: Is this the only way to provide options for ksp plugins, for example `room.generateKotlin`?

PR #1110, opened 2024-02-13
Send `**kwargs` to base `_android_library`

pswaminathan:
Prior to #842, kwargs were sent to the base android library rule. In that PR, one of the branches no longer contained it. This causes any target with resources/manifest not to send kwargs, which means that e.g. `assets` don't get passed down.

### DIFF
--- a/kotlin/internal/jvm/android.bzl
+++ b/kotlin/internal/jvm/android.bzl
@@ -67,6 +67,7 @@ def _kt_android_artifact(
         # TODO(https://github.com/bazelbuild/rules_kotlin/issues/556): replace with starlark
         # buildifier: disable=native-android
         # Do not export deps to avoid all upstream targets to be invalidated when ABI changes.
+        kwargs.pop("exports", None)
         _android_library(
             name = base_name,
             resource_files = resource_files,
--- a/kotlin/internal/jvm/android.bzl
+++ b/kotlin/internal/jvm/android.bzl
@@ -63,7 +63,7 @@ def _kt_android_artifact(
             **kwargs
         )
         exported_target_labels.append(base_name)
-    elif resource_files or manifest:
+    elif resource_files or manifest or assets:
         # TODO(https://github.com/bazelbuild/rules_kotlin/issues/556): replace with starlark
         # buildifier: disable=native-android
         # Do not export deps to avoid all upstream targets to be invalidated when ABI changes.
--- a/kotlin/internal/jvm/android.bzl
+++ b/kotlin/internal/jvm/android.bzl
@@ -49,6 +49,7 @@ def _kt_android_artifact(
 
     exported_target_labels = [kt_name]
     manifest = kwargs.get("manifest", default = None)
+    assets = kwargs.get("assets", default = [])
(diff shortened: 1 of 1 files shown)


kotlin/internal/jvm/android.bzl
```diff
@@ -76,6 +76,7 @@ def _kt_android_artifact(
             enable_data_binding = enable_data_binding,
             tags = tags,
             visibility = ["//visibility:private"],
+            **kwargs
```
Bencodes, line 79: @oliviernotteghem is there a reason you opted to not pass the full kwargs for this case?

Bencodes: @pswaminathan I think you also want to check for assets in the conditional that gates this logic.

```
if resource_files or manifest or assets:
```

kotlin/internal/jvm/android.bzl
```diff
@@ -63,19 +64,20 @@ def _kt_android_artifact(
...
         # TODO(https://github.com/bazelbuild/rules_kotlin/issues/556): replace with starlark
         # buildifier: disable=native-android
         # Do not export deps to avoid all upstream targets to be invalidated when ABI changes.
+        kwargs.pop("exports", default = None)
         _android_library(
             name = base_name,
             resource_files = resource_files,
             deps = deps,
             custom_package = kwargs.get("custom_package", default = None),
-            manifest = manifest,
```
pswaminathan, line 75: This removal is here because, if we are passing `kwargs`, `manifest` will still be in the dict (by virtue of using `.get`), and so will be passed twice.

kotlin/internal/jvm/android.bzl
```diff
@@ -63,19 +64,19 @@ def _kt_android_artifact(
...
         _android_library(
             name = base_name,
             resource_files = resource_files,
             deps = deps,
-            custom_package = kwargs.get("custom_package", default = None),
-            manifest = manifest,
             enable_data_binding = enable_data_binding,
             tags = tags,
             visibility = ["//visibility:private"],
+            **kwargs
```
Bencodes, line 79: I'd still just pass the assets here instead of the entire `kwargs`.

pswaminathan: It's not just assets though. If assets is provided, assets_dir must be provided. Providing one and not the other causes an error in android_library. This is a prime example of why we shouldn't be white-listing which arguments to pass down. 

As another example, we have libraries that use manifest_values and some using proguard_specs. Checking for and whitelisting individual attributes is not a reliable path forward.

oliviernotteghem: @Bencodes : yes, this is to avoid exports which causes extra invalidation / increased build time. But we ran into 2 issues since then 1) assets are not handled indeed 2) resource-less module don't have their manifest merged upstream. I can share the update we've made to fix this if this helps (and we didn't get any issues since then).

oliviernotteghem: here we go https://github.com/bazelbuild/rules_kotlin/pull/1113

Bencodes:

PR #1116, opened 2024-02-23
Refactor KotlinBuilderJvmJdepsTest to do exact dep assertions

Bencodes:
Moving the assertions over so that we are comparing the exact expected jdeps to the actual jdeps.

This avoids us making changes that accidentally leak types between the different dependency types.

### DIFF
--- a/src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
+++ b/src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
@@ -38,7 +38,7 @@ class KotlinBuilderJvmJdepsTest(private val enableK2Compiler: Boolean) {
       return listOf(
         // TODO: Enable k2 for tests
 //        arrayOf(true),
-        arrayOf(false)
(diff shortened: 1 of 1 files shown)


src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
```diff
@@ -38,71 +38,79 @@ class KotlinBuilderJvmJdepsTest(private val enableK2Compiler: Boolean) {
...
       )
       c.addDirectDependencies(TEST_FIXTURES_DEP)
     }
-    val jdeps = depsProto(dependingTarget)
 
-    assertExplicit(jdeps).containsExactly(TEST_FIXTURES_DEP.singleCompileJar())
-    assertImplicit(jdeps).isEmpty()
-    assertUnused(jdeps).isEmpty()
-    assertIncomplete(jdeps).isEmpty()
+    val jdeps = depsProto(dependingTarget)
```
Bencodes, line 101: To de-risk this PR I left the original assertions in place until I finished making all of the new assertions pass.

PR #1115, opened 2024-02-22
Improve jdeps tracking and add test cases

Bencodes:

PR #1118, opened 2024-02-26
Improve jdeps logic to find more explicit and implicit dependencies

scosenza:
# Problem
Kotlin jdeps do not collect many direct and indirect dependencies that are required for compilation. These missing jdep found dependencies were discovered as Airbnb tests a "jdeps compile-time pruning" optimization which provides significant IntelliJ indexing, UI, and compilation performance benefits in our fork of the [Bazel IntelliJ Plugin](https://github.com/bazelbuild/intellij). See also https://github.com/bazelbuild/rules_kotlin/pull/1115 as @Bencodes and I unknowingly started looking at this issue at about the same time!

# Solution
As a followup to https://github.com/bazelbuild/rules_kotlin/pull/1045, write test cases to repro missing deps found when compiling Airbnb's JVM monorepo, and then fix the underlying issues. Note that these fixes were made with *very minimal knowledge* of Kotlin compiler internals, so please suggest alternatives if there are better ways to collect these types. Thanks!

# Test Plan:
* Wrote Kotlin and Java test cases to repro missing dependencies found in the Airbnb monorepo
* Updated Airbnb's minimal rules_kotlin fork to use the changes in this PR and the updated rules_kotlin fork has been successfully running in CI for over a week.

# Appendix
For anyone interested in how we're planning to use jdeps pruning:
* We've been experimenting with using IntelliJ's incremental compiler to perform compilations after the initial Bazel sync. 
* We use jdeps to prune the compile classpath, so that only the interface jars needed to compile the editable targets are added as IntelliJ libraries. For one of our large services, we see a 20x reduction in compile time jars (2000 -> 100). This reduction speeds up indexing, UI responsiveness, and incremental compilation speed (5s -> 1s).
* Using IntelliJ's incremental compiler with a jdeps pruned classpath has been 30-70x faster than "bazel build/test" for the same large service (30-70s -> 1s) on an M2 Mac. Note that we follow "1:1:1" and have a BUILD file in every Java/Kotlin/Scala package.


src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
```diff
@@ -46,6 +46,7 @@ class KotlinBuilderJvmJdepsTest(private val enableK2Compiler: Boolean) {
   val ctx = KotlinJvmTestBuilder()
 
   val TEST_FIXTURES_DEP = Dep.fromLabel(":JdepsParserTestFixtures")
+  val TEST_FIXTURES2_DEP = Dep.fromLabel(":JdepsParserTestFixtures2")
```
scosenza, line 49: To enable test cases where Java deps depend on other Java deps I added a second set of java test fixtures here

src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
```diff
@@ -1385,7 +1621,7 @@ class KotlinBuilderJvmJdepsTest(private val enableK2Compiler: Boolean) {
 
     assertExplicit(jdeps).contains(depWithFunction.singleCompileJar())
     assertImplicit(jdeps).contains(depWithReturnType.singleCompileJar())
-    assertImplicit(jdeps).doesNotContain(depWithReturnTypesSuperType)
+    assertImplicit(jdeps).contains(depWithReturnTypesSuperType.singleCompileJar())
```
scosenza, line 1624: Note that the prior assertion always succeeded as it was missing `.singleCompileJar()`
When .singleCompileJar() is correctly added, the assertion fails. 
We can confirm that depWithReturnTypesSuperType is needed for compilation by commenting out:
`c.addTransitiveDependencies(depWithReturnTypesSuperType)`
and then seeing the compilation fail with:
```
```
Because the super type is actually needed, we've changed the assertion to assert that it is an implicit dep.

src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
```diff
@@ -355,6 +407,46 @@ class KotlinBuilderJvmJdepsTest(private val enableK2Compiler: Boolean) {
...
+        "ConnectionNotFoundException.kt",
+        """
+          package something2
+          
+          class ConnectionNotFoundException : Exception()
+        """
+      )
+    }
+
+    println("=====================================")
```
Bencodes, line 423: Did you mean to check this in?

src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
```diff
@@ -1659,7 +1896,34 @@ class KotlinBuilderJvmJdepsTest(private val enableK2Compiler: Boolean) {
 
     assertExplicit(jdeps).containsAtLeast(depWithFunction.singleCompileJar(), depWithReturnType.singleCompileJar())
     assertImplicit(jdeps).doesNotContain(depWithReturnType.singleCompileJar())
-    assertImplicit(jdeps).doesNotContain(depWithReturnTypesSuperType)
+    assertImplicit(jdeps).contains(depWithReturnTypesSuperType.singleCompileJar())
```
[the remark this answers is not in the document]
Bencodes: I have a PR up that does full `Deps` diffing which should hopefully avoid issues like this where assertions aren't being correctly done.

https://github.com/bazelbuild/rules_kotlin/pull/1116/files

src/test/kotlin/io/bazel/kotlin/builder/tasks/BUILD.bazel
```diff
@@ -19,6 +19,15 @@ java_library(
...
     srcs = glob(["testFixtures/*.java"]),
+    deps = [
+        ":JdepsParserTestFixtures2",
+    ],
+)
+
+java_library(
+    name = "JdepsParserTestFixtures2",
+    testonly = True,
+    srcs = glob(["testFixtures2/*.java"]),
```
Bencodes, line 30: Might look less copy-paste grouping these different test figures inside directories within the `testFixtures` directory instead of just duplicating and adding a numerical index. Ex: `testFixtures/a-sources/` `testFixtures/b-sources/`

src/test/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinBuilderJvmJdepsTest.kt
scosenza: Thanks for flagging. Will remove.

Bencodes:
@scosenza are you able to rebase this branch and resolve the conflicts + organize the test fixtures?

scosenza:
FYI, it looks like WorkerEnvironmentTest may be flakey, as it's passing on my local Mac machine, and passing in the Linux CI build.

scosenza:
> @scosenza are you able to rebase this branch and resolve the conflicts + organize the test fixtures?

@Bencodes I just updated my branch with the latest upstream changes. And thanks again for greatly improving the assertions!

src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt
```diff
@@ -8,39 +8,46 @@ import org.jetbrains.kotlin.analyzer.AnalysisResult
...
 import org.jetbrains.kotlin.descriptors.ModuleDescriptor
-import org.jetbrains.kotlin.descriptors.ParameterDescriptor
 import org.jetbrains.kotlin.descriptors.PropertyDescriptor
 import org.jetbrains.kotlin.descriptors.SourceElement
+import org.jetbrains.kotlin.descriptors.ValueDescriptor
 import org.jetbrains.kotlin.descriptors.impl.LocalVariableDescriptor
 import org.jetbrains.kotlin.extensions.StorageComponentContainerContributor
-import org.jetbrains.kotlin.load.java.descriptors.JavaMethodDescriptor
-import org.jetbrains.kotlin.load.java.descriptors.JavaPropertyDescriptor
+import org.jetbrains.kotlin.js.translate.callTranslator.getReturnType
```
Bencodes, line 22: Is this the right function to be using? It looks like it's being loaded from the javascript package.

scosenza: I remember thinking the same thing when I imported it, but it was needed to access a type which was required to get one of my new tests to pass. Can someone point me to a more appropriate alternative?

Bencodes: I briefly looked around and didn't see anything that stands out. If anything you can just inline the function itself, which is just calling into another function with `resolvedCall` as an argument.

scosenza:
Will def inline then! I recall at the time thinking I needed that method to
access something internal, but I must have confused it with something else.

src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt
scosenza: Thanks for flagging this and taking another look. I just got in front of my computer, and the great news is that I was able to remove the code using that javascript class, and all my tests still pass. I was trying to page back in what I was thinking at the time, and the best I can come up with is that `getReturnType` always returned a KotlinType, while the alternative below that optionally returns a KotlinType. To confirm the new code is functionally equivalent, 
I'll also publish a new internal version of `rules_kotlin` to test in our monorepo, and report back if everything looks good. Thanks again!

Bencodes (approved):

kolloch:
@scosenza the uses of the jdeps pruning sound really exciting. Any chance that people outside of airbnb could use a similar setup or does it all depend on your bazel intellij plugin fork?

PR #1128, opened 2024-03-10
[testing] Switch to junit5

restingbull:
Fixes tests for modern jvms.


kotlin/internal/jvm/jvm.bzl
```diff
@@ -344,6 +344,10 @@ Setup a simple kotlin_test.
             default = Label("@bazel_tools//tools/jdk:TestRunner_deploy.jar"),
             allow_files = True,
         ),
+        "use_testrunner": attr.bool(
```
Bencodes, line 347: What exactly is this flag solving for and when would you use it externally vs. setting it to `False`?

src/main/starlark/core/repositories/setup.bzl
```diff
@@ -56,11 +56,17 @@ def kt_configure():
...
+            "org.junit.jupiter:junit-jupiter-api:5.4.2",
+            "org.junit.jupiter:junit-jupiter-params:5.4.2",
+            "org.junit.jupiter:junit-jupiter-engine:5.4.2",
+            "org.junit.platform:junit-platform-console:1.4.2",
         ],
         repositories = [
             "https://maven-central.storage.googleapis.com/repos/central/data/",
             "https://maven.google.com",
             "https://repo1.maven.org/maven2",
+            "https://jcenter.bintray.com",
```
Bencodes, line 69: I believe jcenter is being shut down, best to not depend on it.

kotlin/internal/jvm/jvm.bzl
restingbull: It matches [java_test.use_testrunner](https://bazel.build/reference/be/java#java_test.use_testrunner). It allows customizing the test execution environment. In this case, we replace the bazel test runner with JUnit5.

src/main/starlark/core/repositories/setup.bzl
restingbull: Removed.

Bencodes (approved):

Issue #924, opened 2023-01-25
Kotlin rules don't work with `bazel --nolegacy_external_runfiles`


srabraham:
There are performance benefits to using `--nolegacy_external_runfiles` ([code here](https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java;l=452-460;drc=3a8da9268ffa9c814de4ce4bc1b0e349b37072a5)) as a Bazel build option, [with some additional context here](https://github.com/bazelbuild/bazel/issues/9306#issuecomment-1370440301). As @alexeagle tells me, it allows for an optimization in runfiles layout.

As things are, this setting appears to be incompatible with rules_kotlin. When I try to build a `kt_jvm_library` with `bazel build --nolegacy_external_runfiles`, I get an error like the following:

<details>
<summary>Click to expand logs</summary>

```
$ bazel run --nolegacy_external_runfiles src/main/kotlin/experimental/users/sabraham:thing
INFO: Analyzed target //src/main/kotlin/experimental/users/sabraham:thing (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
ERROR: /Users/sean/dev/central-ult/src/main/kotlin/experimental/users/sabraham/BUILD.bazel:15:14: KotlinCompile @//src/main/kotlin/experimental/users/sabraham:thing { kt: 1, java: 0, srcjars: 0 } for darwin_arm64 failed: Worker process did not return a WorkResponse:

---8<---8<--- Start of log, file at /private/var/tmp/_bazel_sean/3e0bb96a7b651dc41ed28d370dc8ba44/bazel-workers/worker-4-KotlinCompile.log ---8<---8<---
Exception in thread "main" java.lang.ExceptionInInitializerError
	at io.bazel.kotlin.builder.cmd.Build$main$1.invoke(Build.kt:33)
	at io.bazel.kotlin.builder.cmd.Build$main$1.invoke(Build.kt:29)
	at io.bazel.worker.Worker$Companion.from(Worker.kt:31)
	at io.bazel.kotlin.builder.cmd.Build.main(Build.kt:29)
[...]

Issue #1061, opened 2023-10-27
java not found with bzlmod and `--nolegacy_external_runfiles`


agluszak:
Repro: https://github.com/agluszak/bzlmod-playground/tree/rules_kotlin_legacy_bzlmod

```
andrzej@MUNIT-521 ~/c/b/bzlmod-playground (rules_kotlin_legacy_bzlmod)> bazel test //...
INFO: Analyzed 2 targets (27 packages loaded, 349 targets configured).
INFO: Found 1 target and 1 test target...
FAIL: //:test (see /home/andrzej/.cache/bazel/_bazel_andrzej/554d6595289376b6e5422fdcf90c9ed5/execroot/_main/bazel-out/k8-fastbuild/testlogs/test/test.log)
INFO: Elapsed time: 0.357s, Critical Path: 0.05s
INFO: 2 processes: 2 linux-sandbox.
INFO: Build completed, 1 test FAILED, 2 total actions
//:test                                                                  FAILED in 0.0s
  /home/andrzej/.cache/bazel/_bazel_andrzej/554d6595289376b6e5422fdcf90c9ed5/execroot/_main/bazel-out/k8-fastbuild/testlogs/test/test.log

Executed 1 out of 1 test: 1 fails locally.
andrzej@MUNIT-521 ~/c/b/bzlmod-playground (rules_kotlin_legacy_bzlmod) [3]> cat /home/andrzej/.cache/bazel/_bazel_andrzej/554d6595289376b6e5422fdcf90c9ed5/execroot/_main/bazel-out/k8-fastbuild/testlogs/test/test.log
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //:test
-----------------------------------------------------------------------------
/home/andrzej/.cache/bazel/_bazel_andrzej/554d6595289376b6e5422fdcf90c9ed5/sandbox/linux-sandbox/18/execroot/_main/bazel-out/k8-fastbuild/bin/test.runfiles/_main/test: line 396: /home/andrzej/.cache/bazel/_bazel_andrzej/554d6595289376b6e5422fdcf90c9ed5/sandbox/linux-sandbox/18/execroot/_main/bazel-out/k8-fastbuild/bin/test.runfiles/_main/external/rules_java~6.4.0~toolchains~local_jdk/bin/java: No such file or directory
```

agluszak:
Possibly related: #924

drmercer-lucid:
[...]

Issue #778, opened 2022-06-18
Update Legacy Runfiles


aryeh-looker:
##Repro:

1. Apply this diff:

```
diff --git a/.bazelrc b/.bazelrc
index 6858b64..d7db834 100644
--- a/.bazelrc
+++ b/.bazelrc
@@ -2,3 +2,4 @@ build --strategy=KotlinCompile=worker
 build --test_output=all
 build --verbose_failures
 
+build --nolegacy_external_runfiles
diff --git a/examples/trivial/app/BUILD.bazel b/examples/trivial/app/BUILD.bazel
index 3ea87a0..5e67876 100644
--- a/examples/trivial/app/BUILD.bazel
+++ b/examples/trivial/app/BUILD.bazel
@@ -1,5 +1,5 @@
 load("@rules_java//java:defs.bzl", "java_binary")
-load("@io_bazel_rules_kotlin//kotlin:jvm.bzl", "kt_jvm_library")
+load("@io_bazel_rules_kotlin//kotlin:jvm.bzl", "kt_jvm_library", "kt_jvm_binary")
 load("@io_bazel_rules_kotlin//kotlin:lint.bzl", "ktlint_fix", "ktlint_test")
 
 kt_jvm_library(
@@ -25,3 +25,9 @@ java_binary(
     main_class = "app.MyApp",
     runtime_deps = [":app_lib"],
 )
+
+kt_jvm_binary(
+    name = "myappkt",
+    main_class = "app.MyApp",
+    runtime_deps = [":app_lib"]
+)
\ No newline at end of file
```

2. Run `cd examples/trivial; bazel run //app:myappkt`

##Observed behavior: 

Task fails to run:

```
bazel run //app:myappkt
DEBUG: Rule 'io_bazel_rules_kotlin' indicated that a canonical reproducible form can be obtained by modifying arguments release_archive = "/usr/local/google/home/aryehh/.cache/bazel/_bazel_aryehh/0cd998f3b47b09269faeff3bb012228d/external/io_bazel_rules_kotlin_head" and dropping ["name", "local_release_archive_target"]
DEBUG: Repository io_bazel_rules_kotlin instantiated at:
  /usr/local/google/home/aryehh/Development/rules_kotlin/examples/trivial/WORKSPACE:10:19: in <toplevel>
[...]

PR #1130, opened 2024-03-13
fix running kotlin targets (at least under bzlmod?)

dmgd:
I was just integrating rules_kotlin into an existing bazel project and run into some issues getting a simple kotlin app up and running. I was seeing the error 

```
...<app>.runfiles/_main/external/rules_java~~toolchains~remotejdk21_macos_aarch64/bin/java: No such file or directory
```

I explored a little and saw that these two fields in the java toolchain differed and explained the problem (since _exec_ is what's being used, but _runfiles_ is what points to the actual file.)

```
java_executable_exec_path = external/rules_java~~toolchains~remotejdk21_macos_aarch64/bin/java
java_executable_runfiles_path = ../rules_java~~toolchains~remotejdk21_macos_aarch64/bin/java
```

Looking at the docs for these two fields (https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/starlarkbuildapi/java/JavaRuntimeInfoApi.java#L60 ) it looks to me like exec is for use while running java as _part_ of the action, whereas runfiles is for use when using `bazel run` (but this is all a bit new to me so I might be wrong / perhaps there's some nuance -- e.g. I can see in (what looks to me like) the core java stuff, there's all sorts of complex stuff going on when they set javabin . . . -- https://github.com/bazelbuild/bazel/blob/master/src/main/starlark/builtins_bzl/bazel/java/bazel_java_binary.bzl#L203 )

env details
- macos
- bazel version: 7.1.0
- rules_kotlin version: 1.9.1

[edit] forgot to mention, that I'm using this as a patch and `bazel run <kotlin app target>` is working fine now

### DIFF
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -71,7 +71,7 @@ def _write_launcher_action(ctx, rjars, main_class, jvm_flags):
     """
     jvm_flags = " ".join([ctx.expand_location(f, ctx.attr.data) for f in jvm_flags])
     template = ctx.attr._java_stub_template.files.to_list()[0]
-    java_bin_path = ctx.attr._java_runtime[java_common.JavaRuntimeInfo].java_executable_exec_path
+    java_bin_path = ctx.attr._java_runtime[java_common.JavaRuntimeInfo].java_executable_runfiles_path
 
     if ctx.configuration.coverage_enabled:
         jacocorunner = ctx.toolchains[_TOOLCHAIN_TYPE].jacocorunner


drmercer-lucid:
I believe this would fix #924 and #778 and #1061. What would it take to land this PR?

(I was about to open a PR to fix this but then I saw this PR :smile: )

Bencodes:
This change looks good to me, I might re-open this PR to resolve the conflicts and test against HEAD.

Bencodes:
Merged this https://github.com/bazelbuild/rules_kotlin/pull/1242

Issue #1129, opened 2024-03-11
[Bazel CI] Test logs failed for kotlin-builder at Bazel@HEAD in CI



sgowroji:
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3711#018e2bad-e46a-411e-9bc7-800fb1641710

Platform: Ubuntu

Logs:

```
//src/test/kotlin/io/bazel/kotlin:KotlinJvm13Test
//src/test/kotlin/io/bazel/kotlin:KotlinJvmAssociatesBasicVisibilityTest
//src/test/kotlin/io/bazel/kotlin:KotlinJvmBasicAssertionTest      
//src/test/kotlin/io/bazel/kotlin:KotlinJvmKaptAssertionTest 
...
```

Culprit:

Steps:

```
 git clone https://github.com/bazelbuild/rules_kotlin
git reset e5283665675c94e539d1aa00e9582928d3b25a2e  --hard
export USE_BAZEL_VERSION=3697397569e25a29a828fb46b26302876d372e26
bazel test //:all_tests 
```

CC Greenteam @Wyverald

Wyverald:
This seems to be due to the JDK updates on CI machines. @meteorcloudy do we need to do anything here?

meteorcloudy:
@Bencodes This probably need a similar fix in https://github.com/bazelbuild/rules_kotlin/blob/b9380cd1ce3a2c7ae337531c455c425fac38302c/src/test/kotlin/io/bazel/kotlin/defs.bzl#L26 as https://github.com/bazelbuild/bazel/commit/6783339e3fd21655b7fa3be2daedde4491ae0348 to add `-Djava.security.manager=allow` for the Java test runner.

sgowroji:
CI is passing now and shows Green. We can close this issue.

PR #1132, opened 2024-03-14
Pass -Djava.security.manager=allow in kt_rules_test

Bencodes:
https://github.com/bazelbuild/rules_kotlin/issues/1129

### DIFF
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -1,4 +1,4 @@
-load("@bazel_tools//tools/jdk:toolchain_utils.bzl", "find_java_toolchain")
+load("@bazel_tools//tools/jdk:toolchain_utils.bzl", "find_java_runtime_toolchain")
 load("@rules_java//java:defs.bzl", "JavaInfo", "JavaPluginInfo", "java_common")
 load(
     "//kotlin/internal:defs.bzl",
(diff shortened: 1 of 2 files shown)


restingbull:
I tried this:
```

Error occurred during initialization of VM
--
  | java.lang.Error: Could not create SecurityManager
  | at java.lang.System.initPhase3(java.base@11.0.19/System.java:2065)
  | Caused by: java.lang.ClassNotFoundException: allow
  | at jdk.internal.loader.BuiltinClassLoader.loadClass(java.base@11.0.19/BuiltinClassLoader.java:581)
  | at jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(java.base@11.0.19/ClassLoaders.java:178)
  | at java.lang.ClassLoader.loadClass(java.base@11.0.19/ClassLoader.java:522)
  | at java.lang.Class.forName0(java.base@11.0.19/Native Method)
  | at java.lang.Class.forName(java.base@11.0.19/Class.java:398)
  | at java.lang.System.initPhase3(java.base@11.0.19/System.java:2050)


```

There is some more subtlety here, haven't had a chance to dig in.

Bencodes:
@restingbull we can also pass this at the bazelrc level which I believe solves the issue, at least locally.

restingbull:
> @restingbull we can also pass this at the bazelrc level which I believe solves the issue, at least locally.

Not as far as I can tell? The flag doesn't seem to do what we want it to in the later jvm versions -- that message is from the buildkite pass on this pr. The problem is that `allow` is not valid pre-18, according to the associated [jep](https://openjdk.org/jeps/411). Instead, it's considered to be an alternative security manager... I'm not clear on how we can sniff the java verio

e.g. this fails on 11, but passes on 18+. Until such time as it doesn't exist. This is why I moved us to junit5, ~and am in consideration how to fix the prod issue. The deprecation is gonna get us, one way of another.~

Nevermind, figured it out. Have a patch momentarily.

kotlin/internal/jvm/impl.bzl
```diff
@@ -289,6 +290,12 @@ def kt_jvm_junit_test_impl(ctx):
     jvm_flags = []
     if hasattr(ctx.fragments.java, "default_jvm_opts"):
         jvm_flags = ctx.fragments.java.default_jvm_opts
+
+    java_toolchain = find_java_toolchain(ctx, ctx.attr._java_toolchain)
+
+    if java_toolchain and java_toolchain.java_runtime and java_toolchain.java_runtime.version > 18:
```
Bencodes, line 296: I'd copy what Bazel does exactly so that we have some consistency. Can you add a comment as well linking out to their implementation for reference? https://github.com/bazelbuild/bazel/blob/6d5b084025a26f2f6d5041f7a9e8d302c590bc80/src/main/starlark/builtins_bzl/bazel/java/bazel_java_binary.bzl#L66-L67

Bencodes:
👍

PR #1134, opened 2024-03-15
Use maven lock-file

scaiper:
This PR enables use of a lock-file for `kotlin_rules_maven` repo. This in turn enables caching and reproducible builds.
Had to bump `rules_jvm_external` to 6.0 and some of its transitive deps.


scaiper:
Not sure where this error comes from in the failed check. Seems unrelated.
```
ERROR: https://bcr.bazel.build/modules/rules_jvm_external/6.0/MODULE.bazel:65:13: name 'use_repo_rule' is not defined
```

MODULE.bazel
```diff
@@ -5,11 +5,11 @@ module(
     repo_name = "rules_kotlin",
 )
 
-bazel_dep(name = "platforms", version = "0.0.6")
+bazel_dep(name = "platforms", version = "0.0.8")
```
Bencodes, line 8: Can you also update the WORKSPACE versions of these so that they are kept somewhat in sync with each other?

scaiper: Sure. But I haven't found where `platforms` and `rules_cc` specified in WORKSPACE version.

scaiper: Updated PR

Bencodes:
> Not sure where this error comes from in the failed check. Seems unrelated.
> 
> ```
> ERROR: https://bcr.bazel.build/modules/rules_jvm_external/6.0/MODULE.bazel:65:13: name 'use_repo_rule' is not defined
> ```

You might need to downgrade rules_jvm_external and file an issue upstream for this one.

MODULE.bazel
Bencodes: Can you mirror these changes in `MODULE.release.bazel` so that the release archive also has these updates?

scaiper: Already did. Did I miss something?

scaiper:
> > Not sure where this error comes from in the failed check. Seems unrelated.
> > ```
> > ERROR: https://bcr.bazel.build/modules/rules_jvm_external/6.0/MODULE.bazel:65:13: name 'use_repo_rule' is not defined
> > ```
> 
> You might need to downgrade rules_jvm_external and file an issue upstream for this one.

I'm not sure what to report. I guess `use_repo_rule` was not available in bazel 6.4.0, which is used in this one test.

MODULE.bazel
Bencodes: Nope, I just skimmed over it. Looks good to me.

scaiper:
Indeed, `use_repo_rule` is a 7.x feature. So questing is, is support of 6.x bazel important or it could be dropped?
If 6.x support is desirable, I quess I should try to make it work without bumping the versions.

scaiper:
Version 7.0.2 is fine. Problem is with this test 
```Example - Android using Bazel 6.x on :ubuntu: Ubuntu 18.04 LTS (OpenJDK 11, gcc 7.5.0)```
```
:docker: Pulling gcr.io/bazel-public/ubuntu1804-java11
:docker: Running command in gcr.io/bazel-public/ubuntu1804-java11
:bazel: Using Bazel version 6.4.0
bazel info output_base
2024/03/15 18:12:57 Downloading https://releases.bazel.build/6.4.0/release/bazel-6.4.0-linux-x86_64...
Extracting Bazel installation...
Starting local Bazel server and connecting to it...
ℹ️ Bazel Info
```

Bencodes:
> So questing is, is support of 6.x bazel important or it could be dropped?

At the moment we try to keep compatibility with 5.x, 6.x, and 7.x https://github.com/bazelbuild/rules_kotlin?tab=readme-ov-file#rules_kotlin-bazel-compatibility

Something that's odd is this example test shouldn't even be using Bzlmod https://github.com/bazelbuild/rules_kotlin/blob/be6fb59f169ac6e1dcbf395c325462343a68214b/.bazelci/presubmit.yml#L59C3-L67

scaiper:
Shouldn't it also be specified in the `build_flags`?

scaiper:
It's `bazel info` command that failed, apparently it also needs the `--enable_bzlmod=false` flag.

I've managed to make it work using current versions of dependencies, so I've reverted version bump.

Bencodes (approved):

Issue #1138, opened 2024-03-18
Issue with KSP picking up Kotlin Compiler classpath dependencies


EdbertChan:
We have a project where Kotlin and a Motif (KSP plugin) is being mixed in the same module. 

We are finding that Dagger will run and pick up the Guava classes that are packed in with the KotlinCompiler.jar that is supplied by Jetbrains.

I am working on a solid repro case but I am opening this issue in case anyone else has seen it.

Example:

Consider the below stack trace. I can see that when going through the hierarchy of the object being tested, they are loaded from places that are not supplied by the codebase or by Dagger. It looks like rules_kotlin is hijacking the classpath.

jar:file:/Users/edbert/.bazelcache/workspace/ff080e9ef5254374b7ce1c9f144b976d/execroot/__main__/bazel-out/xplat-opt-exec-ST-4ec9c722f6b9/bin/external/io_bazel_rules_kotlin/src/main/kotlin/build.runfiles/io_bazel_rules_kotlin/src/main/kotlin/jdeps-gen.jar!/com/google/common/graph/AbstractNetwork$1.class

jar:file:/Users/edbert/.bazelcache/workspace/ff080e9ef5254374b7ce1c9f144b976d/execroot/__main__/bazel-out/xplat-opt-exec-ST-4ec9c722f6b9/bin/external/io_bazel_rules_kotlin/src/main/kotlin/build.runfiles/io_bazel_rules_kotlin/src/main/kotlin/jdeps-gen.jar!/com/google/common/graph/AbstractBaseGraph.class

jar:file:/Users/edbert/.bazelcache/workspace/ff080e9ef5254374b7ce1c9f144b976d/execroot/__main__/bazel-out/xplat-opt-exec-ST-4ec9c722f6b9/bin/external/io_bazel_rules_kotlin/src/main/kotlin/build.runfiles/com_github_jetbrains_kotlin/lib/kotlin-compiler.jar!/com/google/common/graph/BaseGraph.class

Stack trace

```
error: [ksp] java.lang.ClassCastException: class com.google.common.graph.AbstractNetwork$1 cannot be cast to class com.google.common.graph.SuccessorsFunction (com.google.common.graph.AbstractNetwork$1 and com.google.common.graph.SuccessorsFunction are in unnamed module of loader org.jetbrains.kotlin.preloading.MemoryBasedClassLoader @370736d9)
        at com.google.common.graph.Traverser.<init>(Traverser.java:72)
[...]

Issue #1074, opened 2023-11-16
Support generating Java code with KSP


lukaciko:
As noted in the README, `kt_ksp_plugin` doesn't support generating Java code. Some annotation processors, like Dagger 2.48+, generate Java code and support KSP.

---

When implemented, updating _examples/dagger_ to use KSP and building it should work:

```diff
diff --git a/examples/dagger/WORKSPACE b/examples/dagger/WORKSPACE
index 213945f..d74fe47 100644
--- a/examples/dagger/WORKSPACE
+++ b/examples/dagger/WORKSPACE
@@ -31,9 +31,9 @@ load("@rules_jvm_external//:defs.bzl", "maven_install")

 maven_install(
     artifacts = [
-        "com.google.dagger:dagger:2.45",
-        "com.google.dagger:dagger-compiler:2.45",
-        "com.google.dagger:dagger-producers:2.45",
+        "com.google.dagger:dagger:2.48",
+        "com.google.dagger:dagger-compiler:2.48",
+        "com.google.dagger:dagger-producers:2.48",
         "javax.inject:javax.inject:1",
         "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.4.2",
     ],
diff --git a/examples/dagger/third_party/BUILD.bazel b/examples/dagger/third_party/BUILD.bazel
index e970b92..53ceec0 100644
--- a/examples/dagger/third_party/BUILD.bazel
+++ b/examples/dagger/third_party/BUILD.bazel
@@ -1,8 +1,8 @@
-load("@rules_java//java:defs.bzl", "java_library", "java_plugin")
+load("@rules_kotlin//kotlin:core.bzl", "kt_ksp_plugin")
+load("@rules_kotlin//kotlin:jvm.bzl", "kt_jvm_library")

-java_plugin(
+kt_ksp_plugin(
     name = "dagger_component_plugin",
-    generates_api = True,
     processor_class = "dagger.internal.codegen.ComponentProcessor",
     visibility = ["//visibility:private"],
     deps = [
@@ -10,9 +10,9 @@ java_plugin(
     ],
 )

-java_library(
+kt_jvm_library(
     name = "dagger",
-    exported_plugins = [":dagger_component_plugin"],
+    plugins = [":dagger_component_plugin"],
     visibility = ["//visibility:public"],
     exports = [
         "@maven//:com_google_dagger_dagger",

```

```
cd examples/dagger
bazelisk build coffee_app
```

Issue #624, opened 2021-11-18
rules_kotlin should link the embeddable Kotlin compiler


mauriciogg:
There are two versions of the Kotlin compiler jar:

kotlin-compiler.jar: this is the vanilla version and is the one distributed in the Kotlin Github repo.
kotlin-compiler-embeddable .jar: this is a shaded version of kotlin-compiler.jar, which among other classes, shades the intellij classes. It can be used to to link safely against the compiler in an environment where there might be class collisions.
The rules_kotlin toolchain links against kotlin-compiler.jar: [KotlinToolchain.kt](https://sourcegraph.com/github.com/bazelbuild/rules_kotlin/-/blob/src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt?L111&subtree=true)
The Gradle Kotlin plugin links against the embeddable compiler: [kotlin-gradle-plugin](https://sourcegraph.com/github.com/JetBrains/kotlin/-/blob/libraries/tools/kotlin-gradle-plugin/build.gradle.kts?L109&subtree=true). NOTE: I'm not very familiar with Gradle code so this might not be 100% true.

Under most circumstances linking against either versions should not make any difference as long as everything is compiled against the same version. The problem arises when third_party dependencies that expect the embeddable compiler are brought into the mix. This is true for compiler plugins built using Gradle. In order to work around this issue the kt_compiler_plugin rule has an attr [target_embedded_compiler](https://sourcegraph.com/github.com/bazelbuild/rules_kotlin/-/blob/kotlin/internal/jvm/impl.bzl?L350&subtree=true) that when true simply deshades the plugin jars so that they can be used against the vanilla compiler. However this does not work as intended (e.g [Anvil](https://github.com/square/anvil)). This approach only deshades the main plugin jar, but none of its dependencies. A plugin might declare multiple jar artifacts as dependencies and some of those might need to be deshaded as well. An easy fix would be to deshade all jars in the transitive dependencies of the plugin, but this is suboptimal and error prone since we might deshade jars that dont need deshading in the fist place. Its not entirely clear how to find out which need to be deshaded and which don't. 

[...]

PR #1139, opened 2024-03-18
Add support for generated Java code with KSP

zalewskise:
### What has changed?
- Added support for Java generated code with KSP
- Renamed `deshading` to `reshading` to cover the case where plugins dependencies are clashing with kotlin compiler ones (proper solution would be: https://github.com/bazelbuild/rules_kotlin/issues/624) and those need to be shaded.
- Added guava shading rules
- Added dagger code, generated with ksp plugin, to ksp sample project

### Why it was changed?
- To be able to use Dagger’s KSP processor with `kt_ksp_plugin`

### Fixes
- https://github.com/bazelbuild/rules_kotlin/issues/1074
- https://github.com/bazelbuild/rules_kotlin/issues/1138


kotlin/internal/jvm/compile.bzl
```diff
@@ -772,7 +772,7 @@ def _run_kt_java_builder_actions(
         java_info = java_common.compile(
             ctx,
             source_files = srcs.java,
-            source_jars = generated_kapt_src_jars + srcs.src_jars,
+            source_jars = generated_kapt_src_jars + srcs.src_jars + generated_ksp_src_jars,
```
Bencodes, line 775: Doesn't this essentially mean that all targets that execute KSP will be forced to run a Java compilation action even if the underlying KSP plugin only produces Kotlin or class file outputs?

kotlin/internal/jvm/compile.bzl
```diff
@@ -760,9 +760,9 @@ def _run_kt_java_builder_actions(
         java_infos.append(kt_java_info)
 
     # Build Java
-    # If there is Java source or KAPT generated Java source compile that Java and fold it into
+    # If there is Java source or KAPT/KSP generated Java source compile that Java and fold it into
     # the final ABI jar. Otherwise just use the KT ABI jar as final ABI jar.
-    if srcs.java or generated_kapt_src_jars or srcs.src_jars:
+    if srcs.java or generated_kapt_src_jars or srcs.src_jars or generated_ksp_src_jars:
```
Bencodes, line 765: Doesn't this essentially mean that all targets that execute KSP will be forced to run a Java compilation action even if the underlying KSP plugin only produces Kotlin or class file outputs?

zalewskise: Great point, that's correct. 

I'm thinking how we could avoid that and I'm not really sure how we could get the information if the plugin is generating java code or not, without getting this provided as one of `kt_ksp_plugin` parameters. 
Could be something like
```python
kt_ksp_plugin = rule(
    attrs = {
        "generates_java": attr.bool(
                doc = """Runs Java compilation action for plugin generating Java output"""",
                default = False,
        ),
       ...
    }
    ...
}    
``` 
What do you think about this approach? Or maybe there is some detail information that I'm missing, that could be used to determine if we could enforce/avoid running Java compilation action for certain KSP plugins? 👀

zalewskise: @Bencodes updated PR with the approach mentioned above

Bencodes (approved):

PR #1145, opened 2024-03-22
[testing] Revert to JUnit4.

restingbull:
JUnit5 support is pretty nascent, and will require actual investment. TLDR; no support for it in the IDE, because it doesn't produce the xml bazel wants.

The impetus for switching to JUnit5 was the SecurityManager, which we fixed in another PR.

Back to JUnit4, until we have to upgrade.


MODULE.bazel
```diff
@@ -5,13 +5,33 @@ module(
     repo_name = "rules_kotlin",
 )
 
-bazel_dep(name = "platforms", version = "0.0.6")
-bazel_dep(name = "bazel_skylib", version = "1.4.2")
-bazel_dep(name = "rules_java", version = "7.2.0")
-bazel_dep(name = "rules_python", version = "0.23.1")
-bazel_dep(name = "rules_cc", version = "0.0.8")
+bazel_dep(
```
Bencodes, line 8: Did you mean to re-format this file again?

Bencodes (approved):

MODULE.bazel
restingbull: 🤬 

I need to teach IntelliJ this file is off limits.

PR #1148, opened 2024-03-27
Enable Bazel 5.x test suite

Bencodes:


restingbull (approved):

restingbull:
Seems like the workspace isn't quite right.

PR #1157, opened 2024-04-05
Remove legacy struct providers

comius:
Legacy struct providers have been deprecated by Bazel. Replacing them with modern providers, will make it possible to simplify and remove legacy handling from Bazel.

More info on: https://bazel.build/extending/rules#migrating_from_legacy_providers
Issue: https://github.com/bazelbuild/bazel/issues/19467

### DIFF
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -41,25 +41,22 @@ def _make_providers(ctx, providers, transitive_files = depset(order = "default")
     files = [ctx.outputs.jar]
     if providers.java.outputs.jdeps:
         files.append(providers.java.outputs.jdeps)
-    return struct(
-        kt = providers.kt,
-        providers = [
-            providers.java,
-            providers.kt,
-            providers.instrumented_files,
-            DefaultInfo(
-                files = depset(files),
-                runfiles = ctx.runfiles(
-                    # explicitly include data files, otherwise they appear to be missing
-                    files = ctx.files.data,
-                    transitive_files = transitive_files,
-                    # continue to use collect_default until proper transitive data collecting is
-                    # implmented.
-                    collect_default = True,
-                ),
+    return [
+        providers.java,
+        providers.kt,
+        providers.instrumented_files,
+        DefaultInfo(
+            files = depset(files),
+            runfiles = ctx.runfiles(
+                # explicitly include data files, otherwise they appear to be missing
+                files = ctx.files.data,
+                transitive_files = transitive_files,
+                # continue to use collect_default until proper transitive data collecting is
+                # implmented.
+                collect_default = True,
             ),
-        ] + list(additional_providers),
-    )
+        ),
+    ] + list(additional_providers)
 
 def _write_launcher_action(ctx, rjars, main_class, jvm_flags):
     """Macro that writes out a launcher script shell script.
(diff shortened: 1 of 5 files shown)


comius:
This is not yet complete cleanup, because rules_kotlin depend on an old release containing .bzl code. To do the full cleanup, this needs to be released, version repinned and then released again.

The change was tested by running tests with `--incompatible_disallow_struct_provider_syntax`

kotlin/internal/jvm/impl.bzl
```diff
@@ -41,25 +41,22 @@ def _make_providers(ctx, providers, transitive_files = depset(order = "default")
     files = [ctx.outputs.jar]
     if providers.java.outputs.jdeps:
         files.append(providers.java.outputs.jdeps)
-    return struct(
-        kt = providers.kt,
```
Bencodes, line 45: I think this is going to cause issues in the intellij plugin where it looks for this specific field. Example:
- https://github.com/bazelbuild/intellij/blob/master/aspect/intellij_info_impl.bzl#L569-L570
- https://github.com/bazelbuild/intellij/blob/master/aspect/intellij_info_impl.bzl#L1024-L1025

comius: You're correct, this would cause issues to IntelliJ integration. Struct providers have been deprecated for a long time, and this is one of the very last uses that I detected on Bazel downstream tests. (There are other failures, but mostly due to the use of old versions of other dependencies).

IntelliJ aspect was also until recently using a struct provider for Android rules, that was removed recently. I'll procure a fix for IntelliJ/Kotlin integration. Do you think we should block this PR until that is released?

I'd really like to flip the flag in Bazel @head to disable struct providers and we're now in the tail of the tail of cleaning up the uses.

Bencodes: > IntelliJ aspect was also until recently using a struct provider for Android rules, that was removed recently. I'll procure a fix for IntelliJ/Kotlin integration. Do you think we should block this PR until that is released?

I'm all for merging this as long as there is a plan in place to fix the Intellij side that they are bought into merging. It would be unfortunate if we drop the struct here and then the Intellij team doesn't merge a similar fix, leaving Kotlin users in limbo.

comius: There's another problem. I started looking how to fix Intellij. rules_kotlin is using KtJvmInfo here. However, the provider is only exposed through @rules_kotlin//kotlin/internal:defs.bzl. Is it ok to depend on that in IntelliJ, or should we first move it to a more public location?

I also wonder why couldn't IntelliJ use just JavaInfo provider. It seems that there's less information in there than in KtJvmInfo (https://github.com/bazelbuild/intellij/pull/1202). But using JavaInfo would mean nothing needs to be changed in IntelliJ. Maybe we can put more information there?

cc @jin

Bencodes: There might be some logic that depends on knowing that a target is a "kotlin" in order to enable some Kotlin specific features inside Intellij. I believe this is also the case for Android targets.

If that's not true then in theory it should be fine to just check that final `JavaInfo` provider. I'd defer to the Intellij team for these questions though since they wrote and maintain the Kotlin integration in the aswb/ijwb plugins.

jin: It's been a while since I looked at this, so memory's bit hazy.

Just by skimming the code, ISTM that as long as we can provide IntelliJ [here](https://github.com/bazelbuild/intellij/blob/master/aspect/intellij_info_impl.bzl#L1024-L1025) with a "marker provider" about a target's Kotlin-ness, then consolidate on `JavaInfo` on everything else, things should continue to work?

comius: Ok, what I figured out is that JavaInfo already provides all the information needed.  This was done through the years with improvements to Java compilation process. (Change that made it seems to be https://github.com/bazelbuild/rules_kotlin/commit/2be75c0151b05a4ac1b17ee509e9d7737c8ccc67)

For the toolchains part, a way to expose it to IntelliJ is via _toolchain attribute. As far as I researched, it's not used by Android dexer. So I replaced it with Kotlin toolchain, which provides language informatin. We are doing exactly the same thing internally.

To support reading in IntelliJ I prepared:
https://github.com/bazelbuild/intellij/pull/6394
(the IntelliJ plugin CI is ATM broken because of a different change)

comius: https://github.com/bazelbuild/intellij/pull/6394 was merged.

@Bencodes friendly ping

Bencodes (approved):

Issue #843, opened 2022-10-10
JDeps compiler plugin doesn't work under the new k2 compiler


rockwotj:
I tried to play with the new k2 compiler now that in 1.7.20 the abi-gen plugin is supported, but the jdeps plugin needs to be updated to support k2. Here is the error that I get:

```
warning: ATTENTION!
 This build uses experimental K2 compiler:
  -Xuse-k2
error: there are some plugins incompatible with K2 compiler:
  io.bazel.kotlin.plugin.jdeps.JdepsGenComponentRegistrar
Please remove -Xuse-k2
```

From the kotlin slack, it seems that the rough guide for k2 plugins are [here](https://github.com/JetBrains/kotlin/blob/master/docs/fir/fir-plugins.md). Mostly we need to switch `JdepsGenExtension` from `AnalysisHandlerExtension` to `FirExtension`

rockwotj:
Seems like this will be possible once https://youtrack.jetbrains.com/issue/KT-51982 is done. I wonder if we should switch the plugin to ksp...

rockwotj:
Full thread: https://slack-chats.kotlinlang.org/t/2910292/it-will-be-done-as-part-of-kapt-support-https-youtrack-jetbr#7f89c88e-2db0-407b-b61d-a9cee20a9faa

Bencodes:
@rockwotj are you able to take a look at fixing up the jdeps plugin so that it supports K2?

rockwotj:
> @rockwotj are you able to take a look at fixing up the jdeps plugin so that it supports K2?

I am hoping to have time, but note that it isn't possible right now until k2 has support for KSP.

jbarr21:
has anyone taken up this work now that Kotlin 2.0 is in Beta5?

rockwotj:
Unlikely, I at least no longer work on a Kotlin codebase and won't be able to look into this.

rockwotj:
Looks like this is fixed

PR #1164, opened 2024-04-26
Prepare Jdeps extension for K2 implementation

jbarr21:
This PR makes 3 small changes to the JDeps extension in preparation of adding the K2 impl (#843):

1. Ensure that we set the correct argument to enable K2 in `KotlinJvmTestBuilder.java`. We also avoid resetting the option by modifying the existing infoBuilder instead of using `CompilationTaskInfo.newBuilder()`.
2. Since the K1 & K2 flows will be different, they will track classes in a different order. We need to sort the `deps` in the proto so that we can properly assert their equality.
3. We split out the Jdeps creation logic from `JdepsGenExtension` into a new base class that can be extended by the K2 implementation. There are no logic changes in this refactor.

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt
+++ b/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt
@@ -1,9 +1,7 @@
 package io.bazel.kotlin.plugin.jdeps
 
-import com.google.devtools.build.lib.view.proto.Deps
 import com.intellij.openapi.project.Project
 import com.intellij.psi.PsiElement
-import io.bazel.kotlin.builder.utils.jars.JarOwner
 import org.jetbrains.kotlin.analyzer.AnalysisResult
 import org.jetbrains.kotlin.config.CompilerConfiguration
 import org.jetbrains.kotlin.container.StorageComponentContainer
@@ -48,9 +46,6 @@ import org.jetbrains.kotlin.types.KotlinType
 import org.jetbrains.kotlin.types.TypeConstructor
 import org.jetbrains.kotlin.types.getAbbreviation
 import org.jetbrains.kotlin.types.typeUtil.supertypes
-import java.io.BufferedOutputStream
-import java.io.File
-import java.nio.file.Paths
 
 /**
  * Kotlin compiler extension that tracks classes (and corresponding classpath jars) needed to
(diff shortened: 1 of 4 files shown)


Bencodes (approved):
Can you take a look at the KtLint errors?

jbarr21:
> Can you take a look at the KtLint errors?

@Bencodes already fixed. the [test failure](https://buildkite.com/bazel/rules-kotlin-kotlin/builds/4169#018f1bd2-0873-427a-ab3d-389cc8da9d68/426-921) passes locally though. any thoughts?

```
❯ bazel test //src/test/kotlin/io/bazel/worker:WorkerEnvironmentTest
...
INFO: Found 1 test target...
Target //src/test/kotlin/io/bazel/worker:WorkerEnvironmentTest up-to-date:
  bazel-bin/src/test/kotlin/io/bazel/worker/WorkerEnvironmentTest.jar
  bazel-bin/src/test/kotlin/io/bazel/worker/WorkerEnvironmentTest.jdeps
INFO: Elapsed time: 24.900s, Critical Path: 16.03s
INFO: 365 processes: 17 internal, 312 darwin-sandbox, 36 worker.
INFO: Build completed successfully, 365 total actions
//src/test/kotlin/io/bazel/worker:WorkerEnvironmentTest                  PASSED in 0.4s
```

Bencodes:
Ah this one is flaky. Should be good to merge.

Bencodes (approved):

src/main/kotlin/io/bazel/kotlin/plugin/jdeps/JdepsGenExtension.kt
```diff
@@ -70,9 +65,10 @@ import java.nio.file.Paths
  * @param configuration the current compilation configuration
  */
 class JdepsGenExtension(
-  val configuration: CompilerConfiguration,
-) :
-  AnalysisHandlerExtension, StorageComponentContainerContributor {
+  configuration: CompilerConfiguration,
+) : BaseJdepsGenExtension(configuration),
```
Bencodes, line 69: Any reason this needs to be a base class and not just some independent class that both implementations can call into?

jbarr21: either one works, but I figured I was going to use a common base for the 2 extensions anyways

PR #1165, opened 2024-04-26
Integration test for KSP processing

restingbull:
Adds an integration test that matches the example repo, and fully exercises the module support (no accidental inclusion of development dependencies).


src/test/kotlin/io/bazel/kotlin/integration/KspCompileTest.kt
```diff
@@ -0,0 +1,244 @@
...
+                }
+              }
+            }
+          }
+      }
+      ?: error("Cannot find release repo")
+  }
+
+  @Test
+  fun `test ksp compile`() {
```
Bencodes, line 38: We should really just switch to https://github.com/bazel-contrib/rules_bazel_integration_test so that we aren't having to write all of this Starlark in Kotlin like this.

Issue #808, opened 2022-08-03
klib dependencies support


Kernald:
For Kotlin/JS (and Kotlin/Native, but I guess not Kotlin/JVM), JetBrains created a new dependency format for modules built with the IR back-end, klib - https://kotlinlang.org/docs/native-libraries.html#library-format

Some projects (e.g. arrow-kt) are only publishing klib artifacts for their JS builds, having support for those here would be nice.

jbarr21:
@Bencodes @restingbull it looks like the kotlin 2.0 releases no longer ship a .jar for JS std libs (only the .klib). i think that'll be a blocker for using Kotlin 2.0.0 in the kotlin rules, which will be needed to get JDeps working with 2.0 (#843)

neckbosov:
Hi, are there some updates here? Kotlin 2.0 was released, and without that it's impossible to use rules_kotlin with that version although 2.0 api/language versions are formally supported.

restingbull:
We are considering removing js support for 2.0 -- until such time as we have proper kotlin native support.

@neckbosov -- are you actively using js support?

neckbosov:
@restingbull no, we are not using it.

PR #1166, opened 2024-04-26
Implement Jdeps using K2 APIs

jbarr21:
This PR moves the kotlin_rules to Kotlin 2.0.0 & adds support for running Jdeps when compiling with K2 (#843).

High level list of changes:
- Enable K2 tests in paramaterized test class
- Compile rules_kotlin with Kotlin 2.0 since we require fixes that were made that aren't in K2 preview of v1.9.23
- Add `JdepsGenExtension2` which leverages `FirAdditionalCheckers` plugin extension to collect info from the FIR tree
- We use the `ClassUsageRecorder` to manage collection of AST info
- The deps info in `ClassUsageRecorder` is init in the registrar & dumped to a proto on the `ClassFileFactoryFinalizerExtension` callback since there is no longer a `onAnalysisCompleted` callback in K2
- Deps info is collected through various declaration & expression checkers using 3 main types `FirClassSymbol`, `FirTypeRef`, & `ConeKotlinType` which is then pushed into the `ClassUsageRecorder` so that supertypes & type args can also be collected, if necessary
- Update the `JdepsGenComponentRegistrar` to invoke K1 or K2 impl based on language version set in the kotlinc configuration
- Override the Kotlin language version to 1.9 for KSP task since KSP does not support 2.0
- Update example apps to work with K2

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/k2/ClassUsageRecorder.kt
+++ b/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/k2/ClassUsageRecorder.kt
@@ -1,0 +1,132 @@
+package io.bazel.kotlin.plugin.jdeps.k2
+
+import org.jetbrains.kotlin.fir.analysis.checkers.context.CheckerContext
+import org.jetbrains.kotlin.fir.declarations.utils.sourceElement
+import org.jetbrains.kotlin.fir.resolve.providers.symbolProvider
+import org.jetbrains.kotlin.fir.symbols.impl.FirClassLikeSymbol
+import org.jetbrains.kotlin.fir.symbols.impl.FirClassSymbol
+import org.jetbrains.kotlin.fir.types.ConeKotlinType
+import org.jetbrains.kotlin.fir.types.FirTypeRef
+import org.jetbrains.kotlin.fir.types.classId
+import org.jetbrains.kotlin.fir.types.coneType
+import org.jetbrains.kotlin.fir.types.forEachType
+import org.jetbrains.kotlin.name.ClassId
+import java.nio.file.Paths
+import java.util.SortedSet
+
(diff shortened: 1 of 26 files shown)


restingbull:
This generally looks fine, but I would prefer to wait until the dependencies are out of RC.

Additionally, we have some concerns about backwards compatibility -- I am looking to setup several older version test for our CI.

jbarr21:
> @jbarr21 Kotlin 2.0 is officially out, are you able to push this PR forward using the official release?

I am able to, but won't until y'all decide what you're going to do about [this](https://github.com/bazelbuild/rules_kotlin/issues/808#issuecomment-2052537827)

jbarr21:
@Bencodes now that #1185 has been merged, I've rebased my PR onto the latest master, updated the example apps for K2, & have everything passing except for ktlint on files that I didn't modify

src/main/kotlin/io/bazel/kotlin/plugin/jdeps/k2/ClassUsageRecorder.kt
```diff
@@ -0,0 +1,122 @@
...
+import java.nio.file.Paths
+import java.util.SortedSet
+
+private const val JAR_FILE_SEPARATOR = "!/"
+private const val ANONYMOUS = "<anonymous>"
+
+class ClassUsageRecorder(
+  internal val explicitClassesCanonicalPaths: MutableSet<String> = mutableSetOf(),
+  internal val implicitClassesCanonicalPaths: MutableSet<String> = mutableSetOf(),
+  private val seen: MutableSet<ClassId> = mutableSetOf(),
```
Bencodes, line 23: This seems to be unused, is that intentional?

Bencodes (approved):

Bencodes:
I was able to get an app building within our Android repo using this PR and the K2 Jdps implementation, so we should be good to merge this now.

PR #1167, opened 2024-04-30
Shade Guava to avoid conflicts with ksp plugins

erikkerber:
We have a KSP plugin locally that uses Guava, and would conflict at class load time.

I'm not 100% sure if this rule is the best rule, but experimentally `com.google.common.base.*` and `com.google.*` both broke other parts of the build process.

### DIFF
--- a/src/main/kotlin/shade.jarjar
+++ b/src/main/kotlin/shade.jarjar
@@ -1,1 +1,2 @@
 rule dagger.** io.bazel.kotlin.builder.dagger.@1
+rule com.google.common.** io.bazel.kotlin.builder.guava.@1


Bencodes:
This should be resolved by https://github.com/bazelbuild/rules_kotlin/pull/1139

erikkerber:
It does not unfortunately. This jarjar rule on top of that change still does fix the build on our side though.

EdbertChan:
Linking https://github.com/bazelbuild/rules_kotlin/issues/1138 since the primary issue I was running into was that Dagger is picking up the Guava classes packed with KotlinCompiler.jar

restingbull:
I'm somewhat inclined to suggest you shade the KSP plugin that uses Guava, tbh.

I suspect the correct answer is that _everyone_ shades guava.

restingbull (approved):

PR #1198, opened 2024-07-16
Update rules_java to 7.7.2

Bencodes:


restingbull:
This makes older bazel version testing painful.

Not sure what the right solution is.

restingbull (approved):

PR #1205, opened 2024-08-09
Update rules_android to 0.5.1

Bencodes:


examples/jetpack_compose/WORKSPACE
restingbull, line 1: Can you check if this works? It doesn't have the .bazelrc changes the others do.

restingbull (approved):

Bencodes:
Related https://github.com/bazelbuild/rules_jvm_external/pull/1215, might be worth while holding off on this PR a bit longer to sort out the `--experimental_google_legacy_api` situation, otherwise all `rules_kotlin` users will be forced to enable that flag.

Bencodes:
All of the native providers are being moved into Starlark which will unblock this PR https://github.com/bazelbuild/rules_android/commits/main/

ahumesky:
See https://github.com/bazelbuild/rules_android/issues/256 for updates on the provider Starlark migration

PR #1213, opened 2024-09-06
[toolchains] Add native.toolchain arguments to define_kt_toolchain

### LINKED ISSUE #1206: How do define multiple Toolchains?
I would like to be able to define and use multiple Kotlin toolchains in my project, which uses bzlmod. I would like to be able to have one toolchain with `strict_deps` and `unused_deps` and one without. 

This itself is a byproduct of how `jvm_test_suite` works. For convenience sake, since you have one shared set of test dependencies, the dependencies supplied to jvm_test_suite must consist of the common dependencies of all your tests. At the same time, some of those dependencies may also go unused (eg. test-coroutines may not be used in every test). 

The declaration of multiple toolchains itself is possible via the `define_kt_toolchain` macro, but the macro elides the `exec_compatible_with` and `target_compatible_with` attributes of `native.toolchain`, which is how I think I should select the appropriate toolchain. For whatever reason, the docs discourage users from directly defining Kotlin toolchains.

Is there another way to define and select among multiple toolchains that _is_ encouraged? If not, perhaps you have guidance on dealing with the test dependencies issue?

Thank you!

restingbull:
fixes #1206

### DIFF
--- a/scripts/generate_docs
+++ b/scripts/generate_docs
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+bazel build //kotlin:stardoc || fail "docs did not generate"
+
+cp -f bazel-bin/kotlin/kotlin.md docs/ || fail "couldn't copy"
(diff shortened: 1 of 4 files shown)


scripts/generate_docs
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+bazel build //kotlin:stardoc || fail "docs did not generate"
+
+cp -f bazel-bin/kotlin/kotlin.md docs/ || fail "couldn't copy"
```
Bencodes, line 5: Did you intend to include this in this PR?

Bencodes (approved):

scripts/generate_docs
restingbull: Yeah, makes it easier to update the docs.

PR #1215, opened 2024-09-12
Update released_rules_kotlin to 1.9.6

Bencodes:
Tracking the upstream releases of this.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -30,7 +30,6 @@ rules_kotlin_bootstrap_extensions = use_extension(
 )
 use_repo(
     rules_kotlin_bootstrap_extensions,
-    "kt_java_stub_template",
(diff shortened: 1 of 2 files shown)


Bencodes:
We may need to actually point this version at something closer to HEAD so we can get more of the Kotlin 2.0 changes.

MODULE.bazel
```diff
@@ -30,7 +30,6 @@ rules_kotlin_bootstrap_extensions = use_extension(
 )
```
jbarr21, line 30: should rules_kotlin also be updated from 1.8.1 to 1.9.6?

restingbull: For completeness, yes.

Practically, it doesn't do anything.

restingbull (approved):

PR #1216, opened 2024-09-12
Add support for Kotlin 2.0.10

Bencodes:
Pulling in the incremental update while we sort out what's happening with https://github.com/bazelbuild/rules_kotlin/pull/1209

This also adds better support for the latest versions of KSP where serialization is needed on the classpath.


src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
```diff
@@ -104,6 +104,27 @@ class KotlinToolchain private constructor(
         ).toPath()
     }
 
+    private val KOTLINX_SERIALIZATION_CORE_JVM by lazy {
```
restingbull, line 107: The keep decomposing the std libs....

restingbull (approved):

src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
jbarr21: this was a new dep added on ksp's symbol-processing

PR #1218, opened 2024-09-12
Move to the official Bazel worker API

Bencodes:


Bencodes:
This is a prototype. The worker API rules don't support the WORKSPACE path so we'd probably need to do some compat. work if we want to ship this.

restingbull (approved):

restingbull:
> This is a prototype. The worker API rules don't support the WORKSPACE path so we'd probably need to do some compat. work if we want to ship this.

Direction looks good. Think we can strip more of the worker code out, though. Will take a stab at it.

PR #1221, opened 2024-09-19
Restingbull/bencodes patch 4

restingbull:


agluszak:
I did an experiment and bumped kotlinc to 2.1.0-Beta1... still the same error :(

Issue #1168, opened 2024-05-06
KSP subplugin conflicts with guava in @com_github_jetbrains_kotlin//lib/kotlin-compiler.jar


xinzhengzhang:
Description:

There are some guava classes in Kotlin-compiler, such as com.google .common.graph, if guava is also used in the subplugin of ksp, then because the parent delegation model of java will go to the father, that is, the guava in [MemoryBasedClassLoader](https://github.com/JetBrains/kotlin/blob/master/compiler/preloader/src/org/jetbrains/kotlin/preloading/MemoryBasedClassLoader.java), which will lead to

```
Exception: java.lang. LinkageError: loader constraint violation: loader'app 'wants to load interface com.google common.base. Supplier. A different interface with the same name was previously loaded by org.jetbrains.kotlin.preloading.MemoryBasedClassLoader @4bbfb90a. (com.google common.base. Supplier is in unnamed module of loaderorg.jetbrains.kotlin.preloading.MemoryBasedClassLoader @4bbfb90a, parent loader'bootstrap')
```

The same error occurs with compile ksp in [cli mode](https://kotlinlang.org/docs/ksp-command-line.html)

Workaround:
Essentially, Kotlinc should solve the problem of isolation from the KSP environment, but considering that rules_kotlin builders currently have a relatively complete guava, I filtered out guava in the MemoryBasedClassLoader load condition, but this solution does not look elegant, and I would like to discuss how to deal with it.

```patch
diff --git a/src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt b/src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
index 18ea6c1..0455481 100644
--- a/src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
@@ -20,6 +20,7 @@ import io.bazel.kotlin.builder.utils.BazelRunFiles
 import io.bazel.kotlin.builder.utils.resolveVerified
 import io.bazel.kotlin.builder.utils.verified
 import io.bazel.kotlin.builder.utils.verifiedPath
+import org.jetbrains.kotlin.preloading.ClassCondition
 import org.jetbrains.kotlin.preloading.ClassPreloadingUtils
 import org.jetbrains.kotlin.preloading.Preloader
 import java.io.File
[...]

PR #1222, opened 2024-10-01
Remove Guava shade

erikkerber:
I had made this addition with #1167, which addressed at the time a class load conflict with an internal KSP plugin we have.

This did not solve all of our class loader problems with KSP, and eventually needed to apply a patch from #1168. I don't believe this rule should be necessary, so I'm proposing to revert.


nkoroste:
I think it's probably fine to leave shaded?

erikkerber:
It might be. I can close now but I also want to double check if this actually broke when being combined with the workaround in 1168.

PR #1225, opened 2024-10-04
[infra] Use kotlin standard libraries from the toolchain, as opposed to hard coded.

restingbull:
Modifies the toolchain to provide a common set of stdlib for the builder and jdeps merger via the classpath advice on a java binary.

Also extends the `examples/trivial` to validate the built binary executes as part of testing the java/kotlin integration.

Aside from improving correctness, the hope is that this will fix the link error seen when upgrading to newer versions of kotlinc.

### DIFF
--- a/examples/trivial/BUILD
+++ b/examples/trivial/BUILD
@@ -27,3 +27,15 @@ ktlint_config(
     experimental_rules_enabled = False,
     visibility = ["//visibility:public"],
 )
+
+sh_test(
+    name = "test_execution",
+    srcs = ["test_execution.sh"],
+    args = [
+        "$(execpath //app:myapp)",
+    ],
+    data = [
+        "//app:myapp",
+    ],
+    deps = [],
+)
--- a/examples/trivial/BUILD
+++ b/examples/trivial/BUILD
@@ -32,10 +32,11 @@ sh_test(
     name = "test_execution",
     srcs = ["test_execution.sh"],
     args = [
-        "$(execpath //app:myapp)",
+        "$(location //app:myapp)",
     ],
     data = [
         "//app:myapp",
+        "@bazel_tools//tools/bash/runfiles",
     ],
     deps = [],
 )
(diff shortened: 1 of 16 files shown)


kotlin/internal/toolchains.bzl
```diff
@@ -200,6 +193,10 @@ _kt_toolchain = rule(
                 "21",
             ],
         ),
+        "js_target": attr.string(
```
Bencodes, line 196: This should have been removed

examples/trivial/BUILD
```diff
@@ -27,3 +27,16 @@ ktlint_config(
     experimental_rules_enabled = False,
     visibility = ["//visibility:public"],
 )
+
+sh_test(
```
Bencodes, line 31: There's a skylib macro for this that asserts if something builds.

Probably worth using just to minimize the surface area of rules Kotlin a bit.

restingbull: Needs to run to test it. Building is easy -- but the kotlin/jvm integration will only fail if it doesn't run.

kotlin/internal/toolchains.bzl
```diff
@@ -122,7 +126,7 @@ _kt_toolchain = rule(
         ),
         "language_version": attr.string(
             doc = "this is the -language_version flag [see](https://kotlinlang.org/docs/reference/compatibility.html)",
-            default = "2.0",
+            default = "1.9",
```
nkoroste, line 129: why are we going back to an older version?

nkoroste (approved):

kotlin/internal/toolchains.bzl
restingbull: I can't merge?

agluszak:
> Aside from improving correctness, the hope is that this will fix the link error seen when upgrading to newer versions of kotlinc.

Yes, this seems to work!

PR #1229, opened 2024-10-22
Bump rules_jvm_external to 6.4

shs96c:


restingbull:
... i suspect the workspace rules are a bit wonk. Looking.

shs96c:
Filed #1230 because you're entirely right.

Issue #752, opened 2022-05-25
Kotlin stdlibs from toolchain can conflict with Maven deps


SanjayVas:
The Kotlin toolchain hard-codes stdlib dependencies from `@com_github_jetbrains_kotlin`. This means that if you're writing a library using rules_kotlin, anything downstream that depends on stdlibs via Maven can get two copies.

At minimum, the stdlib targets in `@com_github_jetbrains_kotlin` should have `maven_coordinates` tags so that libraries exported using rules_jvm_external's `java_export`/`kt_jvm_export` rules will list the appropriate Maven dependencies rather than containing their own copies. If these are also specified as `override_targets` to `maven_install`, then that should also cover downstream Bazel consumers.

What would be more flexible is additionally allowing stdlibs to be specified/overridden in `define_kt_toolchain`. If exposing the `jvm_runtime`/`jvm_stdlibs` attrs is too error-prone, perhaps have a specific label attr for each known stdlib.

Related to #41.

rockwotj:
@SanjayVas one thing we do is set the [-no-stdlib flag](https://github.com/bazelbuild/rules_kotlin/blob/master/src/main/starlark/legacy/kotlin/opts.bzl#L25) in the compiler options and we have a macro that we use and always explictly adds the stdlib from maven coordinates.

SanjayVas:
> @SanjayVas one thing we do is set the [-no-stdlib flag](https://github.com/bazelbuild/rules_kotlin/blob/master/src/main/starlark/legacy/kotlin/opts.bzl#L25) in the compiler options and we have a macro that we use and always explictly adds the stdlib from maven coordinates.

I tried this with rules_kotlin 1.8.1. Unfortunately even with `include_stdlibs = "none"` I'm still seeing stdlib implicitly added. Here's a snippet from a `-kt.jar-0.params` file:

```
-no-stdlib
-Xsam-conversions=class
-Xlambdas=class
-opt-in=kotlin.RequiresOptIn
--direct_dependencies
external/com_github_jetbrains_kotlin/lib/annotations-13.0.jar
external/com_github_jetbrains_kotlin/lib/kotlin-stdlib.jar
external/com_github_jetbrains_kotlin/lib/kotlin-stdlib-jdk7.jar
external/com_github_jetbrains_kotlin/lib/kotlin-stdlib-jdk8.jar
```

Bencodes:
[...]

PR #1232, opened 2024-10-23
define_kt_toolchain: allow overriding kotlin-stdlib

dmivankov:
Related to https://github.com/bazelbuild/rules_kotlin/issues/752


dmivankov:
already present in master :+1:

PR #1236, opened 2024-11-07
Stop suggesting that Java's `exported_plugins` is transitive.

cpovirk:
Apparently it's not:
https://bazel.build/reference/be/java#java_library.exported_plugins


cpovirk:
The current text appears to date from https://github.com/bazelbuild/rules_kotlin/commit/95b642f10fb9c73312edd59bcf6ba6250ef8a2cb#diff-4494836edf6b3f0b67056cb72ef24752d21a8c08eb8f2ca4790780179dd76d81R408.

I didn't immediately find discussion on https://github.com/bazelbuild/rules_kotlin/pull/393 after skimming the review threads as described in https://webapps.stackexchange.com/a/158957/226667.

restingbull (approved):

Issue #1238, opened 2024-12-04
IntelliJ always uses the default toolchain language version


jdai8:
In the move away from legacy struct providers (https://github.com/bazelbuild/rules_kotlin/pull/1157), the `_kt_toolchain` attribute was hardcoded to `//kotlin/internal:default_toolchain_impl`, which doesn't take into account toolchain resolution.

To reproduce, configure a toolchain with a different language version, load in IntelliJ, and verify that the project settings are incorrect.

PR #1231, opened 2024-10-22
Fix workspace resolution for development

restingbull:
Align versions tween bzlmod and workspace
Switch to the fancy bazel ci rules for rbe.
Enable RBE for integration tests (huuuuge difference)

fixes: https://github.com/bazelbuild/rules_kotlin/issues/1230

PR #1239, opened 2024-12-04
Add a current_toolchain target to be referenced by kt_toolchain attribute

jdai8:
Fixes https://github.com/bazelbuild/rules_kotlin/issues/1238

I copied the approach that `rules_java` takes [here](https://github.com/bazelbuild/rules_java/blob/084b75a46739eb531aa3da9ab9fe52c8d44b7b34/toolchains/java_toolchain_alias.bzl#L95).

This will require an accompanying change in the IntelliJ plugin too.

### DIFF
--- a/kotlin/internal/toolchains.bzl
+++ b/kotlin/internal/toolchains.bzl
@@ -342,6 +342,18 @@ def define_kt_toolchain(
         target_settings = target_settings or [],
     )
 
+def _kt_toolchain_alias_impl(ctx):
+    toolchain_info = ctx.toolchains[_TOOLCHAIN_TYPE]
+
+    return [
+        toolchain_info,
+    ]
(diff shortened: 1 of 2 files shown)


restingbull:
FYI @comius

kotlin/internal/toolchains.bzl
```diff
@@ -398,3 +410,8 @@ def kt_configure_toolchains():
     define_kt_toolchain(
         name = "default_toolchain",
     )
+
```
restingbull, line 413: This is both neat and horrible.

restingbull (approved):

restingbull:
RBE failure is fixed in https://github.com/bazelbuild/rules_kotlin/pull/1231

PR #1244, opened 2024-12-14
Create a new "bootstrap" toolchain and compile with 2.1.0

restingbull:
Given that Kotlin 2.1.0 now has a version metadata check during compilation, the idea of bootstrapping kotlin by using the previous is pretty much dead. Probably not the worst thing, as bzlmod did not enjoy the shennanigans necessary to make it work.

This adds a new "cli" toolchain that invokes kotlinc via the established command line interface. Additionally, it sketches out a simpler, more contained toolchain api to build the more complex usecases (such as multiplex workers, jdeps, compiler plugins, etc.) on top of. the end goal is a more agile rule approach -- with the performance optimization (jdeps, class path reductions, etc) tucked behind the simple api.Compiler plugins, linting, and the like should be able to live comfortably in the "production" rule and toolchain combinations.

It also introduces new "core" rules, while limited in scope, should be IDE friendly and generally easier to work with than the previous bootstrapping effort -- and since they use the common providers, they will interoperate with the existing rules.

Changes:
* Workaround change in reflection type resolution for jdk type aliases. see https://github.com/bazelbuild/rules_kotlin/issues/1246
* Update a few dependencies due to byte code shifts.


MODULE.bazel
```diff
@@ -36,7 +36,10 @@ use_repo(
     "released_com_github_jetbrains_kotlin",
 )
 
-register_toolchains("@released_rules_kotlin//kotlin/internal:default_toolchain")
+# Register bootstrap toolchain
+register_toolchains("//src/main/kotlin:bootstrap_toolchain")
+
+register_toolchains("//src/main/starlark/core/compile/...")
```
Bencodes, line 42: Nit: Be explicit about toolchain registration within the rule set so we know what's being set up.

WORKSPACE.dev.bazel
```diff
@@ -21,6 +21,8 @@ load("//kotlin:repositories.bzl", "kotlin_repositories")
 
 kotlin_repositories()
 
+register_toolchains("//src/main/starlark/core/compile/...")
```
Bencodes, line 24: Same here, nice having these be explicit instead of globbing an entire directory tree.

MODULE.bazel
restingbull: Done.

.bazelci/presubmit.yml
```diff
@@ -31,8 +31,8 @@ tasks:
   integration_tests:
     name: "Integration Tests"
     platform: ${{ integration_platform }}
-    test_flags:
-      - "--config=rbe"
+#    test_flags:
```
Bencodes, line 34: Did you mean to leave this one commented out still?

MODULE.bazel
```diff
@@ -22,23 +22,10 @@ use_repo(
...
 )
 
-# Once the released rules_koltin is defined, configure it.
-rules_kotlin_bootstrap_extensions = use_extension(
-    "//src/main/starlark/core/repositories:bzlmod_bootstrap.bzl",
-    "rules_kotlin_bootstrap_extensions",
-)
-use_repo(
-    rules_kotlin_bootstrap_extensions,
-    "released_com_github_google_ksp",
```
Bencodes, line 35: Are these not being synced anymore for some reason?

examples/deps/WORKSPACE
```diff
@@ -37,24 +34,17 @@ load("@rules_jvm_external//:defs.bzl", "maven_install")
...
         "com.google.auto.value:auto-value-annotations:1.6.5",
-        "org.robolectric:robolectric:4.7.3",
     ],
     repositories = [
         "https://maven.google.com",
         "https://repo1.maven.org/maven2",
     ],
 )
 
+register_toolchains("//bzl:experimental_toolchain")
```
Bencodes, line 46: Why does this need to be registered here in the example app? Is this requirement moving forward?

kotlin/internal/toolchains.bzl
```diff
@@ -174,8 +176,8 @@ _kt_toolchain = rule(
             cfg = "target",
         ),
         "jvm_target": attr.string(
-            doc = "the -jvm_target flag. This is only tested at 1.8.",
-            default = "1.8",
+            doc = "the -jvm_target flag.",
+            default = "11",
```
Bencodes, line 180: I'll do this against the main branch, but there are a few more JDK releases that we can add here.

kotlin/internal/defs.bzl
```diff
@@ -23,30 +29,16 @@ load(
 TOOLCHAIN_TYPE = "%s" % Label("//kotlin/internal:kt_toolchain_type")
 
 # Java toolchains
-JAVA_TOOLCHAIN_TYPE = "@bazel_tools//tools/jdk:toolchain_type"
-JAVA_RUNTIME_TOOLCHAIN_TYPE = "@bazel_tools//tools/jdk:runtime_toolchain_type"
+JAVA_TOOLCHAIN_TYPE = _JAVA_TOOLCHAIN_TYPE
```
Bencodes, line 32: Nice clean up here

src/main/starlark/core/compile/cli/BUILD.bazel
```diff
@@ -0,0 +1,40 @@
...
+        "//kotlin/compiler:annotations",
+        "//kotlin/compiler:kotlin-stdlib",
+        "//kotlin/compiler:kotlin-stdlib-jdk7",
+        "//kotlin/compiler:kotlin-stdlib-jdk8",
+        "//kotlin/compiler:kotlinx-coroutines-core-jvm",
+        "//kotlin/compiler:trove4j",
+    ],
+    kotlinc = ":kotlinc",
+    language_version = versions.KOTLIN_CURRENT_COMPILER_RELEASE.version,
+    zip = "@bazel_tools//tools/zip:zipper",
```
Bencodes, line 33: I'd just make this the toolchain default instead of making it customizable. Could even make it internal with `_zip`.

MODULE.bazel
restingbull: I removed the build-kotlin-with kotlin, so we don't use the released rules anymore.

examples/deps/WORKSPACE
restingbull: Defining a toolchain and not registering it means it will never be chosen.

App didn't compile without the toolchain -- complained about all sorts of things.

kotlin/internal/toolchains.bzl
restingbull: Got tired of the warnings. Will revert.

src/main/starlark/core/compile/cli/BUILD.bazel
restingbull: After getting kicked off using Bazel's jarjar, I'm leaving options open.

src/main/starlark/core/repositories/kotlin/capabilities_2.1.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,16 @@
+# Copyright 2024 The Bazel Authors. All rights reserved.
```
Bencodes, line 1: the options are missing from here and it needs to be registered in `src/main/starlark/core/repositories/compiler.bzl`, otherwise all of the 2.1 options won't be usable.

src/main/starlark/core/compile/cli/compile.bzl
```diff
@@ -0,0 +1,126 @@
...
+    # On bazel 6, this check isn't possible...
+    if getattr(java_runtime, "version", 0) >= 17:
+        jvm_flags = jvm_flags + " -Djava.security.manager=allow"
+
+    classpath = path_separator.join(
+        ["${RUNPATH}%s" % (j.short_path) for j in runtime_jars.to_list() + toolchain_info.kotlin_stdlib.transitive_compile_time_jars.to_list()],
+    )
+    needs_runfiles = "0" if java_bin_path.startswith("/") or (len(java_bin_path) > 2 and java_bin_path[1] == ":") else "1"
+
+    actions.expand_template(
```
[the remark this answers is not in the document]
restingbull: Added for completeness of the toolchain. If you want complete freedom to play with the kotlin compiler, you could use this.

Basically, it's a thin wrapper around kotlinc -- it can't support persistent workers (at least I don't see an easy path on it) -- but it will always be forward compatible (until they change the cli interface... I give it two major versions). It's the "no warranty" rules and toolchain for the discerning kotlin developer. I've even added basic plugin support: https://github.com/bazelbuild/rules_kotlin/pull/1251

src/main/starlark/core/compile/cli/toolchain.bzl
```diff
@@ -0,0 +1,95 @@
...
+    java_runtime = ctx.toolchains[JAVA_RUNTIME_TOOLCHAIN_TYPE].java_runtime
+    java_toolchain = ctx.toolchains[JAVA_TOOLCHAIN_TYPE].java
+
+    toolchain_info = KotlincJvmCompileInfo(
+        jvm_target = ctx.attr.jvm_target,
+        api_version = ".".join(ctx.attr.api_version.split(".")[:2]),
+        language_version = ".".join(ctx.attr.api_version.split(".")[:2]),
+        executable_zip = ctx.attr.zip[DefaultInfo].files_to_run,
+        kotlinc = ctx.attr.kotlinc[DefaultInfo].files_to_run,
+        compile_mnemonic = "CliKotlinc",
```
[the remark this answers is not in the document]
restingbull: Ah, right verbose actually does something with K2.

.bazelci/presubmit.yml
restingbull: Fixed.

src/main/starlark/core/repositories/kotlin/capabilities_2.1.bzl.com_github_jetbrains_kotlin.bazel
restingbull: Need to finish automating the process. :p

Bencodes (approved):
👍 pending https://github.com/bazelbuild/rules_kotlin/pull/1244#discussion_r1896169600

PR #1247, opened 2024-12-17
Add support for -Xconsistent-data-class-copy-visibility

Bencodes:
This flag is new in 2.1.0 and will likely cause a lot of compiler warnings for teams as they upgrade.

Underlying docs: https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-consistent-copy-visibility/


restingbull (approved):
Wow. That is going to be a _noisy_ migration.

Issue #1245, opened 2024-12-16
Add support for experimental treatInternalAsPrivate and removePrivateClasses in abi.jar


rbeazleyspot:
This is in relation to https://youtrack.jetbrains.com/issue/KT-65690 and https://youtrack.jetbrains.com/issue/KT-64590

Current ABI jar generation includes not only public classes but internal and in some instances private which I belives can lead to unnecessary build invalidations.

As far as I can tell we would need to include optional configuration of the jvm-abi plugin in or around [here](https://github.com/bazelbuild/rules_kotlin/blob/master/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt#L88-L90)

You might expect something along these lines in that file

```
plugin(plugins.jvmAbiGen) {
  flag("outputDir", directories.abiClasses)
  if(task.info.publicOnlyAbiJar) {
    flag("treatInternalAsPrivate", "true")
    flag("removePrivateClasses", "true")
   }
}
```

I think it would make sense to bundle both these plugin options behind one experimental flag on the rule.

Your thoughts on this?

Im currently exploring how this might look in the codebase but progress is slow (however I do have a failing test for the core part of this)

rbeazleyspot:
If I get this to work im thinking of extending the CompilationTaskInfo proto with 
```    // Include only public symbols in abi.jar
    bool public_only_abi_jar = 12;
```

and adding this attribute to kt_toolchain
```
        "experimental_use_public_only_abi_jars": attr.bool(
            doc = """Compile using public only abi jars.
            This applies the following two compiler plugin options.
            plugin:org.jetbrains.kotlin.jvm.abi:treatInternalAsPrivate=true
            plugin:org.jetbrains.kotlin.jvm.abi:removePrivateClasses=true
            Can be disabled for an individual target using the tag.
            `kt_abi_plugin_incompatible`""",
            default = False,
        ),
```

rbeazleyspot:
Ive opened a [draft pr](https://github.com/bazelbuild/rules_kotlin/pull/1248) for your consideration

Issue #1021, opened 2023-08-31
Associates or friends leak the internals of their compile classpath


Bencodes:
The flatten jar macro currently collects the entire compilation classpath of any target passed is as an associate which gives access to all of the Kotlin `internal` functions of not just the direct associate dependencies, but all of the compile time jars provided by the associate dependencies.

https://github.com/bazelbuild/rules_kotlin/blob/e1a92bc4077fb28456e34329e480cf6bfb0d1e31/kotlin/internal/jvm/associates.bzl#L55-L67

Bencodes:
Assuming this is not the intended behavior, replacing the `JavaInfo#compile_jars` with just the [direct outputs](https://bazel.build/versions/6.0.0/rules/lib/JavaInfo#java_outputs) will likely be a breaking change for some people and need to be feature flagged and gradually rolled out.

restingbull:
This is implementing strict associates. Flag it?

PR #1248, opened 2024-12-18
Add support for experimental treatInternalAsPrivate

rbeazleyspot:
Adds support for experimental options when producing abi.jar, specifically [KT-65690](https://youtrack.jetbrains.com/issue/KT-65690) and [KT-64590](https://youtrack.jetbrains.com/issue/KT-64590) as discussed in  #1245

Have added a couple of integration tests to perform basic checks of flags being applied

I expect there is more to do here and happy to work on it if its considered for merging

### DIFF
--- a/kotlin/internal/toolchains.bzl
+++ b/kotlin/internal/toolchains.bzl
@@ -86,7 +86,8 @@ def _kotlin_toolchain_impl(ctx):
             "supports-multiplex-workers": "1" if ctx.attr.experimental_multiplex_workers else "0",
         },
         experimental_use_abi_jars = ctx.attr.experimental_use_abi_jars,
-        experimental_use_public_only_abi_jars = ctx.attr.experimental_use_public_only_abi_jars,
+        experimental_treat_internal_as_private_in_abi_jars = ctx.attr.experimental_treat_internal_as_private_in_abi_jars,
+        experimental_remove_private_classes_in_abi_jars = ctx.attr.experimental_remove_private_classes_in_abi_jars,
         experimental_strict_kotlin_deps = ctx.attr.experimental_strict_kotlin_deps,
         experimental_report_unused_deps = ctx.attr.experimental_report_unused_deps,
         experimental_reduce_classpath_mode = ctx.attr.experimental_reduce_classpath_mode,
@@ -203,13 +204,18 @@ _kt_toolchain = rule(
             `kt_abi_plugin_incompatible`""",
             default = False,
         ),
-        "experimental_use_public_only_abi_jars": attr.bool(
-            doc = """Compile using public only abi jars.
-            This effectively applies the following two compiler plugin options:
+        "experimental_treat_internal_as_private_in_abi_jars": attr.bool(
+            doc = """This applies the following compiler plugin option:
               plugin:org.jetbrains.kotlin.jvm.abi:treatInternalAsPrivate=true
+            Can be disabled for an individual target using the tag.
+            `kt_treat_internal_as_private_in_abi_plugin_incompatible`""",
+            default = False,
+        ),
+        "experimental_remove_private_classes_in_abi_jars": attr.bool(
+            doc = """This applies the following compiler plugin option:
               plugin:org.jetbrains.kotlin.jvm.abi:removePrivateClasses=true
             Can be disabled for an individual target using the tag.
-            `kt_public_only_in_abi_plugin_incompatible`""",
+            `kt_remove_private_classes_in_abi_plugin_incompatible`""",
             default = False,
         ),
         "experimental_strict_kotlin_deps": attr.string(
@@ -298,7 +304,8 @@ def define_kt_toolchain(
         api_version = None,
         jvm_target = None,
         experimental_use_abi_jars = False,
-        experimental_use_public_only_abi_jars = False,
+        experimental_treat_internal_as_private_in_abi_jars = False,
+        experimental_remove_private_classes_in_abi_jars = False,
         experimental_strict_kotlin_deps = None,
         experimental_report_unused_deps = None,
         experimental_reduce_classpath_mode = None,
@@ -325,7 +332,8 @@ def define_kt_toolchain(
             _NOEXPERIMENTAL_USE_ABI_JARS: False,
             "//conditions:default": experimental_use_abi_jars,
         }),
-        experimental_use_public_only_abi_jars = experimental_use_public_only_abi_jars,
+        experimental_treat_internal_as_private_in_abi_jars = experimental_treat_internal_as_private_in_abi_jars,
+        experimental_remove_private_classes_in_abi_jars = experimental_remove_private_classes_in_abi_jars,
         experimental_multiplex_workers = experimental_multiplex_workers,
         experimental_strict_kotlin_deps = experimental_strict_kotlin_deps,
         experimental_report_unused_deps = experimental_report_unused_deps,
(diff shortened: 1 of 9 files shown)


kotlin/internal/toolchains.bzl
```diff
@@ -202,6 +203,15 @@ _kt_toolchain = rule(
             `kt_abi_plugin_incompatible`""",
             default = False,
         ),
+        "experimental_use_public_only_abi_jars": attr.bool(
+            doc = """Compile using public only abi jars.
+            This effectively applies the following two compiler plugin options.
+            plugin:org.jetbrains.kotlin.jvm.abi:treatInternalAsPrivate=true
+            plugin:org.jetbrains.kotlin.jvm.abi:removePrivateClasses=true
+            Can be disabled for an individual target using the tag.
+            `kt_abi_plugin_incompatible`""",
```
Bencodes, line 212: This should probably have a tag of its own to disable this behavior at the target level so that adoption across larger projects is easier.

Bencodes:
@rbeazleyspot PR looks good to me, can you take a look at the failing CI checks and regenerate the docs?

rbeazleyspot:
> @rbeazleyspot PR looks good to me, can you take a look at the failing CI checks and regenerate the docs?

will do, i'm going to make the change you suggested as integrating into our codebase showed up a couple of targets with interesting (TM) issues

rbeazleyspot:
Ive pushed an update to the rule to use its own tag as suggested, regenerated the docs (there were some reformatting changes that i unstaged, not sure if that was expected or not), and updated the compilation avoidance advice.


Finally Ive since been testing this further in our app and finding i need to apply kt_public_only_in_abi_plugin_incompatible to some associate targets, so im a little concerned here.

I wondered if something in [associates.bzl](https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/internal/jvm/associates.bzl#L56) needs to change, but I havnt looked into this at all yet. It could be our macros are breaking things a little but the effect is certainly that internal access in an associate is not possible ie its linking against the abi jar (first?) or something.

I'll continue to investigate this.


And finally finally, due to the bug mentioned in the compilation avoidance doc i've been testing this locally with compiler 2.1 (after a few tweaks to the rules codebase)... so just a heads up 🖖

rbeazleyspot:
oh, finally finally v3 final..... i'm not sure whats going on with the CI failure, any [pointers](https://xkcd.com/138/)?

rbeazleyspot:
It occurred to me last night that the problem with associates this change has shown is maybe an issue in its own right.... should I open one for this @Bencodes ? ie compilation with associates use compile_jars rather than output jars

I guess this only worked as at the time there were no ABI jars generated so it defaulted to output jars?
Then when the ABI jars were generated they were (incorrectly) including more than just public api so things still just worked ™️

Bencodes:
@rbeazleyspot can you rebase this branch against HEAD, I believe the current failing REB CI job has already been fixed.

kotlin/internal/toolchains.bzl
```diff
@@ -202,6 +203,15 @@ _kt_toolchain = rule(
             `kt_abi_plugin_incompatible`""",
             default = False,
         ),
+        "experimental_use_public_only_abi_jars": attr.bool(
+            doc = """Compile using public only abi jars.
+            This effectively applies the following two compiler plugin options:
+              plugin:org.jetbrains.kotlin.jvm.abi:treatInternalAsPrivate=true
```
Bencodes, line 209: Do these flags need to be paired together to work?

Bencodes:
> It occurred to me last night that the problem with associates this change has shown is maybe an issue in its own right.... 

A new issue would be good for this just to track it. There are a few known edge cases with associates, some of which you've called out here. The ones that I'm aware of are:
1. Associate dep collection brings the entire classpath along with each associate target. This unintentionally leaks some transitive classpath internals beyond the direct associate dependency. I filed an issue for this one a while back https://github.com/bazelbuild/rules_kotlin/issues/1021
2. As you mentioned associates uses the compile jar instead of the output/runtime jar. This one is worth fixing inside the master branch.
3. There are some additional edge cases that will need to be looked at around ensuring that if we collect an output jar for an associate dependency, it's ABI doesn't make it into the list of standard classpath jars. It's been a while since i've looked into this, but I remember finding that if you pass both the ABI and runtime jars to kotlinc you get into some weird territory where one may get picked over the other, produced an unexpected error (maybe this has improved in 2.x).

I hacked together a not-so-amazing solution to all of these issues a while back since we had been testing out these abi-gen-compiler changes internally to see what kind incremental compilation improvements we could expect here. To do this I had to chop up `associates.bzl` quite heavily so that I could build a relationship between the compile and output jars up front, and then swap jars in and out accordingly when constructing that final classpath for the `KotlinCompile` action.

kotlin/internal/toolchains.bzl
rbeazleyspot: in my (limited) experience yes, they only appear to work together... however i will split them, maybe there are variations in compiler versions....

rbeazleyspot: maybe it would be better if we pivoted on this idea; if we keep down this road there will be a parallel release train with the kotlin compiler. Each option added for the abi plugin will be mirrored in a similar commit/release here. Could we instead have some `experimental_abi_plugin_flags` dictionary or similar that is unpacked when configuring the plugin in KotlinJvmTaskExecutor ?

On the other hand i quite like the idea of restricting what options are available such that these rules can be opinionated about how to build abi jars and increase compilation avoidance :D

rbeazleyspot: it looks like [this](https://github.com/JetBrains/kotlin/blob/9df698ee9f3693d2002b269ab944d0f5dcb75481/plugins/jvm-abi-gen/src/org/jetbrains/kotlin/jvm/abi/JvmAbiClassBuilderInterceptor.kt#L250-L255) conditional means both flags are needed as its gated behind `removePrivateClasses`...

```
private fun shouldRemoveFromAbi(irClass: IrClass?, removePrivateClasses: Boolean, treatInternalAsPrivate: Boolean): Boolean = when {
    irClass == null -> false
    irClass.isFileClass -> false
    removePrivateClasses -> irClass.isVisibilityStrippedFromAbi(stripInternal = treatInternalAsPrivate)
    else -> false
}
```

rbeazleyspot: anyway, i've pushed another commit that splits these flags... its no effort to revert that commit if we so wish

rbeazleyspot:
> > It occurred to me last night that the problem with associates this change has shown is maybe an issue in its own right....
> 
> A new issue would be good for this just to track it. There are a few known edge cases with associates, some of which you've called out here. The ones that I'm aware of are:
> 

https://github.com/bazelbuild/rules_kotlin/issues/1250

rbeazleyspot:
happy new year @Bencodes, hope you got to switch off over the holiday period ... what shall we do about this PR ?

ive not had time to look at PR-ing something for https://github.com/bazelbuild/rules_kotlin/issues/1250 but with a fair wind i could try this month

rbeazleyspot:
> @rbeazleyspot I'll take another look at this one this week after I do some additional testing.

thanks @Bencodes ... i'm also in the bazel slack workspace if you wanted potentially shorter feedback comms

Bencodes (approved):

Bencodes (approved):

PR #1254, opened 2024-12-30
Query the release target in the integration tests

Bencodes:
Adding a Bazel query to make sure that the release archive is structurally sound.


restingbull:
... should I ask?

restingbull (approved):

PR #1260, opened 2025-01-12
Add support for strict associate deps

Bencodes:
Enables support for strict Kotlin associate dependencies.

When `--@rules_kotlin//kotlin/settings:experimental_strict_associate_dependencies=True` is enabled, `rules_kotlin` will collect ONLY the direct `java_output.compile_jar` for each given associate dependency, ensuring that only the Kotlin internals for the targets explicitly listed out as associates will be exposed.

This behavior differs from what's currently in `rules_kotlin` where it collects the entire transitive compile jars classpath for each associate dependency, which technically leaks the Kotlin internals of the entire classpath for each associate dependency.

https://github.com/bazelbuild/rules_kotlin/issues/1021

### DIFF
--- a/kotlin/internal/jvm/associates.bzl
+++ b/kotlin/internal/jvm/associates.bzl
@@ -30,14 +30,14 @@ def _collect_associates(ctx, toolchains, associate):
     them as a depset.
 
     There are two outcomes for this marco:
-    1. When `experimental_prune_transitive_deps` is enabled and the tag override has not been provided, only the
+    1. When `experimental_strict_associate_dependencies` is enabled and the tag override has not been provided, only the
         direct java_output compile jars will be collected for each associate target.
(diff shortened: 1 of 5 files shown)


restingbull:
We should add a readme section for this.

kotlin/internal/jvm/associates.bzl
```diff
@@ -39,7 +56,7 @@ def _get_associates(ctx, associates):
         jars = []
         module_names = []
         for a in associates:
-            jars.append(depset(transitive = [a[JavaInfo].compile_jars, a[_KtJvmInfo].module_jars]))
+            jars.append(_collect_associates(ctx = ctx, toolchains = toolchains, associate = a))
```
Bencodes, line 59: Unclear if the `a[_KtJvmInfo].module_jars` still needs to be included here. We don't hit this case in our codebase and I don't have enough historical context around why we need this here.

restingbull (approved):

Issue #1267, opened 2025-01-28
rules_kotlin should migrate off `@bazel_tools//tools/java:java_stub_template.txt`


hvadehra:
It's used in a couple of places https://github.com/bazelbuild/rules_kotlin/blob/096170ffd892817044d56e8959c5e73a5856a98d/src/main/starlark/core/compile/cli/toolchain.bzl#L74 and https://github.com/bazelbuild/rules_kotlin/blob/096170ffd892817044d56e8959c5e73a5856a98d/kotlin/internal/jvm/jvm.bzl#L139

The template now lives in `@rules_java` and in Bazel@HEAD the target in `@bazel_tools` is just a redirect. Given that the template is no longer versioned with Bazel, it can no longer be considered stable, and the Bazel team is looking to make bazel_tools empty in future releases anyway.

The code expanding the template must remain in sync with the template, the recommendation is to fork the template and maintain a local copy that matches the template expansion code. This will enable rulesets to evolve independently and keep the template always in sync with the rule code.

PR #1270, opened 2025-01-30
Pull java_stub_template.txt as an http_file

Bencodes:


MODULE.bazel
```diff
@@ -19,6 +19,7 @@ use_repo(
     "com_github_jetbrains_kotlin",
     "com_github_jetbrains_kotlin_git",
     "com_github_pinterest_ktlint",
+    "java_stub_template_txt",
```
Bencodes, line 22: This is being removed from Bazel and doesn't seem to be exposed by rules_java anymore.

restingbull:
Per https://github.com/bazelbuild/rules_kotlin/issues/1267, we need to copy the template into the repo.

 @nkoroste had issues with CI downloading this, and it will probably go away/move in the future.

restingbull (approved):

PR #1272, opened 2025-02-10
Make KSP work with IntelliJ

ben-pomelo:


Bencodes:
@ben-pomelo did you mean to close this one?

ben-pomelo:
@Bencodes Yes, apologies. This was meant to be a PR just in my own fork: https://github.com/ben-pomelo/rules_kotlin/pull/1 I did not intend to push this back to main. Do you think it's a valuable fix that everyone would benefit from? TBH i am still trying to get this to work after upgrading to Bazel 8 in our repo but no success yet. I'm not sure if this is the right fix!

I'm having the same problem reported here: https://youtrack.jetbrains.com/issue/BAZEL-1559/KSP-generated-source-are-unknown-to-Intellij

Bencodes:
@ben-pomelo if you end up finding something that works, we'd happy take the patches. We ended up solving this internally on the ASWB plugin side by unpacking the generated jars and passing them to Intellij that way.

ben-pomelo:
@Bencodes We were able to get it working with the following PR: https://github.com/ben-pomelo/rules_kotlin/pull/2

@rob-pomelo and I were playing with some alternative solutions, and found that it also works if we return the generated src and class jars from `_run_kt_java_builder_actions` explicitly and add another pack call:

```
    generated_class_jar = java_common.pack_sources(
        ctx.actions,
        output_source_jar = ctx.actions.declare_file(ctx.label.name + "-genclass.jar"),
        class_jars = generated_class_jars,
        java_toolchain = toolchains.java,
    ) if generated_src_jars else None
```

Which may be a preferable solution to the maintainers here. 

For now we are unblocked via using a bazel patch to get our KSP working again 🎉 but i'm down to work together to bring these changes back to the main branch if you are interested! I just want to get approval that our changes aren't specific to our KSP needs and work generically for all users.

Bencodes:
The PR that you originally put up is a good starting point. I can leave some comments on that once you have it up.

Adding an additional pack_sources call might be the safer approach here as I believe there is a world where both KSP and Kapt can generate source/class jars.

PR #1274, opened 2025-02-13
Fixes for Windows issues found in the course of updating rules_jvm_external to use Starlark Android rules

ahumesky:
In the course of updating rules_jvm_external to use the Starlark Android rules here: https://github.com/bazel-contrib/rules_jvm_external/pull/1297 a couple issues were found in rules_jvm_external's Windows tests:

1) Use platform-specific path separators
2) Add Kotlin standard library jars to the `data` of tool targets so that the Windows Java launcher can find them in the runfiles manifest

This fixes issues like this in Kotlin tools:
`LAUNCHER ERROR: Rlocation failed on _main/external/rules_kotlin++rules_kotlin_extensions+com_github_jetbrains_kotlin_git/lib/annotations-13.0.jar, path doesn't exist in MANIFEST file`

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/cmd/BUILD.bazel
+++ b/src/main/kotlin/io/bazel/kotlin/builder/cmd/BUILD.bazel
@@ -1,13 +1,5 @@
+load("//kotlin/compiler:compiler.bzl", _KOTLIN_STDLIBS = "KOTLIN_STDLIBS")
 load("//src/main/kotlin:bootstrap.bzl", "kt_bootstrap_binary", "kt_bootstrap_library")
-
-_KOTLIN_STDLIBS = [
-    "//kotlin/compiler:annotations",
-    "//kotlin/compiler:kotlin-stdlib",
-    "//kotlin/compiler:kotlin-stdlib-jdk7",
-    "//kotlin/compiler:kotlin-stdlib-jdk8",
-    "//kotlin/compiler:kotlinx-coroutines-core-jvm",
-    "//kotlin/compiler:trove4j",
-]
 
 kt_bootstrap_library(
     name = "build_lib",
--- a/src/main/kotlin/io/bazel/kotlin/builder/cmd/BUILD.bazel
+++ b/src/main/kotlin/io/bazel/kotlin/builder/cmd/BUILD.bazel
@@ -1,5 +1,14 @@
 load("//src/main/kotlin:bootstrap.bzl", "kt_bootstrap_binary", "kt_bootstrap_library")
 
+_KOTLIN_STDLIBS = [
+    "//kotlin/compiler:annotations",
+    "//kotlin/compiler:kotlin-stdlib",
+    "//kotlin/compiler:kotlin-stdlib-jdk7",
+    "//kotlin/compiler:kotlin-stdlib-jdk8",
+    "//kotlin/compiler:kotlinx-coroutines-core-jvm",
+    "//kotlin/compiler:trove4j",
+]
+
 kt_bootstrap_library(
     name = "build_lib",
     srcs = ["Build.kt"],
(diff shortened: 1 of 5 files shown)


ahumesky:
Looks like this fix breaks a bunch of other stuff -- not terribly shocking unfortunately given the nature of the changes. I'll make this a draft until this passes CI

ahumesky:
Looks like the integration tests are failing with `name 'bzl_library' is not defined (did you mean 'objc_library'?)` which doesn't look related to this change at first glance

restingbull:
> Looks like the integration tests are failing with `name 'bzl_library' is not defined (did you mean 'objc_library'?)` which doesn't look related to this change at first glance

Looking...

qq: why doesn't windows get the stdlibs from runfiles?

ahumesky:
> qq: why doesn't windows get the stdlibs from runfiles?

I'm told it's because on linux, all (or at least more) inputs are added to the runfiles, and this was not done (fixed?) in the windows implementation of runfiles. I looked through bazel's code to try to find the different, but didn't get very far in the time I had

ahumesky:
> Looks like this fix breaks a bunch of other stuff -- not terribly shocking unfortunately given the nature of the changes. I'll make this a draft until this passes CI

The breakages (at least those not in the integration tests) were due to switch from `path` to `short_path`, and this turned out to be because I was testing with Bazel 8 on my Windows machine but Bazel 7 on linux -- so this would seem to mean that that separate from all these changes, Bazel 8 and Bazel 7 are doing something different with runfiles manifests at least on windows -- but that will be a problem for another day. I would expect if we start changing how runfiles manifests work, a lot of stuff would break, so I'm not sure what the story here is.

ahumesky:
The `path` vs `short_path` issue in Bazel 8 vs Bazel 7 appears to be https://github.com/bazelbuild/bazel/issues/23574, adding `--legacy_external_runfiles=true` gets my fixes to rules_jvm_external (https://github.com/bazel-contrib/rules_jvm_external/pull/1297) and this PR working again on Bazel 8

ahumesky:
I updated this PR with the latest commits, and the integration tests are now passing, but curiously `//src/test/kotlin/io/bazel/worker:WorkerEnvironmentTest` is failing, while it passed in previous runs

ahumesky:
`//src/test/kotlin/io/bazel/worker:WorkerEnvironmentTest` recently failed in the same way for other PRs (e.g. https://buildkite.com/bazel/rules-kotlin-kotlin/builds/4818#01951a32-b2b2-48ab-8462-c0644843ce25) so this would no appear to be related to this PR

ahumesky:
With everything passing except for `WorkerEnvironmentTest`, I'll take this out of draft

src/main/kotlin/io/bazel/kotlin/builder/cmd/BUILD.bazel
```diff
@@ -1,5 +1,14 @@
 load("//src/main/kotlin:bootstrap.bzl", "kt_bootstrap_binary", "kt_bootstrap_library")
 
+_KOTLIN_STDLIBS = [
```
ahumesky, line 3: this is repeated in `src/main/starlark/core/compile/cli/BUILD.bazel` below, is there a preferred place to consolidate these?

restingbull: [artifacts.bzl](https://github.com/bazelbuild/rules_kotlin/blob/master/src/main/starlark/core/repositories/kotlin/artifacts.bzl) is the best place for right now.

I've been moving deck chairs of late.

restingbull:
> With everything passing except for `WorkerEnvironmentTest`, I'll take this out of draft

I keep disabling this, and then failing to push the pr.

src/main/kotlin/io/bazel/kotlin/builder/cmd/BUILD.bazel
restingbull: Wait, wait. That will produce a cycle.

restingbull: [here should be safe](https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/compiler/compiler.bzl)

ahumesky: thanks! moved

restingbull (approved):

ahumesky:
thank you!

PR #1277, opened 2025-02-20
Pass module_name from kt_android_library to kt_jvm_library

sukolsak:
Currently, `kt_android_library` targets can only access `internal` declarations in a _single_ other `kt_android_library` target, but not _multiple_. This is because each `kt_android_library` has a distinct default module name and [targets in the `associates` attribute must be in the same module](https://github.com/bazelbuild/rules_kotlin/blob/274ac9cbb61dcf35f27385ef02dfb8003e40fec6/kotlin/internal/jvm/associates.bzl#L47). This change allows users to set the module name of `kt_android_library`, thus enabling multiple `kt_android_library` targets to be in the same module.


Bencodes:
Happy to merge this in the short term to unblock if this is causing issues, but we will be swapping these rules out with actual rules hopefully really soon.

sukolsak:
Yes, if this could be merged, that would be great. Thank you!

restingbull (approved):

PR #1279, opened 2025-02-27
Don't glob for .bazel files in the release archive

Bencodes:


Bencodes:
Seeing some failures because of this on the Bazel central registry side https://github.com/bazelbuild/bazel-central-registry/pull/3893

restingbull (approved):
Ow.

PR #1286, opened 2025-03-18
Update README.md

cgruber:
As of bazelbuild/rules_kotlin#1139 ksp plugins which generate Java seem to be supported.


Bencodes (approved):
Nice catch thanks for correcting this!

Issue #1289, opened 2025-03-24
Support Kotlin 2.1.20


neckbosov:
Kotlin 2.1.20 has just been released, please add support for it, looks like it's necessary to merge #1268, maybe something else.

I use rules_kotlin and kapt and would like to try the new k2 mode for kapt. It seems like there is no way to enable it in the current rules version, in kotlin 2.1.20 it's enabled by default.

Bencodes:
This is now supported in the main branch.

PR #1298, opened 2025-04-09
Add support for Kotlin 2.1.20

Bencodes:
https://github.com/JetBrains/kotlin/releases/tag/v2.1.10

https://github.com/bazelbuild/rules_kotlin/issues/1289


src/main/kotlin/io/bazel/kotlin/plugin/jdeps/k2/JdepsK2Utils.kt
```diff
@@ -28,7 +28,7 @@ internal fun SourceElement.binaryClass(): String? =
   when (this) {
     is KotlinJvmBinarySourceElement -> binaryClass.location
     is JvmPackagePartSource -> this.knownJvmBinaryClass?.location
-    is JavaBinarySourceElement -> this.javaClass.virtualFile.path
+    is VirtualFileBasedSourceElement -> this.virtualFile.path
```
Bencodes, line 31: This is a breaking change, we might be able to work around this with some less-than-amazing reflection in the short term but it'll be pretty ugly.

restingbull:
They really aren't making it easy on us.

restingbull (approved):

nkoroste (approved):

PR #1299, opened 2025-04-10
Kotlin 2.1.10+ prep

Bencodes:
Kotlin 2.1.10 introduced a breaking change where a class that we depend on in `JdepsK2Utils` was renamed and the way you access the virtual file has changed.  Moving over to reflection for backwards compatibility until we can cut ties with the previous class.

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/k2/JdepsK2Utils.kt
+++ b/src/main/kotlin/io/bazel/kotlin/plugin/jdeps/k2/JdepsK2Utils.kt
@@ -21,7 +21,6 @@ private object RefCache {
   }
 }
 
-
 /**
  * Returns whether class is coming from JVM runtime env. There is no need to track these classes.
  *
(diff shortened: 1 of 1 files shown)


src/main/kotlin/io/bazel/kotlin/plugin/jdeps/k2/JdepsK2Utils.kt
```diff
@@ -25,11 +39,16 @@ internal fun DeserializedContainerSource.classId(): ClassId? =
...
-    is JavaBinarySourceElement -> this.javaClass.virtualFile.path
-    else -> null
+  if (this is KotlinJvmBinarySourceElement) {
+    binaryClass.location
+  } else if (this is JvmPackagePartSource) {
+    this.knownJvmBinaryClass?.location
+  } else if (RefCache.jbseClass != null && RefCache.jbseClass!!.isInstance(this)) {
+    val jClass = RefCache.jbseGetJavaClassMethod!!.invoke(this)
+    val virtualFile = jClass!!.javaClass.getMethod("getVirtualFile").invoke(jClass)
+    virtualFile.javaClass.getMethod("getPath").invoke(virtualFile) as? String
```
Bencodes, line 49: Using `!!` here since we are using reflection and probably want to fail-fast if the underlying methods change somehow.

restingbull (approved):
Such fun.

Issue #1193, opened 2024-07-10
_ksp_compiler_plugin_repository_impl Fails on Windows


Poligun:
First of all I think we should check the `exec_result` at this [line](https://github.com/bazelbuild/rules_kotlin/blob/b076b4acbf06bc3833e9f70ee11e5a93afa77ab8/src/main/starlark/core/repositories/ksp.bzl#L23) since it's quite common that `mv` would fail.

```
        result = repository_ctx.execute(args, quiet = False)
        if result.return_code != 0:
            fail("Failed to move jar: %s" % result.stderr)
```

Now if I run `bazel build //kotlin/compiler:symbol-processing-api`, the error I got after adding the lines above:

```
Error in fail: Failed to move jar: java.io.IOException: ERROR: src/main/native/windows/process.cc(202): CreateProcessW("mv" com/google/devtools/ksp/symbol-processing/1.9.23-1.0.20/symbol-processing-1.9.23-1.0.20.jar symbol-processing.jar): The system cannot find the file specified.
```

The issue not reproducible on MacOS or on Linux.

esm-p:
@restingbull I still receive the following error with rules_kotlin version 1.9.6:
`ERROR: D:/bazel_tmp/33apilyc/external/rules_kotlin/kotlin/compiler/BUILD.bazel:23:17: no such target '@com_github_google_ksp//:symbol-processing-cmdline.jar': target 'symbol-processing-cmdline.jar' not declared in package 
'' defined by D:/bazel_tmp/33apilyc/external/com_github_google_ksp/BUILD.bazel (Tip: use `query "@com_github_google_ksp//:*"` to see all the targets in that package) and referenced by '@rules_kotlin//kotlin/compiler:symbol-processing-cmdline'`

Is there something that needs to be done to get this issue resolved?
I am using windows 10 22H2

restingbull:
Interesting. It must not be related to the symlink, as that code path is unused.

esm-p:
Tested this also on windows 11. First it did work, but after updating to version 23H2 it throws the same error again.

esm-p:
Hi @restingbull, is there any update on this topic?

restingbull:
Given that it worked before the update, and I have yet to repro it myself, I'm afraid it exists as an exercise for the user.

The revised implementation does not rely on any bash functions... meaning an error may be a bazel bug. I'll give it another round, but I do not have a clear understanding why windows fails.

[...]

PR #1313, opened 2025-05-27
Use `export_files` to export the KSP JAR files

kpitt:
This fixes the`target '@@rules_kotlin~~rules_kotlin_extensions~com_github_google_ksp//:symbol-processing-cmdline.jar' is not visible from target '@@rules_kotlin~//kotlin/compiler:symbol-processing-cmdline'` error described in [my comment](https://github.com/bazelbuild/rules_kotlin/issues/1193#issuecomment-2912995135) on bazelbuild/rules_kotlin#1193.

The legacy behavior of Bazel is that input source files are implicitly exported with the `default_visibility` specified in the `BUILD` file.  The `--incompatible_no_implicit_file_export` option disables this legacy behavior, making all input source files private by default, and the Bazel documentation recommends that files should always be exported explicitly if they are needed outside the current package:

> Avoid relying on the legacy behavior. Always write an `exports_files` declaration whenever a source file target needs non-private visibility.

(see https://bazel.build/concepts/visibility#source-file-target-visibility)

I was not able to reproduce the similar `target 'symbol-processing-cmdline.jar' not declared in package  '' defined by D:/bazel_tmp/33apilyc/external/com_github_google_ksp/BUILD.bazel` error that was also mentioned in bazelbuild/rules_kotlin#1193 for `rules_kotlin` version 1.9.6, so I cannot confirm if this fix also addresses that error.

### DIFF
--- a/src/main/starlark/core/repositories/BUILD.com_github_google_ksp.bazel
+++ b/src/main/starlark/core/repositories/BUILD.com_github_google_ksp.bazel
@@ -13,8 +13,16 @@
 # limitations under the License.
 package(default_visibility = ["//visibility:public"])
 
-# KSP filegroup containing everything that might be needed
+# Collect list of all KSP files that might be needed
+_KSP_FILES = glob(["**"])
+
+# KSP filegroup containing all KSP files to export
 filegroup(
     name = "ksp",
-    srcs = glob(["**"]),
+    srcs = _KSP_FILES,
 )
+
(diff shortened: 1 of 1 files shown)


restingbull:
Thanks for the careful diagnosis!

Windows is being the canary here, as we haven't seen this particular issue elsewhere.

restingbull (approved):

PR #1324, opened 2025-05-31
Apply equivalent of changes from 4e3b18e to release build file

ahumesky:
Add the equivalent to the changes to the development BUILD files in https://github.com/ahumesky/rules_kotlin/commit/4e3b18e552a62eca2f727b7fe6950ee3a3a506a2 to the release BUILD file for jdeps_merger. Fixes for https://github.com/bazel-contrib/rules_jvm_external/pull/1297

### DIFF
--- a/src/main/kotlin/BUILD.release.bazel
+++ b/src/main/kotlin/BUILD.release.bazel
@@ -76,6 +76,14 @@ java_binary(
 
 java_binary(
     name = "jdeps_merger",
+	data = [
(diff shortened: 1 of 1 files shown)


ahumesky:
I was able to use targets like `//kotlin/compiler:annotations` in https://github.com/bazelbuild/rules_kotlin/pull/1274, but these do not exist in the release build files, so i used the equivalent final targets.

Bencodes (approved):

PR #1325, opened 2025-06-02
Add experimental Build Tools API (BTAPI) for compilation

erikkerber:
Enables opting in to building with the Build Tools API (BTAPI). A simpler subset of #1316 

The default configuration for `CompilationService` lands in a similar invocation of `K2JVMCompiler` that the current `BazelK2JVMCompiler` does, so this should very much be a drop-in replacement.

This approach still pins the BTAPI to `2.1.20`, which may result in issues when using a different version of Kotlin. If the BTAPI is never shipped along Kotlin itself and that _is_ a problem, then we'll have to somehow tie the version in Maven to the Kotlin version.


src/main/kotlin/io/bazel/kotlin/compiler/BuildToolsAPICompiler.kt
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2018 The Bazel Authors. All rights reserved.
```
restingbull: Should probably update the copyright on this.

Bencodes (approved):

PR #1326, opened 2025-06-02
Ability to tune unused dependency handling per target

larshagencognite:


kotlin/internal/jvm/jvm.bzl
```diff
@@ -249,6 +249,18 @@ _common_attr = utils.add_dicts(
             providers = [_JavacOptions],
             mandatory = False,
         ),
+        "experimental_report_unused_deps": attr.string(
```
Bencodes: We'd like to avoid collecting excessive attributes like these on the top level rule definitions so that we aren't cluttering our rule definitions over time.

Would you be open to using `tags` to opt targets out of this feature? There are a few examples around the codebase already where we follow this pattern.

larshagencognite: I think tags for the unused deps report mode would work well, I assume you mean like this?
```
kt_experimental_report_unused_deps_off <-> experimental_report_unused_deps="off"
kt_experimental_report_unused_deps_warn <-> experimental_report_unused_deps="warn"
kt_experimental_report_unused_deps_error <-> experimental_report_unused_deps="error"
```
When it comes to defining a list of labels to ignore, I would assume it would have to be something like this?
```
kt_experimental_unused_deps_ignored_target:<target_label1>
kt_experimental_unused_deps_ignored_target:<target_label2>
kt_experimental_unused_deps_ignored_target:<target_label3>
...
```
Seems doable, did I capture your intention?

restingbull: You can leave out the `kt_experimental_` for the tags. Since unused deps is going to be a fun exercise to get right, I expect this api will be around for a while.

PR #1329, opened 2025-06-07
Generate KSP classes jar as output

### LINKED ISSUE #1328: KSP plugin doesn't provide generated classes as a jar
Right now, the `kt_ksp_plugin` with `kt_jvm_library` produces an srcjar https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/internal/jvm/compile.bzl#L424 which can contain the generated Kotlin sources and the `META-INF/` files. However, it doesn't include the generated `.class` files a processor may generate.

An example is illustrated in https://github.com/smocherla-brex/rules_kotlin-ksp-classes (Bazel 7, rules_kotlin 1.9.6, Kotlin 1.9) with Micronaut which generates `.class` files using ASM rather than `.kt` files with KSP. This means the final JAR won't have access to beans generated by KSP and will fail at runtime as illustrated as the created bean cannot be found.

To support this, it seems like an additional output jar can be created additionally with `-ksp-kt-genclasses.jar` and forward it downstream to other targets through `JavaInfo` so that it's available at runtime (or maybe add it to the existing src jar but colocating source files and class files seems odd)? I'd be happy to help contribute it if it makes sense.

smocherla-brex:
Fixes #1328 

Right now, the ksp generated src JAR has the generated sources and files from `META-INF/`. This doesn't suffice for KSP processors (Micronaut) which rely on generated .class files in annotation processing which isn't included in that jar now. We could include these files in that src jar but wasn't sure if it makes sense to include .class files, so I've added another output to the `KotlinKsp` task which generates this jar and is then forwarded as one of the outputs of the `kt_jvm_library` action.

I'll test this internally and verify it fixes our issue (EDIT: confirmed it fixes out issue and beans are found at runtime). I wanted to add tests with Micronaut in `src/test` along with the existing KSP test cases, but Micronaut 4/KSP requires Java 17 which would require upgrading rules_kotlin to Java 17 which could be a breaking change for other users, so I didn't (let me know if I should). Instead I've added tests with a dummy processor that generates bytecode and updated the tests to verify the functionality.


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
```diff
@@ -145,6 +145,9 @@ class KotlinJvmTaskExecutor
           if (outputs.generatedKspSrcJar.isNotEmpty()) {
             context.execute("creating KSP generated src jar", ::createGeneratedKspKotlinSrcJar)
           }
+          if (outputs.generatedKspClassesJar.isNotEmpty()) {
+            context.execute("creating KSP generated classes jar", ::createdGeneratedKspClassesJar)
```
Bencodes, line 151: Would you be able to add some tests for this to `KotlinJvmKspAssertionTest`?

smocherla-brex: I wasn't able to add tests with Micronaut because that'd require upgrading the repo to Java 17 (which introduces other breakages and probably out of scope for this PR). I'm not aware of other KSP processors that generate bytecode on the fly, so I wrote an example/dummy processor that generates class files and then added test cases to validate those class files are present in the final jar. Let me know if that works.

src/test/data/jvm/ksp/bytecodegenerator/processor/src/main/com/example/BytecodeGeneratorProcessor.kt
```diff
@@ -0,0 +1,92 @@
...
+import com.google.devtools.ksp.symbol.KSAnnotated
+import com.google.devtools.ksp.symbol.KSClassDeclaration
+import org.objectweb.asm.ClassWriter
+import org.objectweb.asm.Opcodes
+import src.test.data.jvm.ksp.bytecodegenerator.annotation.GenerateBytecode
+
+/**
+ * An example KSP processor that generates bytecode to be used only for tests
+ */
+class BytecodeGeneratorProcessor(
```
smocherla-brex, line 13: This was written purely for tests to validate the functionality.

smocherla-brex:
Hi @Bencodes, would you be able to take another look at this PR? Thanks in advance!

Issue #894, opened 2022-12-17
Compile Kotlin Android targets natively using the new rules_android Starlark rules


Bencodes:
By moving to the new `rules_android` Starlark rules we can natively compile Android targets.

Bencodes:
Prototype has been pushed but we are waiting for a few pull requests in rules_android to land. https://github.com/bazelbuild/rules_kotlin/pull/868

tdibattista:
Are we still waiting for more progress on the rules_android side of things before this can advance? I'm just wondering what the status is here because this change would save me a lot of headache in getting `kt_android_local_test` to work with our platforms setup.

Bencodes:
@tdibattista It's getting closer to being ready but there are still a hand full of PRs that we'd like to see get merged before we move over to the Starlark rules. It would be great to also see Bzlmod support merged into rules_android.

It's technically possible to switch already but there are a bunch of patches that you need to maintain against rules_android in order for everything to work.

We are still working with the `rules_android` maintainers to get all these PRs landed.

PR #1333, opened 2025-06-16
Support Starlark rules_android

Bencodes:
Swapping out the kt_android_ sandwich for a proper Starlark rule implementation that fully leverages the new processing pipeline inside of `rules_android`.

This pull request contains a few breaking changes that folks will likely need to handle:
1. The `_kt` targets that people may have been referencing in the past no longer exist. If this causes issues we can explore introducing some thin aliases to help ease the migration process.
2. The `//third_party:android_sdk` target that was left public has been removed and no longer exists. Moving forward you should rely on the one provided by `rules_android`.

https://github.com/bazelbuild/rules_kotlin/issues/894


kotlin/internal/jvm/android.bzl
```diff
@@ -11,123 +11,8 @@
...
-        shard_count = shard_count,
-        custom_package = kwargs.get("custom_package", default = None),
-        manifest = manifest,
-        manifest_values = manifest_values,
-        tags = kwargs.get("tags", default = None),
-        testonly = testonly,
-        exec_properties = exec_properties,
-        nocompress_extensions = nocompress_extensions,
-    )
+kt_android_library = _kt_android_library
```
Bencodes, line 17: We can swap this out for a `macro` that aliases the `_kt` targets for people so that we aren't outright introducing breaking changes here.

kotlin/internal/jvm/android.bzl
```diff
@@ -11,123 +11,8 @@
...
+kt_android_local_test_impl = _kt_android_local_test_impl
```
restingbull: Should this be the `kt_andorid_local_test`?

kotlin/internal/jvm/kt_android_library.bzl
```diff
@@ -0,0 +1,53 @@
...
+load(
+    "//kotlin/internal/jvm:kt_android_library_impl.bzl",
+    _kt_android_library_impl = "kt_android_library_impl",
+)
+load(
+    "//kotlin/internal/utils:utils.bzl",
+    _utils = "utils",
+)
+
+_ATTRS = _utils.add_dicts(_BASE_ATTRS, _attrs.lib_common_attr, {
```
restingbull, line 40: Do we expect to have additional attributes here?

kotlin/internal/jvm/compile.bzl
```diff
@@ -987,3 +1029,11 @@ def export_only_providers(ctx, actions, attr, outputs):
             extensions = ["kt", "java"],
         ),
     )
+
```
restingbull, line 1032: 👍  Probably overdue to to do this.

kotlin/internal/jvm/android.bzl
restingbull: Yeah, we ought to do that. Bit of a goof to leave them public.

restingbull (approved):
Generally looks good.

Small nits.

kotlin/internal/jvm/kt_android_library.bzl
Bencodes: None at the moment, left the placeholder here in case we do need them at some point.

kotlin/internal/jvm/android.bzl
Bencodes: I believe they kind of needed to be public to support some specific use cases where you actually needed the Kotlin provider. I lean more towards just doing a breaking change here since it's a relatively easy find-and-replace.

PR #1337, opened 2025-06-18
Ensure that associates are being included in the source dep_infos list

Bencodes:
Ensuring that the top-level dep_infos list contains the complete set of deps collected from both `associates` and `deps`. Without this we have modules in our codebase that fail to compile due to jars missing from the classpath.

I'm only able to encounter this issue with `experimental_prune_transitive_deps` enabled, which makes sense seeing as the non-pruning mode explicitly collects the transitive compile jars from associates.


rbeazleyspot:
🫡

rbeazleyspot:
any other flags set, eg experimental_strict_associate_dependencies ?

Bencodes:
@rbeazleyspot We do also set `experimental_strict_associate_dependencies`.

PR #1340, opened 2025-06-23
Filter abi jar from jdeps classpath

rbeazleyspot:
Following on from [#1337](https://github.com/bazelbuild/rules_kotlin/pull/1337) (a particularly l33t PR), it was noticed that the `--direct_dependencies` (used for the jdeps compiler plugin) include the ABI jar for an associate when instead the runtime class jar should be used.

This is a proposal to remedy this, instead of [generating](https://github.com/bazelbuild/rules_kotlin/pull/1340/files#diff-ca072bf9da461834e9b015042cb38ba7d06b7977889bc1956b4da376b9df2512L472) a list of compile jars from the targets deps at the point of compiler flag generation. This list is [pre-generated](https://github.com/bazelbuild/rules_kotlin/pull/1340/files#diff-16bd765ba02398058bd856317b38e6baa035de90e54c1b12bcfa05567abb3dd0R69) in the jvm_deps functions where the abi jar can be filtered out.

It turns out that this list is identical to the compile_jars list when experimental_prune_transitive_deps is enabled, so when this stops being experimental things should get much simpler.

### DIFF
--- a/kotlin/internal/jvm/associates.bzl
+++ b/kotlin/internal/jvm/associates.bzl
@@ -95,6 +95,11 @@ def _get_associates(ctx, toolchains, associates):
             dep_infos = java_infos,
         )
 
+def _filter_abi_associate_jar(transitive, associates):
+    compile_depset_list = depset(transitive = transitive).to_list()
+    return [jar for jar in compile_depset_list if not _sets.contains(associates.abi_jar_set, jar)]
+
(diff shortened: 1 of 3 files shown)


Bencodes (approved):

Bencodes:
Fixed as part of https://github.com/bazel-contrib/rules_kotlin/pull/1691

PR #1341, opened 2025-06-24
Update Kotlin to 2.2.0

Bencodes:


xinzhengzhang:
In Kotlin 2.2.0, lib/trove4j has been removed. We also need it as a dependency of stdlib.

agluszak:
@Bencodes can be closed now I guess

PR #1343, opened 2025-06-25
Add example analysis_test with configuration

rbeazleyspot:
**WIP**

exploring analysis tests with different configuration properties set

**arrange.bzl** produces a small set of targets 

[main_target_library] --dep--> [dependency_a] --dep--> [dependency_a_trans_dep_jar]

**experimental_prune_transitive_deps_tests.bzl**

has two tests, one test enables experimental_prune_transitive_deps the other dosnt.

basic assertions around the KotlinCompile action to check --classpath, --direct_dependencies and inputs

[dependency_a_trans_dep_jar] should be dropped when the experimental flag is enabled

for these tests to work I think all flags need to be public, ie not the ones in the internal/kotlin/kt_configure_toolchains function

These tests dont run perfectly yet but i thought i would share what i have so far, i have an issue with checking absolute paths atm

```1 expected elements missing from argv:
  0: bazel-out/darwin_arm64-fastbuild/bin/src/test/starlark/rules/enabled_dependency_a.abi.jar
actual argv:
  0: external/+rules_kotlin_extensions+com_github_jetbrains_kotlin_git/lib/annotations-13.0.jar
  1: external/+rules_kotlin_extensions+com_github_jetbrains_kotlin_git/lib/kotlin-stdlib.jar
  2: external/+rules_kotlin_extensions+com_github_jetbrains_kotlin_git/lib/kotlin-stdlib-jdk7.jar
  3: external/+rules_kotlin_extensions+com_github_jetbrains_kotlin_git/lib/kotlin-stdlib-jdk8.jar
  4: bazel-out/darwin_arm64-fastbuild-ST-d9dfb3145612/bin/src/test/starlark/rules/enabled_dependency_a.abi.jar
```
  
  im guessing darwin_arm64-fastbuild-ST-d9dfb3145612 is something to do with the "reconfiguration" when the analysis test runs with the new flags

### DIFF
--- a/src/test/starlark/rules/BUILD.bazel
+++ b/src/test/starlark/rules/BUILD.bazel
@@ -1,0 +1,5 @@
+load(":experimental_prune_transitive_deps_tests.bzl", "experimental_prune_transitive_deps_tests")
+
+experimental_prune_transitive_deps_tests(
+    name = "experimental_prune_transitive_deps_tests",
+)
--- a/src/test/starlark/rules/arrangement.bzl
+++ b/src/test/starlark/rules/arrangement.bzl
@@ -1,0 +1,40 @@
+load("//kotlin:jvm.bzl", "kt_jvm_import", "kt_jvm_library")
+
+def arrange(test):
+    dependency_a_trans_dep_jar = test.artifact(
+        name = "dependency_a_trans_dep.abi.jar",
+    )
+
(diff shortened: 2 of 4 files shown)


agluszak:
@Bencodes this one can be closed, we already merged it in https://github.com/bazelbuild/rules_kotlin/pull/1418

PR #1347, opened 2025-06-29
Try re-enabling Javascript support and add basic support for kotlin-native toolchain

smocherla-brex:
JS support was dropped in https://github.com/bazelbuild/rules_kotlin/pull/118 due to Kotlin 2 switching to klibs from JARS for JS libraries and artifacts. Additionally this also tries to add basic support for kotlin-native toolchain (although unused in this PR as it doesn't seem needed, but maybe useful in the future to support other usecases.)

This aims to re-enable the JS rules in a way that works with the Kotlin 2 architecture. This PR is also added in draft mode to also start a discussion around the feasibility of actually merging this and the maintainer's thoughts on the same. A

Implementation details:
- Add back `kt_js_import` and `kt_js_library` rules. The rules do not deal with JARs at all and klibs are the only way artifacts are passed is through `klibs`, and the final intended binary will be `js`
- Some of the stuff is restored from https://github.com/bazelbuild/rules_kotlin/pull/118 but I've done cleanup to remove any JAR related stuff

I can currently compile the simple examples with `deps` and get klib outputs as an example

```
WARNING: Build options --java_runtime_version and --tool_java_runtime_version have changed, discarding analysis cache (this can be expensive, see https://bazel.build/advanced/performance/iteration-speed).
INFO: Analyzed target //src/test/data/js/deps:main (0 packages loaded, 8489 targets configured).
INFO: Found 1 target...
Target //src/test/data/js/deps:main up-to-date:
  bazel-bin/src/test/data/js/deps/main.klib
INFO: Elapsed time: 5.015s, Critical Path: 4.65s
INFO: 4 processes: 1 internal, 1 darwin-sandbox, 2 worker.
INFO: Build completed successfully, 4 total actions
```

There's an issue with classloader pollution while trying to generate JS files for the final output
```

### CODE REVIEW

File: examples/node/WORKSPACE [hunk 1]
```diff
@@ -0,0 +1,107 @@
...
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+local_repository(
+    name = "rules_kotlin",
+    path = "../..",
+)
+
+# required with Bazel 8/workspace
+http_archive(
```
restingbull, line 11: Why do we need `rules_python`?

kotlin/internal/js/js.bzl
```diff
@@ -0,0 +1,63 @@
...
+    attrs = {
+        "srcs": attr.label_list(
+            doc = "A list of source files to be transpiled to JS",
+            allow_files = [".kt"],
+        ),
+        "deps": attr.label_list(
+            doc = "A list of other kt_js_library that this library depends on for compilation",
+            providers = [_KtJsInfo],
+        ),
+        "output_kind": attr.string(
```
restingbull, line 39: What is the rationale of changing the outputs via attributes (`output_kind`, `sourcemap`)? 

It's not a common pattern for bazel.

restingbull (commented):
first pass. Have some concerns about the api of kt_js_library.

restingbull (changes requested):
Please add starlark unittests, similar to [the existing compile tests](https://github.com/bazelbuild/rules_kotlin/blob/master/src/test/starlark/compile/rule_tests.bzl)

smocherla-brex:
> I'd almost prefer to start with a PR that adds kotlin native toolchain. When I last reviewed getting js working, it appeared to be saner to start with native and then add js as an output.

I didn't do it because kotlinc-js is not part of that, it still seems bundled with the regular Kotlin compiler and there weren't any hooks to invoke the K2JS compiler with kotlin-native (they seem to be just sharing the klib infra for the most part). I can create a separate PR for it though.

kotlin/internal/js/js.bzl
smocherla-brex: It was added to avoid creating `.js` outputs on every `kt_js_library` target as it's only really required in the final target when running it as Node binary (as an example) - only klibs are required for sharing between targets for compilation. The JS creation part seemed quite slow so I made it conditional to avoid slowing down the builds - I can remove it though and just have it generated by default if desired.

examples/node/WORKSPACE
smocherla-brex: This was needed only for this example (not the root module) with Bazel 8 workspace. It appears related to this change with Bazel 8 https://github.com/bazelbuild/bazel/issues/23043. Without this, I get the error

```
USE_BAZEL_VERSION=8.2.1 bazel query //... --enable_workspace --noenable_bzlmod
WARNING: WORKSPACE support will be removed in Bazel 9 (late 2025), please migrate to Bzlmod, see https://bazel.build/external/migration.
ERROR: /private/var/tmp/_bazel_smocherla/bfd93a74a834f9acb8d10b4323733a2c/external/rules_python/python/py_cc_link_params_info.bzl:6:84: name 'PyCcLinkParamsProvider' is not defined
ERROR: error loading package under directory '': error loading package '': Internal error while loading Starlark builtins: Failed to autoload external symbols: compilation of module 'python/py_cc_link_params_info.bzl' failed Most likely you need to upgrade the version of rules repository in the WORKSPACE file.
```

It gets fixed with
```
USE_BAZEL_VERSION=8.2.1 bazel query //... --enable_workspace --noenable_bzlmod --incompatible_autoload_externally=-@rules_python
WARNING: WORKSPACE support will be removed in Bazel 9 (late 2025), please migrate to Bzlmod, see https://bazel.build/external/migration.
//:.aspect_rules_js/node_modules/accepts@1.3.8
//:.aspect_rules_js/node_modules/accepts@1.3.8/dir
//:.aspect_rules_js/node_modules/accepts@1.3.8/pkg
//:.aspect_rules_js/node_modules/accepts@1.3.8/ref
...
```

I can add that flag as well (it appears some internal bazel repos rely on rules_python) and because this is WORKSPACE, it breaks unless it's already pulled in.

kotlin/internal/js/js.bzl
restingbull: If generating the js outputs is too slow, the usual approach is to move them to a separate action. The outputs will only be generated if the artifact is used.

The problem with changing the outputs via attributes is that consumers of the library do not have the option to use the library in a different context without modifying the rule, or duplicating it. If the library is being consumed from a repository or, usually in large code bases, is not owned by the consumer, this becomes a problem.

Additionally, there is the additional cognitive load that developer must know the state of each library when adding them as dependencies. Given that Bazel already requires developers to understand and assemble the build graph _in addition_ to writing code (and declaring dependencies again in the source), it's preferable to make the rules as simple and straight forward as possible.

So, let's generate all the outputs for the first pr -- then, optimize for build time by splitting the slow outputs from the fast in other pr. This will complicate the compile process, so some care will need to be taken to avoid increasing the maintenance burden and complexity of the optimization.

restingbull:
> > I'd almost prefer to start with a PR that adds kotlin native toolchain. When I last reviewed getting js working, it appeared to be saner to start with native and then add js as an output.
> 
> I didn't do it because kotlinc-js is not part of that, it still seems bundled with the regular Kotlin compiler and there weren't any hooks to invoke the K2JS compiler with kotlin-native (they seem to be just sharing the klib infra for the most part). Specifically the [flag](https://github.com/JetBrains/kotlin/blob/646c479d631a3dc66cabdbb53130924a46efd40f/compiler/cli/cli-common/gen/org/jetbrains/kotlin/cli/common/arguments/K2JSCompilerArguments.kt#L218) required to create the JS outputs seemingly exists only with kotlinc-js from my understanding. I can create a separate PR for adding the kotlin-native toolchain though.

As I understand it, the core build graph of non-jvm compilation produces klibs until a leaf node (js, native binary) is needed. I expect jvm compilation will eventually reach that state. Therefore, it makes more sense to ensure a simple klib pipeline, and _then_ look at how the js and native "binaries" are produced.

I'd like to see the native pr to understand how the underlying compilation will change, and see where things can be simplified or are fundamentally complex.

restingbull:
Please hit the "Re-request review" button when you'd like another pass.

smocherla-brex:
> I'd almost prefer to start with a PR that adds kotlin native toolchain. When I last reviewed getting js working, it appeared to be saner to start with native and then add js as an output.

I took a stab at bootstrapping support for the Kotlin/Native toolchain and producing klibs generically to potentially extend it to other platforms (like JS here) https://github.com/bazelbuild/rules_kotlin/pull/1351. It's in draft because it doesn't work on RBE yet (I'm looking into it) but would be interested to see your thoughts on it and if that's the direction you were intending.

examples/node/WORKSPACE
restingbull: ... python is an even bigger mess than I remembered.

PR #1351, opened 2025-07-13
Add initial support for Kotlin Native toolchain and klib compilation

smocherla-brex:
In https://github.com/bazelbuild/rules_kotlin/pull/1347#issuecomment-3039833874, it was mentioned that it's preferred to bootstrap a Kotlin native toolchain first to leverage the new IR compiler to utilize a compilation model to output klibs in a generic way that can be leveraged for JS and potentially other platforms in the future. This PR tries to add that to the toolchain and also adds a new rule `kt_klib_library` rule which is intended to use the Kotlin-native compiler to produce klibs for platform indepedent code that can be shared between various platform specific targets (think of `commonMain` in a gradle project) and can be in the future consumed by platform specific JS/JVM (and maybe WASM targets as well).

Some notes on the implementation details:
1. I had to write a repository rule `kotlin_native_compiler_repository` to bootstrap/download the native distributions for different platforms supported along with its "capabilities" repository similar to how kotlinc is setup. 
2. Make sure to expose `konan.home` or the native compiler files as a TreeArtifact/directory to avoid passing a massive number of files during compilation.
3. Pass the relevant flags during klib compilation with its own task executor with key ones being `produce=library`
4. Additionally, expose alias targets like `//kotlin/compiler:kotlin-native` which is supposed to alias to the native compiler jar/distribution on the execution platform.
5. Add examples for simple one (using stdlib) and one with deps (`kt_klib_library` depending on another target). Add some junit test cases to assert klib has some entries we expect.
6. Add a few starlark unit test cases.


MODULE.bazel
```diff
@@ -19,6 +19,8 @@ bazel_dep(name = "rules_shell", version = "0.4.1")
 
 bazel_dep(name = "buildifier_prebuilt", version = "8.0.3", dev_dependency = True)
 
+bazel_dep(name = "aspect_bazel_lib", version = "2.19.4")
```
smocherla-brex: I added this to leverage `copy_to_directory` to expose the Kotlin native compiler distribution as a directory in the toolchain [here](https://github.com/bazelbuild/rules_kotlin/pull/1351/files#diff-7c672e82b36304d0c98565a503c039e0ab75d7c924a7961857aea0c9c3666357R24) (as it's massive and has a ton of files)

kotlin/internal/klib/klib.bzl
```diff
@@ -0,0 +1,78 @@
...
+    )
+
+    return [
+        DefaultInfo(files = depset(outputs)),
+        _KtKlibInfo(
+            klibs = depset(outputs),
+        ),
+    ]
+
+kt_klib_library = rule(
```
smocherla-brex: This is the new `kt_klib_library` rule that produces klibs using the kotlin/native toolchain. Potentially we could just have it as action and have an attr on the existing rules to trigger klib compilation on them as well (I'm not sure if that's better - let me know)

src/main/kotlin/io/bazel/kotlin/builder/tasks/klib/KotlinKlibTaskExecutor.kt
```diff
@@ -0,0 +1,87 @@
...
+      }
+
+      val execRoot = fileSystem.getPath(".").absolute()
+      return mutableListOf<String>().apply {
+        addAll(passThroughFlagsList)
+        // kotlin-native klib compilation requires caching and no way to disable, so pass a temp directory
+        // within the current working directory to isolate its cache (which is unique to this worker anyway)
+        // Ideally we disable caching though and rely only on Bazel
+        add("-Xauto-cache-dir=${autoCacheDirectory.absolutePathString()}")
+        add("-Xauto-cache-from=${autoCacheFromDirectory.absolutePathString()}")
```
smocherla-brex: Without this, the compiler tries to write within the cache in the external repository where `konan.home` is and bazel errors out. Ideally we disable caching but there doesn't seem to be a flag for this.

src/main/kotlin/io/bazel/kotlin/compiler/BazelK2NativeCompiler.kt
```diff
@@ -0,0 +1,42 @@
...
+class BazelK2NativeCompiler {
+  fun exec(
+    errStream: java.io.PrintStream,
+    vararg args: String,
+  ): ExitCode {
+    System.setProperty("zip.handler.uses.crc.instead.of.timestamp", "true")
+    val konanHome = System.getenv("KONAN_HOME")
+    requireNotNull(konanHome) {"KONAN_HOME env var must be set!"}
+
+    System.setProperty("konan.home", konanHome)
```
smocherla-brex: This is needed for the compiler to find the native code/distribution to produce klib.

src/main/starlark/core/repositories/capabilities.bzl
```diff
@@ -0,0 +1,98 @@
+load("//src/main/starlark/core/repositories/kotlin:templates.bzl", "TEMPLATES")
+
+def _kotlin_capabilities_impl(repository_ctx):
```
smocherla-brex, line 3: This was copied into a separate file to allow leveraging it with `kotlin_native_compiler_repository`

src/main/starlark/core/repositories/BUILD.com_github_jetbrains_kotlin_native.bazel
```diff
@@ -0,0 +1,63 @@
...
+    name = "bin",
+    srcs = glob(["bin/**"]),
+)
+
+write_file(
+    name = "klib_system_cache_marker",
+    out  = "klib/cache/macos_arm64STATIC/klib_cache_marker",
+    content = [
+        "Marker file intended to create the klib system cache placeholder even if we don't use it. The native compiler errors out otherwise",
+        "See See https://github.com/JetBrains/kotlin/blob/v2.1.21/kotlin-native/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt#L567"
```
smocherla-brex: The compiler on bootstrap tries to create `klib/cache/<platform>STATIC` directory in the external repo and this is non-hermetic and causes Bazel to fail. To workaround it (unfortunately there doesn't appear to be a way to disable it with a flag), so we create this marker/dummy file to avoid running into the issue (happy to consider other solutions)

MODULE.bazel
restingbull: Please do not use `copy_to_directory` it's basically antagonistic to RBE.


[copy_to_directory uses](https://github.com/bazel-contrib/bazel-lib/blob/e55168fda556b14f5356602665c2e5ba9737b294/lib/private/copy_to_directory.bzl#L283-L306) uses a tree artifact (via [declare_directory](https://bazel.build/rules/lib/builtins/actions#declare_directory) which is, under the hood, a tarball.) It doesn't save time or number of files. Why? Because bazel only works with files, rather than directories. The intent of that feature is handle actions where the outputs are unknown before execution. `copy_to_directory` _does_ know the inputs (all the files in the glob), and then applies filters etc to the them via an action... Which means we take a well known set of inputs, execute an action, place them in a tarball, _and then unpack them everytime we need them_ (give or take -- there are some optimizations under the hood.)

For RBE, this is ugly: first, transfer all the files into a remote action; second, tar all the outputs and place them in CAS. Given that the files _have not changed_, this bloats the CAS size, increases transfer costs, adds additional actions... for no good reason, other than laziness on the rule writers part. After all, `copy_to_directory` _could_ be written in starlark to do the filtering... or the rules _could_ handle the filtering during globbing, or...

You get the idea.

kotlin/internal/klib/klib.bzl
restingbull: `kt_library` fits the naming convention: `kt_js_library`, `kt_jvm_library`.

kotlin seems to be moving towards using klib as the core intermediate format (js and native), so I suspect java may follow at some point.

src/main/kotlin/io/bazel/kotlin/compiler/BazelK2NativeCompiler.kt
```diff
@@ -0,0 +1,42 @@
...
+import org.jetbrains.kotlin.config.Services
+
+@Suppress("unused")
+class BazelK2NativeCompiler {
+  fun exec(
+    errStream: java.io.PrintStream,
+    vararg args: String,
+  ): ExitCode {
+    System.setProperty("zip.handler.uses.crc.instead.of.timestamp", "true")
+    val konanHome = System.getenv("KONAN_HOME")
```
restingbull: Why is this provided as an environment variable, rather that passed in as an argument?

kotlin/compiler/compiler.bzl
```diff
@@ -63,3 +63,18 @@ def kt_configure_compiler():
 
     _import_artifacts(KOTLINC_ARTIFACTS.jvm, kt_jvm_import)
     _import_artifacts(KOTLINC_ARTIFACTS.core, kt_jvm_import)
+    _import_artifacts(KOTLIN_NATIVE_ARTIFACTS.linux_x86_64, kt_jvm_import, compiler_repo = _KT_NATIVE_COMPILER_REPO_PREFIX + "_linux_x86_64")
+    _import_artifacts(KOTLIN_NATIVE_ARTIFACTS.macos_x86_64, kt_jvm_import, compiler_repo = _KT_NATIVE_COMPILER_REPO_PREFIX + "_macos_x86_64")
+    _import_artifacts(KOTLIN_NATIVE_ARTIFACTS.macos_aarch64, kt_jvm_import, compiler_repo = _KT_NATIVE_COMPILER_REPO_PREFIX + "_macos_aarch64")
+    _import_artifacts(KOTLIN_NATIVE_ARTIFACTS.windows_x86_64, kt_jvm_import, compiler_repo = _KT_NATIVE_COMPILER_REPO_PREFIX + "_windows_x86_64")
+
+    # a convenience alias for kotlin-native to be referenced in other places
+    native.alias(
```
restingbull: What is the usecase for this? It's far preferable to use specific toolchains.

kotlin/internal/klib/klib.bzl
```diff
@@ -0,0 +1,78 @@
...
+    libraries = depset(transitive = deps_klibs)
+    builder_args.add_all("--sources", ctx.files.srcs)
+    builder_inputs, _, input_manifests = ctx.resolve_command(tools = [toolchains.kotlinbuilder, toolchains.konan_home])
+
+    builder_args.add("--strict_kotlin_deps", "off")
+    builder_args.add("--reduced_classpath_mode", "off")
+    builder_args.add("--output_klib", klib.path)
+
+    deps_klibs = []
+    for dep in ctx.attr.deps:
```
restingbull: This is duplicated from line #13?

kotlin/internal/klib/klib.bzl
```diff
@@ -0,0 +1,78 @@
...
+        deps_klibs.append(dep[_KtKlibInfo].klibs)
+    libraries = depset(transitive = deps_klibs)
+    builder_args.add_all("--klibs", libraries, omit_if_empty = False)
+
+    # This will be a directory we need to propagate to the compiler
+    konan_home = toolchains.konan_home[DefaultInfo].files.to_list()[0]
```
restingbull: Rule of thumb: if you unpack a depset with `to_list`, you are doing it wrong.

Let's step back and ask the question: does starlark need to know about konan home?

Or can we derive it in the action? I suspect we could just pass the files into a well known place and work from there.

kotlin/internal/klib/klib.bzl
```diff
@@ -0,0 +1,78 @@
...
+        env = {
+            "REPOSITORY_NAME": utils.builder_workspace_name(ctx),
+            "KONAN_HOME": konan_home.path,
+        },
+    )
+
+    return [
+        DefaultInfo(files = depset(outputs)),
+        _KtKlibInfo(
+            klibs = depset(outputs),
```
restingbull: You'll want to include the transitive outputs here.

src/main/kotlin/io/bazel/kotlin/builder/tasks/klib/KotlinKlibTaskExecutor.kt
restingbull: We've got a different effort looking at incremental caching. This is good enough.

src/main/starlark/core/repositories/BUILD.com_github_jetbrains_kotlin_native.bazel
```diff
@@ -0,0 +1,81 @@
...
+write_file(
+    name = "klib_system_cache_marker_macos_x64",
+    out = "klib/cache/macos_x64STATIC/klib_cache_marker",
+    content = [
+        "Marker file intended to create the klib system cache placeholder even if we don't use it. The native compiler errors out otherwise",
+        "See See https://github.com/JetBrains/kotlin/blob/v2.1.21/kotlin-native/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt#L567",
+    ],
+)
+
+copy_to_directory(
```
restingbull: As noted above, don't do this.

In fact, this is a prefect place to just use a filegroup. We don't need to move these files, just know where they are... enough to set KONAN_HOME.

src/main/starlark/core/repositories/BUILD.com_github_jetbrains_kotlin_native.bazel
```diff
@@ -0,0 +1,81 @@
...
+    name = "konan_home",
+    # TODO: figure out if all files are needed
+    srcs = glob([
+        "konan/**",
+        "klib/**",
+    ]) + [
+        ":klib_system_cache_marker_linux_x64",
```
restingbull: Should these be in all platforms?

src/test/starlark/internal/klib/klib_tests.bzl
```diff
@@ -0,0 +1,53 @@
+load("@rules_testing//lib:analysis_test.bzl", "analysis_test")
+load("@rules_testing//lib:test_suite.bzl", "test_suite")
+load("//kotlin/internal:defs.bzl", "KtKlibInfo")
+load("//kotlin/internal/klib:klib.bzl", "kt_klib_library")
+
+def _common_assertions(env, target):
```
restingbull: Add a test for library -> library to ensure that transitive klibs are included for compilation.

Add test for data to ensure runfiles are included.

kotlin/internal/toolchains.bzl
```diff
@@ -379,6 +385,12 @@ def define_kt_toolchain(
         jvm_runtime = jvm_runtime if jvm_runtime != None else [
             Label("//kotlin/compiler:kotlin-stdlib"),
         ],
+        konan_home = select({
```
restingbull: Each of these should be a separate toolchain with different platform constraints. 

If nothing else, it will make debugging toolchain issues sane.

restingbull (commented):
Good start. Needs some TLC around packaging konan home and configuring toolchians properly.

src/main/kotlin/io/bazel/kotlin/builder/tasks/klib/KotlinKlibTaskExecutor.kt
```diff
@@ -0,0 +1,87 @@
...
+        add("-Xklib-normalize-absolute-path")
```
restingbull: [normalize-absolute-path](https://github.com/JetBrains/kotlin/blob/7038c0d4c963236ade7861755ba72e1b05a0ca46/compiler/ir/serialization.common/src/org/jetbrains/kotlin/backend/common/serialization/IrSerializationSettings.kt#L29-L31) will only fix the slashes.

In order to support rbe and remote caching, we'll need to do something intelligent with relative paths in klibs](https://kotlinlang.org/docs/native-libraries.html#using-relative-paths-in-klibs).

src/main/starlark/core/repositories/versions.bzl
```diff
@@ -74,12 +76,44 @@ versions = struct(
...
         sha256 = "1ba08a8b45da99339a0601134cc037b54cf85e9bc0edbe76dcbd27c2d684a977",
     ),
+    KOTLIN_NATIVE_CURRENT_RELEASE_LINUX_X86_64 = version(
+        version = _DEFAULT_KOTLIN_COMPILER_RELEASE_VERSION,
+        url_templates = [
+            "https://github.com/JetBrains/kotlin/releases/download/v{version}/kotlin-native-prebuilt-linux-x86_64-{version}.tar.gz",
+        ],
+        sha256 = "42fb88529b4039b6ac1961a137ccb1c79fc80315947f3ec31b56834c7ce20d0b",
+        strip_prefix_template = "kotlin-native-prebuilt-linux-x86_64-{version}",
+    ),
```
kitterion, line 92: I'm not sure about the pure klib generating mode, but this will not be enough in the long run. Kotlin Native compiler downloads some host-specific stuff (clang+llvm, libffi, lldb, sysroot on linux) on first launch at runtime which is difficult to express in bazel.

In theory, one could parse the [konan.properties](https://github.com/JetBrains/kotlin/blob/v2.2.0/kotlin-native/konan/konan.properties) file and figure out the necessary artifacts but that seems like a lot of work with no guarantees of stability.

I simply include them with the specific compiler version and platform ([1](https://github.com/kitterion/rules_kotlin_native/blob/d6d201771628ae3742fc44419cb6d1e4fe8f97f7/kotlin_native/versions.bzl#L19)). There might be a better way but even rules_rust has a [giant file](https://raw.githubusercontent.com/bazelbuild/rules_rust/7e9252fb43259f4e3eb379b5e184ca10ea75b5b8/rust/known_shas.bzl) with known artifacts and their hashes.

As a precaution against this, I also specify [`-Xoverride-konan-properties=airplaneMode=true`](https://github.com/kitterion/rules_kotlin_native/blob/d6d201771628ae3742fc44419cb6d1e4fe8f97f7/kotlin_native/internal/native.bzl#L40) (which, annoyingly, is a flag that is not accepted by the cinterop compiler).

src/main/kotlin/io/bazel/kotlin/builder/tasks/klib/KotlinKlibTaskExecutor.kt
kitterion: The link only mentions `-Xklib-relative-path-base`, but [`-Xdebug-prefix-map`](https://github.com/kitterion/rules_kotlin_native/blob/d6d201771628ae3742fc44419cb6d1e4fe8f97f7/tools/KonancWrapper.java#L14) is also required, otherwise the debug info will have absolute paths.

src/main/kotlin/io/bazel/kotlin/builder/tasks/klib/KotlinKlibTaskExecutor.kt
kitterion: This is odd, I've never had this problem.
I think I avoided it because instead of it being implicit, I pass `-nostdlib` (or was it `-no-default-libs` that does this? I pass both) and all the standard library stuff is passed as `-library <path>`.
I see 2 advantages to do it this way:
- it allows to "lift" the cache generation into bazel as separate actions, and the resulting artifacts can be cached by bazel (including remotely instead of just being local)
- cross compilation is much easier for not officially supported host-target combinations (like building macos artifacts on linux). I've yet to test this but I think it should work.

src/main/starlark/core/repositories/kotlin/artifacts.bzl
```diff
@@ -70,9 +70,51 @@ KOTLINC_ARTIFACTS = struct(
     ),
 )
 
+KOTLIN_NATIVE_ARTIFACTS = struct(
+    linux_x86_64 = struct(
+        plugin = {},
+        compile = {
+            "kotlin-native-linux-x86_64": "konan/lib/kotlin-native.jar",
+            "trove4j-linux-x86_64": "konan/lib/trove4j.jar",
```
kitterion: Fun fact, this file is missing in 2.2.0 version. I also had it specified explicitly, but i have since replaced it with a [`glob`](https://github.com/kitterion/rules_kotlin_native/blob/d6d201771628ae3742fc44419cb6d1e4fe8f97f7/kotlin_native/toolchains.bzl#L86).

src/main/kotlin/io/bazel/kotlin/builder/tasks/klib/KotlinKlibTaskExecutor.kt
smocherla-brex: > This is odd, I've never had this problem. I think I avoided it because instead of it being implicit, I pass `-nostdlib` (or was it `-no-default-libs` that does this? I pass both) and all the standard library stuff is passed as `-library <path>`. I see 2 advantages to do it this way:
> 
> * it allows to "lift" the cache generation into bazel as separate actions, and the resulting artifacts can be cached by bazel (including remotely instead of just being local)
> * cross compilation is much easier for not officially supported host-target combinations (like building macos artifacts on linux). I've yet to test this but I think it should work.

Interesting, I'll double check again. If I remember not setting this led to the `klib/cache` folder under the external konan home repository being used as the cached and Bazel errors out as it breaks the sandboxing rules (access denied basically). I'll check if it is an issue (or not) with `-nostdlib`.

> cross compilation is much easier for not officially supported host-target combinations (like building macos artifacts on linux). I've yet to test this but I think it should work.

Interesting, that would make sense - I haven't tested with host-target platforms being different but I'll check and follow-up what my observations are.

src/main/kotlin/io/bazel/kotlin/compiler/BazelK2NativeCompiler.kt
smocherla-brex: Probably should be a flag for consistency but I didn't want to intercept the arguments passed to the compiler and then set it. I can update it though (would be nice if this could be passed as a flag to the compiler but there doesn't seem to be from what I can tell)

src/main/starlark/core/repositories/BUILD.com_github_jetbrains_kotlin_native.bazel
smocherla-brex: I created it for all since Konan sets it up for the target platform and we don't know in advance the target platform a user selects in the repository rule https://github.com/JetBrains/kotlin/blob/v2.1.21/kotlin-native/backend.native/compiler/ir/backend.native/src/org/jetbrains/kotlin/backend/konan/KonanConfig.kt#L532 Probably could be a better way to do it though.

kotlin/internal/toolchains.bzl
smocherla-brex: Makes sense to put it into its own toolchain.

src/main/kotlin/io/bazel/kotlin/builder/tasks/klib/KotlinKlibTaskExecutor.kt
smocherla-brex: Thanks for the pointers, I will update them

src/main/starlark/core/repositories/versions.bzl
smocherla-brex: I see what you mean - there's a lot of embedded stuff in there and I didn't put a lot of thought yet into Cinterop or other native support (I had mostly considered JVM/bS). But I do agree this would not enough especially for host/target platform not being the same. And I think the toolchain we bootstrap here should be extensible to suporot them. It'd be nice to wire up the toolchain from toolchains_llvm for this but I guess there's no way beyond parsing konan.properties and making it understand a hermetic toolchain (assuming it's possible).

> I simply include them with the specific compiler version and platform ([1](https://github.com/kitterion/rules_kotlin_native/blob/d6d201771628ae3742fc44419cb6d1e4fe8f97f7/kotlin_native/versions.bzl#L19)). There might be a better way but even rules_rust has a [giant file](https://raw.githubusercontent.com/bazelbuild/rules_rust/7e9252fb43259f4e3eb379b5e184ca10ea75b5b8/rust/known_shas.bzl) with known artifacts and their hashes.

Wow, that's a massive file. I guess that approach could be ok too if generating it could be automated.

kotlin/internal/klib/klib.bzl
```diff
@@ -10,8 +10,12 @@ def _kt_klib_library(ctx):
 
     toolchains = ctx.toolchains[_TOOLCHAIN_TYPE]
     deps_klibs = []
+    transitive_klibs = []
     for dep in ctx.attr.deps:
         deps_klibs.append(dep[_KtKlibInfo].klibs)
+        deps_klibs.append(dep[_KtKlibInfo].transitive_klibs)
```
restingbull: Does kotlin native support direct dependency compilation?

If not, just add the transitive deps to the `depset` `transitive` argument.

kotlin/compiler/compiler.bzl
smocherla-brex: It was used to pass the kotlin-native.jar to KotlinBuilder but I've created a toolchain now, so will update to reference it from there directly.

MODULE.bazel
smocherla-brex: Updated and removed `copy_to_directory`. I've instead used `directory` from bazel-skylib https://github.com/bazelbuild/bazel-skylib/blob/main/rules/directory/directory.bzl which seems to work as well.

kotlin/internal/klib/klib.bzl
smocherla-brex: It does seem to support it, so I've removed it from being passed to the action (and just keep it in the provider's depset)

src/main/kotlin/io/bazel/kotlin/builder/tasks/klib/KotlinKlibTaskExecutor.kt
smocherla-brex: Actually these flags are not required after removing `copy_to_directory`. I ran into issues when using `konan.home` exposed as an output of `copy_to_directory`. I've switched to `directory` from bazel-skylib which exposes source files in the compiler distribtuion as a directory and it doesn't seem to be happening there.

src/main/kotlin/io/bazel/kotlin/builder/tasks/klib/KotlinKlibTaskExecutor.kt
smocherla-brex: Updated to add `-Xklib-relative-path-base` and `Xdebug-prefix-map` to the compiler invocation.

src/main/starlark/core/repositories/BUILD.com_github_jetbrains_kotlin_native.bazel
smocherla-brex: This is actually no longer required after removing `copy_to_directory` and replacing it with `directory` from bazel-skylib

src/main/starlark/core/repositories/BUILD.com_github_jetbrains_kotlin_native.bazel
smocherla-brex: Removed `copy_to_directory` and instead used `directory` from bazel-skylib

kotlin/internal/klib/klib.bzl
smocherla-brex: Removed this and used `directory` from bazel-skylib and access konan.home from the toolchain and avoid using this depset all together.

src/main/kotlin/io/bazel/kotlin/compiler/BazelK2NativeCompiler.kt
smocherla-brex: Removed this as an env var and instead now pass it as an arg to kotlinbuilder and then `System.setProperty` in the task executor.

src/main/starlark/core/repositories/BUILD.com_github_jetbrains_kotlin_native.bazel
smocherla-brex: This is no longer required after moving away from `copy_to_directory`.

src/main/kotlin/io/bazel/kotlin/builder/tasks/knative/KotlinNativeTaskExecutor.kt
```diff
@@ -0,0 +1,98 @@
+package io.bazel.kotlin.builder.tasks.knative
```
smocherla-brex, line 1: I had use to `knative` instead of `native` in the package name here because `native` is a java keyword and the Dagger annotation processing on KotlinBuilder was failing due to that

src/main/starlark/core/repositories/kotlin/artifacts.bzl
```diff
@@ -70,9 +70,113 @@ KOTLINC_ARTIFACTS = struct(
...
+        },
+        runtime = {},
+        exec_compatible_with = ["@platforms//os:linux", "@platforms//cpu:x86_64"],
+        # the targets have been extracted here manually by either running kotlinc-native -list-targets for the relevant distribution
+        # or listing entries under <konan.home>/targets/, and then map Bazel platforms for it so that we can create the relevant toolchains
+        # with the right target_compatible_with
+        targets = {
+            ("@platforms//os:android", "@platforms//cpu:armv7"): ["android_arm32"],
```
smocherla-brex, line 80: This is a mapping I added for all the targets for each of  compiler distributions and then map to Bazel platforms used in `target_compatible_with` in the respective toolchain. (open to better ideas on doing it)

kotlin/internal/klib/klib.bzl
smocherla-brex: Updated to `kt_library`

kotlin/compiler/compiler.bzl
smocherla-brex: I've tried moving this to a dedicated CLI toolchain but unfortunately we need the runfiles path to `kotlin-native.jar` as a jvm flag to the builder `java_binary` target [here](https://github.com/bazelbuild/rules_kotlin/blob/master/src/main/kotlin/io/bazel/kotlin/builder/cmd/BUILD.bazel#L48) and it doesn't seem to be possible to provide it through the toolchain (since it uses `java_binary` with `jvm_flags`). The builder compilation still uses a dedicated toolchain (`kotlin_native_cli_toolchain`) and the jar from there but I've not found a way to re-use that with the binary runfiles.

src/test/starlark/internal/klib/klib_tests.bzl
smocherla-brex: Update rule to include runfiles and added starlark tests for them.

PR #1354, opened 2025-07-18
[kapt] allow options to be passed to kapt and javac annotation processing. 

restingbull:
Implements `kapt_compiler_plugin`, which allows the use of `kt_plugin_cfg` to pass option to annotation processors.

This is step one of moving kapt out of the builder, and enhancing the `kt_compiler_plugin` support.

## Changes:
* `resolve_cfg` signature change (no one is using it, as it's limited and complicated) to return a list of providers. This allows a compiler plugin to export configurations in a clearer manner.
* Add `annotation_processor_options` to  `JavacOptions`. This is *not* exposed on `kt_javac_opts`, as `annotation_processor_options` should always be associated with an annotation processor -- to avoid orphaned arguments being passed indefinitely.  A rule could exploit this to make a mess, but that is on the author. I hope to avoid regretting this feature.
* modify `base64` args encoding to optimize on single argument values. The whole protocol for this in kapt is a right mess. 
* Fix javac annotation processing to include exported java plugins. Need better tests on this.
* And a heavy bunch of test code. Some refactoring could make it less verbose.

## Example:
This will add the undocumented argument `com.google.auto.value.OmitIdentifiers` to AutoValue processing. `:autovalue_no_identifiers` can be added to `kt_jvm_library.plugins` to trigger annotation processing.
```
kt_plugin_cfg(
    name = "autovalue_no_identifiers",
    options = {
        "com.google.auto.value.OmitIdentifiers": ["true"], # declare the OmitIdentifiers option value.
    },
    plugin = "//kotlin/compiler:kapt", # targets the configuration to kapt
    deps = [
        ":autovalue", # adds the AutoValue annotation processor to javac and kotlinc kapt
    ],
)
```


restingbull:
fyi: @shs96c, @cheister

Issue #1362, opened 2025-09-12
Support adding "file" dependencies to kt_compiler_plugin or kt_plugin_cfg


rbeazleyspot:
As mentioned [here](https://bazelbuild.slack.com/archives/CD9FZL3KR/p1757604857114379) it looks like its impossible to add files as dependencies to compiler plugins.

My usecase is to supply a stability configuration file to the jetpack compose compiler plugin.

PR #1363, opened 2025-09-16
Add DefaultInfo provider to kt_compiler_plugin and kt_plugin_cfg deps list

rbeazleyspot:
This is to enable #1362.

With this change in place you can do something like

```
kt_compiler_plugin(
    name = "jetpack_compose_compiler_plugin",
    id = "androidx.compose.compiler.plugins.kotlin",
    options = {
        "sourceInformation": "true",  # Required for AS Layout Inspector, disable for release builds
        "stabilityConfigurationPath": "$(locations :stability_config)",
    },
    target_embedded_compiler = True,
    deps = [
        ":stability_config",
        "@maven_rules_kotlin_example//:org_jetbrains_kotlin_kotlin_compose_compiler_plugin_embeddable",
    ],
)

filegroup(
    name = "stability_config",
    srcs = ["stability.conf"],
)
```

I tested this by modifying `examples/jetpack_compose` //app:compose_example_app with a locally built version of the rules (the example code above was lifted from that)

### DIFF
--- a/kotlin/internal/jvm/jvm.bzl
+++ b/kotlin/internal/jvm/jvm.bzl
@@ -537,7 +537,7 @@ kt_jvm_library(
     attrs = {
         "deps": attr.label_list(
             doc = "The list of libraries to be added to the compiler's plugin classpath",
-            providers = [JavaInfo],
+            providers = [[JavaInfo], [DefaultInfo]],
             cfg = "exec",
             aspects = [_kt_compiler_deps_aspect],
         ),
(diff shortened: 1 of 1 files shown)


rbeazleyspot:
@restingbull WDYT ? think it aligns with your current changes?

restingbull:
I meant to follow the convention for `data`? `deps` is generally reserved for known types, rather than arbitrary files.

Though, we'll need to pull `DefaultInfo` off the deps as well. The `data` attribute has a bit of a cursed history.

PR #1364, opened 2025-09-17
Add data attr to kt_plugin_cfg and kt_compiler_plugin

rbeazleyspot:
Alternative for #1363 using data attribute

This adds a data attribute to kt_plugin_cfg and kt_compiler_plugin

ctx.expand_location is curried with the targets listed in data

tested [here](https://github.com/rbeazleyspot/rules_kotlin/commit/43aa72413044cdb298f18a86a961348811ac6119) using bazelisk build //app:compose_example_app


rbeazleyspot:
if this looks good I can make the PR mergable

PR #1371, opened 2025-10-02
ktlint_fix: /bin/bash -> /usr/bin/env bash

dmivankov:
/usr/bin/env is more portable, can take binary from PATH

Upstream https://github.com/bazelbuild/rules_java/blob/master/java/bazel/rules/java_stub_template.txt has adopted it too


kotlin/internal/lint/ktlint_fix.bzl
```diff
@@ -15,7 +15,7 @@ def _ktlint_fix_impl(ctx):
 
     # Much of the following is lifted from:
     # https://cs.opensource.google/bazel/bazel/+/refs/tags/4.0.0:src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt;l=114
```
dmivankov:
```suggestion
    # https://github.com/bazelbuild/rules_java/blob/master/java/bazel/rules/java_stub_template.txt
```
might be good to apply too

Bencodes (approved):

Bencodes (approved):

Issue #1372, opened 2025-10-02
ktlint_fix doesn't use java toolchain to run its binary


dmivankov:
If `java` isn't available on PATH or default search locations, running ktlint_fix target will fail
```console
$ bazel run --run_under "sh -c" sometarget_ktlint_fix
...
+ external/rules_kotlin++rules_kotlin_extensions+com_github_pinterest_ktlint/file/downloaded --format --editorconfig=.editorconfig --relative ...somepaths...
external/rules_kotlin++rules_kotlin_extensions+com_github_pinterest_ktlint/file/downloaded: line 10: [: : integer expected
external/rules_kotlin++rules_kotlin_extensions+com_github_pinterest_ktlint/file/downloaded: line 12: exec: java: not found
```
locating the script
```console
$ find bazel-bin/ | grep external/rules_kotlin++rules_kotlin_extensions+com_github_pinterest_ktlint/file/downloaded
```
and then running it
```console
$ sh -x bazel-bin/..sometarget_..ktlint_fix-lint-fix.runfiles/_main/external/rules_kotlin++rules_kotlin_extensions+com_github_pinterest_ktlint/file/downloaded
++ java -version
++ sed -E -n 's/.* version "([^.-]*).*".*/\1/p'
+ JV=
++ '[' '' -ge 16 ']'
bazel-bin/..sometarget.._ktlint_fix-lint-fix.runfiles/_main/external/rules_kotlin++rules_kotlin_extensions+com_github_pinterest_ktlint/file/downloaded: line 10: [: : integer expected
++ echo ''
+ X=
+ exec java -Xmx512m -jar bazel-bin/..somepath.._ktlint_fix-lint-fix.runfiles/_main/external/rules_kotlin++rules_kotlin_extensions+com_github_pinterest_ktlint/file/downloaded
bazel-bin/..somepath.._ktlint_fix-lint-fix.runfiles/_main/external/rules_kotlin++rules_kotlin_extensions+com_github_pinterest_ktlint/file/downloaded: line 12: exec: java: not found
```

dmivankov:
@restingbull linked PR doesn't touch `java` setup so the issue should still be open

PR #1375, opened 2025-10-08
ktlint_fix: use `java` from toolchain rather than from default PATH

dmivankov:
Should fix https://github.com/bazelbuild/rules_kotlin/issues/1372

### DIFF
--- a/kotlin/internal/lint/ktlint_fix.bzl
+++ b/kotlin/internal/lint/ktlint_fix.bzl
@@ -92,11 +92,12 @@ fi
 SRCS=({srcs})
 SRCS=${{SRCS[@]/#/$BUILD_DIR}}
 
-"$TOOL" {args} $SRCS
+PATH=\"$(dirname "{binjava}"):$PATH\" "$TOOL" {args} $SRCS
 """.format(
         executable = ctx.executable._ktlint_tool.path,
         args = " ".join(args),
         srcs = " ".join([src.path for src in ctx.files.srcs]),
+        binjava = ctx.attr._java[java_common.JavaRuntimeInfo].java_executable_runfiles_path,
     )
 
     content = ctx.expand_location(content, [ctx.attr._ktlint_tool])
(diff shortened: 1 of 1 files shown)


kotlin/internal/lint/ktlint_fix.bzl
```diff
@@ -140,6 +141,9 @@ ktlint_fix = rule(
             executable = True,
             cfg = "target",
         ),
+        "_java": attr.label(
+            default = "@bazel_tools//tools/jdk:current_java_runtime",
```
restingbull: Should be pulled from the java runtime toolchain -- `@bazel_tools//tools/jdk:runtime_toolchain_type`, rather than an attribute.

This is to ensure that the binary will respect platform during compilation. With RBE, it's not unusual to build binaries for a `target` platform on an incompatible `exec`. e.g. it ain't always `host`. :)

Otherwise, looks good.

kotlin/internal/lint/ktlint_fix.bzl
```diff
@@ -97,7 +97,7 @@ PATH=\"$(dirname "{binjava}"):$PATH\" "$TOOL" {args} $SRCS
         executable = ctx.executable._ktlint_tool.path,
         args = " ".join(args),
         srcs = " ".join([src.path for src in ctx.files.srcs]),
-        binjava = ctx.attr._java[java_common.JavaRuntimeInfo].java_executable_runfiles_path,
+        binjava = ctx.toolchains["@bazel_tools//tools/jdk:runtime_toolchain_type"].java_runtime.java_executable_runfiles_path,
```
Bencodes: There are constants for these https://github.com/bazelbuild/rules_kotlin/blob/45d4446888a332375de0c073a8c0696668093b18/src/main/starlark/core/compile/common.bzl#L5C39-L5C82

kotlin/internal/lint/ktlint_fix.bzl
```diff
@@ -141,6 +144,9 @@ ktlint_fix = rule(
             cfg = "target",
         ),
     },
+    toolchains = [
+        "@bazel_tools//tools/jdk:runtime_toolchain_type",
```
restingbull: JAVA_RUNTIME_TOOLCHAIN_TYPE

dmivankov:
```suggestion
        JAVA_RUNTIME_TOOLCHAIN_TYPE,
```

PR #1378, opened 2025-10-21
Add env and env_inherit support to kt_jvm_binary

### LINKED ISSUE #1359: Allow passing in env variables into kt_jvm_binary
Similar to `java_binary` and `kt_jvm_test` (see https://github.com/bazelbuild/rules_kotlin/issues/815 fixed in github.com/bazelbuild/rules_kotlin/pull/837), `kt_jvm_binary` should also support env variables.

As mentioned by the docs,
```
**Note:** This rule does not have all of the features found in [`java_binary`](https://docs.bazel.build/versions/master/be/java.html#java_binary).
It is appropriate for building workspace utilities. `java_binary` should be preferred for release artefacts.
```
to be appropriate for building workspace utilities, it makes sense to support env variables (for example, setting runfiles locations).

agluszak:
This adds parity with java_binary and kt_jvm_test by allowing kt_jvm_binary to specify environment variables when executed with bazel run.

Two new attributes are added:
- env: Dictionary of environment variables to set
- env_inherit: List of environment variable names to inherit from the shell environment

The implementation uses RunEnvironmentInfo provider, following the same pattern as the existing kt_jvm_test support (PR #837).

Fixes #1359

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,11 @@ Also, `kt_jvm_*` rules support the following standard `java_*` rules attributes:
   * `resources`
   * `resources_strip_prefix`
   * `exports`
-  
+
+Additionally, `kt_jvm_binary` and `kt_jvm_test` support environment variable configuration:
+  * `env` - Dictionary of environment variables to set when the target is executed with `bazel run` or `bazel test`
+  * `env_inherit` - List of environment variable names to inherit from the shell environment
+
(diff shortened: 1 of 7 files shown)


Bencodes:
@agluszak did you mean to close this?

Bencodes:
https://github.com/bazelbuild/rules_kotlin/pull/1379 ah just saw that this got opened up again as a different PR.

PR #1380, opened 2025-10-21
kt_jvm_binary: resolve JVM via runfiles path (fixes #1332)

### LINKED ISSUE #1332: java not found for kt_jvm_binary targets executed from a rule
I have a `kt_jvm_binary` [target](https://github.com/tharakadesilva/rules_kotlin_1332/blob/main/downloader/BUILD.bazel#L9-L14) defined as follows:

```
kt_jvm_binary(
    name = "DownloaderApplication",
    main_class = "myapp.DownloaderKt",
    visibility = ["//:__subpackages__"],
    runtime_deps = [":downloader"],
)
```

When I run this as `bazel run //downloader:DownloaderApplication asd asd`, things are working as expected.

However, when I try to run this from a [rule](https://github.com/tharakadesilva/rules_kotlin_1332/blob/main/rules/download.bzl#L21-L25), I run into:

```
ERROR: /Users/tharakadesilva/git_tree/tharakadesilva/rules_kotlin_1332/BUILD.bazel:3:14: Downloading data failed: (Exit 126): DownloaderApplication failed: error executing DownloadData command (from target //:download_data) bazel-out/darwin_arm64-opt-exec-ST-d57f47055a04/bin/downloader/DownloaderApplication --output bazel-out/darwin_arm64-fastbuild/bin/output.csv some args

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
bazel-out/darwin_arm64-opt-exec-ST-d57f47055a04/bin/downloader/DownloaderApplication: line 412: /private/var/tmp/_bazel_tharakadesilva/b372dc11837e8cf37fe5cff4ecc060ad/sandbox/darwin-sandbox/14/execroot/_main/../rules_java++toolchains+remotejdk11_macos_aarch64/bin/java: No such file or directory
bazel-out/darwin_arm64-opt-exec-ST-d57f47055a04/bin/downloader/DownloaderApplication: line 412: exec: /private/var/tmp/_bazel_tharakadesilva/b372dc11837e8cf37fe5cff4ecc060ad/sandbox/darwin-sandbox/14/execroot/_main/../rules_java++toolchains+remotejdk11_macos_aarch64/bin/java: cannot execute: No such file or directory
Target //:download_data failed to build
```

---

## Steps to Reproduce

Full repro: https://github.com/tharakadesilva/rules_kotlin_1332/tree/main

Run: `bazel build :download_data`

---

## Workaround

Seems to be related to: https://github.com/bazelbuild/rules_kotlin/pull/1242

Using the workaround [here](https://github.com/bazelbuild/rules_kotlin/issues/778#issuecomment-2177994390) (using `java_binary`) works.

agluszak:
Use JavaRuntimeInfo.java_executable_runfiles_path (already in use)
and normalize it like rules_java does: prepend workspace name for
relative paths, and use ${JAVA_RUNFILES}/ prefix in the stub template.

This makes kt_jvm_binary work when executed as a tool inside another
action and under Bazel 8's default --nolegacy_external_runfiles.

The fix follows the same approach as rules_java:
- Normalize java_bin_path by prepending workspace name for relative paths
- Use ${JAVA_RUNFILES}/ prefix which the template resolves at runtime
- Keep third_party/java_stub_template.txt unchanged

Note: rules_java uses ctx.configuration.runfiles_enabled() to check
whether runfiles are enabled, but this is a private API that third-party
rules cannot access. Our implementation uses ${JAVA_RUNFILES}/ prefix
unconditionally, which works in both runfiles-enabled and manifest-only
modes since the stub template handles both cases.

Add an integration test that runs a kt_jvm_binary from a rule action
to verify the fix works correctly.

Fixes #1332

### DIFF
--- a/kotlin/internal/jvm/impl.bzl
+++ b/kotlin/internal/jvm/impl.bzl
@@ -71,6 +71,15 @@ def _write_launcher_action(ctx, rjars, main_class, jvm_flags):
     java_runtime = ctx.toolchains["@bazel_tools//tools/jdk:runtime_toolchain_type"].java_runtime
     java_bin_path = java_runtime.java_executable_runfiles_path
 
+    # Normalize java_bin_path like rules_java does: prepend workspace name for relative paths
+    if not _is_absolute(java_bin_path):
+        java_bin_path = ctx.workspace_name + "/" + java_bin_path
+
+    # Construct JAVABIN substitution with ${JAVA_RUNFILES}/ prefix for relative paths
+    # This works in both runfiles-enabled and manifest-only modes
+    prefix = "" if _is_absolute(java_bin_path) else "${JAVA_RUNFILES}/"
+    java_bin = "JAVABIN=${JAVABIN:-" + prefix + java_bin_path + "}"
+
     # Following https://github.com/bazelbuild/bazel/blob/6d5b084025a26f2f6d5041f7a9e8d302c590bc80/src/main/starlark/builtins_bzl/bazel/java/bazel_java_binary.bzl#L66-L67
     # Enable the security manager past deprecation until permanently disabled: https://openjdk.org/jeps/486
     # On bazel 6, this check isn't possible...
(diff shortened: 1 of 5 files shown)


third_party/java_stub_template.txt
```diff
@@ -244,6 +244,20 @@ fi
 # Set JAVABIN to the path to the JVM launcher.
 %javabin%
 
+# Resolve JAVABIN via runfiles if it's a relative path.
```
Bencodes: Isn't this a deviation from the one that Bazel provides?

https://github.com/bazelbuild/rules_java/blob/master/java/bazel/rules/java_stub_template.txt

PR #1358, opened 2025-08-21
Update rules_android to fix --incompatible_disable_native_repo_rules failure

restingbull:
This will pull in https://github.com/bazelbuild/rules_android/pull/376 and fix part of the [release failures](https://buildkite.com/bazel/bcr-presubmit/builds/18423/steps/canvas?sid=0198cdb5-4b4a-459a-9762-9bcb94947930)

PR #1381, opened 2025-10-22
Remove WORKSPACE.bzlmod

agluszak:
As of rules_android 0.6.6 it android_sdk_repository is no longer needed. See https://github.com/bazelbuild/rules_android/issues/144


examples/android/WORKSPACE
```diff
@@ -1,121 +0,0 @@
-workspace(name = "android_example")
```
Bencodes, line 1: Why remove this? We still have to support WORKSPACE in this project and have these in place to test against that behavior.

agluszak:
https://github.com/bazelbuild/rules_kotlin/pull/1358 includes this

PR #1382, opened 2025-10-22
Bump bazel version to 8.4.2

agluszak:


restingbull:
@Bencodes / @nkoroste -- we good with dropping testing on bazel 7?

examples/multiplex/WORKSPACE
```diff
@@ -1,44 +0,0 @@
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
```
Bencodes, line 1: Why did this get dropped?

Bencodes:
@restingbull is the idea to drop support for 7.x completely here? If so then the README likely needs to be updated along with this PR.

agluszak:
Restored and fixed that WORKSPACE file in the `multiplex` example

.bazelci/presubmit.yml
```diff
@@ -13,9 +13,9 @@ matrix:
     # - macos
     # - windows re-enable when rules_bazel_integration_test can support custom test runner on windows.
   test_flags:
-    - ["--enable_bzlmod=true"]
```
Bencodes, line 16: What was the reasoning behind removing this?

Bencodes (approved):

.bazelci/presubmit.yml
Bencodes: Hmm it looks like even no 7.6.1 it was enabled by default https://github.com/bazelbuild/bazel/blob/release-7.6.1/src/main/java/com/google/devtools/build/lib/packages/semantics/BuildLanguageOptions.java#L256

PR #1384, opened 2025-10-22
Bump some bazel deps

agluszak:


agluszak:
> What's the impetus?

The fact that the newer versions might get pulled transitively by other deps anyway. And to make sure that they don't break rules_kotlin

agluszak:
+ Bazel 9 support :)

agluszak:
Got hit by https://github.com/bazelbuild/rules_android/issues/410 (the problem is not within rules_android itself tho)

PR #1386, opened 2025-10-23
Windows portability fixes

FaBrand:
I would like to add some compatibility changes i had before as patches that should be  (mostly) compatible

One that change breaks compatibility with bazel version < 7.1.0 (the `get_auth` method from `bazel_tools` got introduced with 7.1.0)

Confined the buildifier formatting changes in one commit for reviewability

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -17,7 +17,7 @@ bazel_dep(name = "rules_android", version = "0.6.4")
 bazel_dep(name = "bazel_features", version = "1.25.0")
 bazel_dep(name = "rules_shell", version = "0.4.1")
 
-bazel_dep(name = "buildifier_prebuilt", version = "8.0.3", dev_dependency = True)
+bazel_dep(name = "buildifier_prebuilt", version = "8.2.0.2", dev_dependency = True)
 
 rules_java_toolchains = use_extension("@rules_java//java:extensions.bzl", "toolchains")
 use_repo(rules_java_toolchains, "remote_java_tools")


kotlin/internal/jvm/compile.bzl
```diff
@@ -433,8 +433,8 @@ def _run_ksp_builder_actions(
         transitive_runtime_jars = transitive_runtime_jars,
         plugins = plugins,
         outputs = {
-            "ksp_generated_java_srcjar": ksp_generated_java_srcjar,
             "ksp_generated_classes_jar": ksp_generated_classes_jar,
+            "ksp_generated_java_srcjar": ksp_generated_java_srcjar,
```
restingbull, line 437: What's happening here? Looks like a sort or something?

FaBrand: I had buildifier set-up to automatically fix the [unsorted-dict-items](https://github.com/bazelbuild/buildtools/blob/main/WARNINGS.md#unsorted-dict-items). I decided to keep it as a simple fix, but bundled in the first of the three commits.

FaBrand: Filtering the "auto-format" commit makes it a bit easier to review:
https://github.com/bazelbuild/rules_kotlin/pull/1386/files/24ca76eb315256ad964d007cd365e997ae226fa8..365f45137b7fb28c353fc2beba946d82364f4f07

restingbull: Could you include an update to [the buildifier rules](https://github.com/bazelbuild/rules_kotlin/blob/master/BUILD#L86-L121), since we'll need to keep this in order?

FaBrand: @restingbull To confirm since i'm not 100% sure which of the two you mean?

a) Include the `unsorted-dict-items` in the buildifier.{check,fix} targets so everything gets sorted
b) Revert the auto-sort commit and add the warning to the excluded ones

I pushed 2 more commits.

One that [adds the option to the buildifier rule](https://github.com/bazelbuild/rules_kotlin/pull/1386/commits/e296d4f990f45ebceef517152929e762b3646eb3) and [one that fixes a deprecation error](https://github.com/bazelbuild/rules_kotlin/pull/1386/commits/34796dc46e728546a4bb4c16795a560f5695d419) i found along the way

PR #1387, opened 2025-10-24
Add support for Kotlin 2.2+

### LINKED ISSUE #1373: Kotlin 2.2 compiler release isn't yet supported
https://github.com/JetBrains/kotlin/commit/951bf8eb0323d95bd3b0a1aae43776b934a8f876
removed trove4j jar but it is still directly referenced by rules_kotlin and will fail on missing jar

Continuation of https://github.com/bazelbuild/rules_kotlin/issues/1365 and https://github.com/bazelbuild/rules_kotlin/pull/1367

agluszak:
Kotlin 2.2 removed trove4j and android-extensions jars from the distribution and changed FIR method signatures.

Fixes #1373

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,7 +6,7 @@ module(
 )
 
 bazel_dep(name = "protobuf", version = "29.0", repo_name = "com_google_protobuf")
-bazel_dep(name = "rules_proto", version = "7.1.0", repo_name = "rules_proto")
+bazel_dep(name = "rules_proto", version = "6.0.2", repo_name = "rules_proto")
 bazel_dep(name = "abseil-py", version = "2.1.0", repo_name = "py_absl")
 bazel_dep(name = "rules_cc", version = "0.0.16")
 bazel_dep(name = "platforms", version = "0.0.11")
--- a/MODULE.release.bazel
+++ b/MODULE.release.bazel
@@ -11,7 +11,7 @@ bazel_dep(name = "rules_java", version = "8.9.0")
 bazel_dep(name = "rules_android", version = "0.6.4")
 bazel_dep(name = "bazel_features", version = "1.25.0")
 bazel_dep(name = "protobuf", version = "29.0", repo_name = "com_google_protobuf")
-bazel_dep(name = "rules_proto", version = "7.1.0", repo_name = "rules_proto")
+bazel_dep(name = "rules_proto", version = "6.0.2", repo_name = "rules_proto")
 bazel_dep(name = "abseil-py", version = "2.1.0", repo_name = "py_absl")
 bazel_dep(name = "rules_cc", version = "0.0.16")
 
--- a/src/main/starlark/core/repositories/versions.bzl
+++ b/src/main/starlark/core/repositories/versions.bzl
@@ -59,8 +59,8 @@ versions = struct(
         ],
     ),
     RULES_PROTO = version(
-        version = "7.1.0",
-        sha256 = "14a225870ab4e91869652cfd69ef2028277fc1dc4910d65d353b62d6e0ae21f4",
+        version = "7.0.2",
+        sha256 = "0e5c64a2599a6e26c6a03d6162242d231ecc0de219534c38cb4402171def21e8",
         strip_prefix_template = "rules_proto-{version}",
         url_templates = [
             "https://github.com/bazelbuild/rules_proto/releases/download/{version}/rules_proto-{version}.tar.gz",
@@ -88,11 +88,11 @@ versions = struct(
         sha256 = "6550f1117d7c9590cc9a5075b92682a218c8e1df4093d7e683d73cc481733dd1",
     ),
     KOTLIN_BUILD_TOOLS_IMPL = version(
-        version = "2.1.20",
+        version = "2.2.21",
         url_templates = [
             "https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-build-tools-impl/{version}/kotlin-build-tools-impl-{version}.jar",
         ],
-        sha256 = "6e94896e321603e3bfe89fef02478e44d1d64a3d25d49d0694892ffc01c60acf",
+        sha256 = "458b63ef8fc50a0a3180fe688ec1b995745992a05c495dac61e58d1088927a80",
     ),
     RULES_ANDROID = version(
         version = "0.6.4",


restingbull:
I think something is slightly off with the [stdlib](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/019a1652-30b5-46c3-8c41-2e58324a07d9/examples/custom_compiler_version_8.2.1_test/test_attempts/attempt_1.log):
```
java.lang.NoSuchMethodError: 'kotlin.sequences.Sequence kotlin.sequences.SequencesKt.sequenceOf(java.lang.Object)'
```

and possible the [plugin](https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/019a1652-30b5-46c3-8c41-2e58324a07d9/examples/custom_compiler_version_8.2.1_test/test_attempts/attempt_1.log):
```
 java.lang.AbstractMethodError: Missing implementation of resolved method 'abstract void check(org.jetbrains.kotlin.fir.analysis.checkers.context.CheckerContext, org.jetbrains.kotlin.diagnostics.DiagnosticReporter, org.jetbrains.kotlin.fir.declarations.FirDeclaration)' of abstract class org.jetbrains.kotlin.fir.analysis.checkers.declaration.FirDeclarationChecker.
	at org.jetbrains.kotlin.fir.analysis.checkers.declaration.DeclarationCheckersDiagnosticComponent.visitConstructor(DeclarationCheckersDiagnosticComponent.kt:302)
```

How about we break this up into versioning artifacts and updating to 2.2+? I can't quite figure out which is breaking the integration tests.

agluszak:
So it looks like it's about a stdlib version mismatch. `kotlin.sequences.SequencesKt.sequenceOf(java.lang.Object)` was only introduced in 2.2

agluszak:
So I'm not sure if this versioning thingy is going to work because of bootstrapping :/

src/main/kotlin/io/bazel/kotlin/generate/WriteKotlinArtifacts.kt
```diff
@@ -0,0 +1,270 @@
+package io.bazel.kotlin.generate
```
Bencodes: There is a world where we dynamically run this inside of the repo creation and not require that it be written as a standalone tool that has to be managed independently.

agluszak: Hey, I pinged you on the Bazel slack

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/KotlinJvmTaskExecutor.kt
```diff
@@ -80,6 +80,9 @@ class KotlinJvmTaskExecutor
                             inputs.directDependenciesList.forEach {
                               flag("direct_dependencies", it)
                             }
+                            inputs.classpathList.forEach {
+                              flag("full_classpath", it)
```
Bencodes, line 84: What's the reasoning behind this change?

src/main/kotlin/io/bazel/kotlin/plugin/jdeps/BaseJdepsGenExtension.kt
```diff
@@ -16,23 +16,69 @@ abstract class BaseJdepsGenExtension(
...
     doStrictDeps(configuration, targetLabel, directDeps, explicitDeps)
   }
 
+  /**
+   * Creates a mapping from canonical filesystem paths to original classpath paths.
+   * This ensures jdeps uses the same paths that were on the compiler's classpath.
+   * When multiple classpath entries point to the same canonical file, prefer runfiles paths.
+   */
+  private fun createCanonicalToClasspathMap(classpathEntries: List<String>): Map<String, String> {
+    val mapping = mutableMapOf<String, String>()
```
Bencodes, line 43: Are these jdeps changes related to this supporting Kotlin 2.2 or is this fixing something independent?

src/main/kotlin/io/bazel/kotlin/plugin/jdeps/BUILD.bazel
```diff
@@ -25,6 +25,9 @@ kt_bootstrap_library(
         "k2/checker/declaration/*.kt",
         "k2/checker/expression/*.kt",
     ]),
+    kotlinc_opts = {
```
Bencodes: Should we just convert this to an array type so that we aren't messing around with dict key,values like this?

src/main/kotlin/io/bazel/kotlin/plugin/jdeps/BaseJdepsGenExtension.kt
agluszak: Supporting kotlin 2.2. I don't understand what exactly changed on kotlinc's side, but without these changes the tests fail

src/main/kotlin/io/bazel/kotlin/plugin/jdeps/BUILD.bazel
agluszak: I don't have any strong opinion about that

Bencodes: Let's just make this an array so that we aren't having to special case things inside the impl.

agluszak: I think I misunderstood what you meant - it's not about making turning kotlinc_opts into an array _in general_, right?

restingbull: I believe that is what @Bencodes means. The CLI tooling is meant to be very hands off, and if we have flags without values, a dict is no longer a good representation of it.

agluszak:
https://github.com/bazelbuild/rules_kotlin/pull/1393 will fix integration tests failing, but I need to investigate what happened to RBE

Bencodes (approved):

src/main/kotlin/io/bazel/kotlin/plugin/jdeps/k2/checker/declaration/BasicDeclarationChecker.kt
```diff
@@ -11,11 +11,8 @@ import org.jetbrains.kotlin.fir.declarations.toAnnotationClassLikeSymbol
 internal class BasicDeclarationChecker(
   private val classUsageRecorder: ClassUsageRecorder,
 ) : FirBasicDeclarationChecker(MppCheckerKind.Common) {
-  override fun check(
-    declaration: FirDeclaration,
-    context: CheckerContext,
-    reporter: DiagnosticReporter,
-  ) {
+  context(context: CheckerContext, reporter: DiagnosticReporter)
+  override fun check(declaration: FirDeclaration) {
```
restingbull, line 15: We'll need to call this out in the release notes.

PR #1389, opened 2025-10-30
 Update `resource_strip_prefix` logic to support path handling across modules

agluszak:
Update `resource_strip_prefix` logic to support path handling across modules

Add an example with a nested module structure

### DIFF
--- a/src/test/starlark/BUILD.bazel
+++ b/src/test/starlark/BUILD.bazel
@@ -1,6 +1,6 @@
-load("//src/test/starlark/internal/jvm:resource_strip_prefix_test.bzl", "myrules_test_suite")
+load("//src/test/starlark/internal/jvm:resource_strip_prefix_test.bzl", "strip_resource_prefix_test_suite")
 load(":convert_test.bzl", "convert_test_suite")
 
 convert_test_suite(name = "convert_tests")
 
-myrules_test_suite(name = "resource_strip_prefix_tests")
+strip_resource_prefix_test_suite(name = "resource_strip_prefix_tests")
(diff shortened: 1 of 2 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -274,8 +278,23 @@ def _fold_jars_action(ctx, rule_kind, toolchains, output_jar, input_jars, action
 
 def _resourcejar_args_action(ctx, extra_resources = {}):
     res_cmd = []
+
+    # Get the strip prefix from the File object if provided
+    strip_prefix = None
```
Bencodes, line 279: https://github.com/bazelbuild/bazel/blob/fc2239dba7db6551439d4f1f96be8433d35fa2a1/src/main/java/com/google/devtools/build/lib/rules/java/JavaHelper.java#L97-L113

Bazel impl. for reference

kotlin/internal/jvm/compile.bzl
```diff
@@ -274,8 +278,23 @@ def _fold_jars_action(ctx, rule_kind, toolchains, output_jar, input_jars, action
...
+        # Use the full path directly - it could be a directory or a file
+        file = ctx.file.resource_strip_prefix
+        file_path = file.path
+
+        # if dirname starts with ctx.label.package, we need to remote ctx.label.package, because it means that
+        # we've hit the edge case when there is a target with the same name as the package
+        if file.dirname.startswith(ctx.label.package + "/"):
+            strip_prefix = file_path[len(ctx.label.package) + 1:]
+        else:
+            strip_prefix = file_path
```
Bencodes, line 294: Cases like these we might be better off writing unit tests instead of building out an entire integration test inside the examples directory, since running those integration tests is extremely slow.

agluszak: Actually, the integration test is testing the "happy" path of this `if` - it's a repro of the case we've been running into internally at JB. I'm not sure if it's possible to easily write an equivalent unit test, because it depends on bazel's internal path resolution for modules.

But regarding this specific if (i.e. the case when there's a //resources:resources filegroup target) - sure, I can write a unit test

PR #1390, opened 2025-10-30
Update Java runtime version to remotejdk_17 in .bazelrc

agluszak:
Makes it easier to run tests locally

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -7,9 +7,9 @@ matrix:
     - ubuntu2404
     - ubuntu2204
   integration_shard_flags:
-    - ["--test_tag_filters=shard_0", "--config=rbe"]
-    - ["--test_tag_filters=shard_1", "--config=rbe"]
-    - ["--test_tag_filters=shard_2", "--config=rbe"]
+    - ["--test_tag_filters=shard_0"]
+    - ["--test_tag_filters=shard_1"]
+    - ["--test_tag_filters=shard_2"]
     # - macos
     # - windows re-enable when rules_bazel_integration_test can support custom test runner on windows.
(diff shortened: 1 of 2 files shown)


.bazelci/presubmit.yml
```diff
@@ -7,9 +7,9 @@ matrix:
     - ubuntu2404
     - ubuntu2204
   integration_shard_flags:
-    - ["--test_tag_filters=shard_0", "--config=rbe"]
-    - ["--test_tag_filters=shard_1", "--config=rbe"]
-    - ["--test_tag_filters=shard_2", "--config=rbe"]
+    - ["--test_tag_filters=shard_0"]
```
Bencodes, line 10: I believe we want this so that we are exercising RBE as well, since that's something that we need to support.

agluszak: I don't think the `--config=rbe` flag is needed for actual RBE to work. In fact, I removed it, because otherwise the RBE test was failing with something along the lines of "no config named rbe found" (which is true, because I removed the `:rbe` part from .bazelrc).

Bencodes: Is it somehow related to this? https://github.com/bazelbuild/rules_kotlin/blob/9734aa6f309bd1216f332910f958f17fdd22f8c8/MODULE.bazel#L49

agluszak: https://bazel.build/versions/8.4.0/remote/ci doesn't say anything about the rbe config.

And our RBE job runs just fine without --config=rbe - https://buildkite.com/bazel/rules-kotlin-kotlin/builds/5446/steps/canvas?sid=019a3523-31c1-41b9-98db-9e5fa9fd2a08

Issue #1275, opened 2025-02-13
Add KSP2 support


jbarr21:
KSP2 is no longer a compiler plugin & has its own entrypoint
https://github.com/google/ksp/blob/main/docs/ksp2.md#using-ksp-2-from-command-line-and-in-program

jbarr21:
as of KSP [v2.1.20-2.0.0](https://github.com/google/ksp/releases/tag/2.1.20-2.0.0), KSP2 is now the default version (using Kotlin 2.x APIs)

PR #1394, opened 2025-11-08
Draft: Starlark kapt & KSP2

### LINKED ISSUE #1311: Move kapt to use kt_compiler_plugin
Unblocks using kt_plugin_cfg to pass plugin options.

Less difficult than moving KSP.

agluszak:
Closes #1311, #1275


kotlin/internal/jvm/compile.bzl
```diff
@@ -359,6 +380,198 @@ def _run_merge_jdeps_action(ctx, toolchains, jdeps, outputs, deps):
...
+
+    # Package generated classes and resources into a jar
+    # This includes:
+    # - Classes generated directly by KSP processors (not compiled from sources)
+    # - Resources like META-INF/services, META-INF/proguard, etc.
+    generated_class_jar = ctx.actions.declare_file(
+        ctx.label.name + "-ksp2-generated-classes.jar",
+    )
+
+    # Merge class_out_dir and resource_out_dir into one jar
```
restingbull: Instead of `run_shell`, move the execution logic to a worker binary. 

Amoung other things, it makes unittesting easier -- and removes an action. They add up.

kotlin/internal/jvm/compile.bzl
```diff
@@ -359,6 +380,198 @@ def _run_merge_jdeps_action(ctx, toolchains, jdeps, outputs, deps):
...
+        transitive_runtime_jars: Runtime JARs
+
+    Returns:
+        struct with:
+            kotlin_srcjar: Generated Kotlin sources
+            java_srcjar: Generated Java sources
+    """
+
+    # Declare output directories
+    kotlin_out_dir = ctx.actions.declare_directory(
```
restingbull, line 427: What is the advantage of all these directories as outputs?

I ask because `delcare_directory` is not only inefficient, but is only useful in very specific circumstances -- when you do not know the output of an action, but need to get specific artifacts out of it.

kotlin/internal/jvm/compile.bzl
```diff
@@ -359,6 +380,198 @@ def _run_merge_jdeps_action(ctx, toolchains, jdeps, outputs, deps):
...
+    )
+    caches_dir = ctx.actions.declare_directory(
+        ctx.label.name + "-ksp2-caches",
+    )
+
+    # Collect processor JARs
+    # annotation_processors is a DEPSET (not list!), each element has processor_jars (depset)
+    # Must convert to list first before iterating
+    processor_jars_depsets = [p.processor_jars for p in annotation_processors.to_list()]
+    processor_jars = depset(transitive = processor_jars_depsets).to_list()
```
restingbull: Never call `to_list` on a depset. It explodes the starlark memory usage.

kotlin/internal/jvm/compile.bzl
```diff
@@ -359,6 +380,198 @@ def _run_merge_jdeps_action(ctx, toolchains, jdeps, outputs, deps):
         toolchain = _TOOLCHAIN_TYPE,
     )
 
+def _run_ksp2_standalone_action(
```
restingbull, line 395: This mirrors a lot of the existing worker code -- without being able to support persistent and multiplex. 

Please look at reusing/following the worker pattern.

kotlin/internal/jvm/compile.bzl
```diff
@@ -359,6 +380,198 @@ def _run_merge_jdeps_action(ctx, toolchains, jdeps, outputs, deps):
...
+        command = """
+            mkdir -p "$TMP"
+            if [ -d "$CLASS_DIR" ] && [ "$(ls -A "$CLASS_DIR")" ]; then
+                cp -r "$CLASS_DIR"/* "$TMP/" 2>/dev/null || true
+            fi
+            if [ -d "$RESOURCE_DIR" ] && [ "$(ls -A "$RESOURCE_DIR")" ]; then
+                cp -r "$RESOURCE_DIR"/* "$TMP/" 2>/dev/null || true
+            fi
+            if [ "$(ls -A "$TMP")" ]; then
+                jar cf "$OUTPUT" -C "$TMP" .
```
restingbull: This assumes that `jar` is on the path, which can be untrue. All tools should be passed into the action.

kotlin/internal/jvm/compile.bzl
```diff
@@ -403,44 +616,62 @@ def _run_kapt_builder_actions(
...
+    # If no annotation processors, nothing to do
+    if not annotation_processors:
+        return None
+
+    kapt_id = "org.jetbrains.kotlin.kapt3"
+
+    # Check if KAPT plugin is present in classpath
+    has_kapt = any([
+        kapt_id in str(jar) or "kotlin-annotation-processing" in str(jar)
+        for jar in plugins.stubs_phase.classpath.to_list()
```
restingbull: ... this is looking at the paths jars for the plugin id? 

Asides from the issue with `to_list`, this is doesn't make sense.

kotlin/internal/jvm/compile.bzl
```diff
@@ -641,7 +880,22 @@ def _kt_jvm_produce_output_jar_actions(
     annotation_processors = _plugin_mappers.targets_to_annotation_processors(ctx.attr.plugins + ctx.attr.deps)
     ksp_annotation_processors = _plugin_mappers.targets_to_ksp_annotation_processors(ctx.attr.plugins + ctx.attr.deps)
     transitive_runtime_jars = _plugin_mappers.targets_to_transitive_runtime_jars(ctx.attr.plugins + ctx.attr.deps)
-    plugins = _new_plugins_from(ctx.attr.plugins + _exported_plugins(deps = ctx.attr.deps))
+
+    # Automatically add KAPT plugin when there are annotation processors
+    all_plugins = ctx.attr.plugins + _exported_plugins(deps = ctx.attr.deps)
+
+    # Check if KAPT plugin is already present by looking for its ID
```
restingbull: The general pattern is to build the kapt dependency into the toolchain/rule dependencies, rather than adding here.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
```diff
@@ -139,19 +139,56 @@ internal fun encodeMap(options: Map<String, String>): String {
...
+/**
+ * Find annotation processing plugin jar from classpath.
+ *
+ * Searches for the plugin jar in the stubs_plugin_classpath based on plugin ID.
+ *
+ * @param pluginId Plugin identifier (e.g., "org.jetbrains.kotlin.kapt3")
+ * @param classpath List of jar paths from stubs_plugin_classpath
+ * @return Path to plugin jar or null if not found
+ */
+private fun findAnnotationProcessingPluginJar(
```
restingbull: Walk me through why we need to find jars on the classpath?

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
```diff
@@ -161,191 +198,116 @@ internal fun JvmCompilationTask.kaptArgs(
...
-      .let { args ->
+  val processorNames = config.processorsList.joinToString(", ")
+
+  return context.execute("${config.pluginId} ($processorNames)") {
+    val args =
+      baseArgs()
+        .plus(
+          plugins(
+            options =
+              inputs.stubsPluginOptionsList.filterNot { o ->
```
restingbull, line 232: How would an `-Xplugin` be added in the options list?

If it is, why add it again? Ideally, the arguments would be generated in starlark. (the base64 encoding is still an open ended issue.)

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
```diff
@@ -161,191 +198,116 @@ internal fun JvmCompilationTask.kaptArgs(
...
+                o.startsWith(config.pluginId)
+              },
+            // Filter out the annotation processing plugin jar to avoid duplicate -Xplugin
+            // It will be added by annotationProcessingArgs() below
+            classpath = inputs.stubsPluginClasspathList.filterNot { it == pluginJar },
```
restingbull: Having poked around a bit in this -- it really doesn't seem to matter if the plugin jar is in the classpath.

If it does ,it's better to sort that out in starlark.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
```diff
@@ -161,191 +198,116 @@ internal fun JvmCompilationTask.kaptArgs(
...
   }
-}
 
-private fun kspKotlinToolchainVersion(version: String): String {
-  // KSP doesn't support Kotlin 2.0 yet, so we need to use 1.9
-  return if (version.toFloat() >= 2.0) "1.9" else version
+  // Generic annotation processing path (KAPT, etc.)
+  val config = inputs.annotationProcessing
+
+  // Extract plugin jar from classpath (provided by Starlark layer)
```
restingbull: It would be better to either pass the plugin jar as a separate argument, or generate the -Xplugin flag in starlark.

Issue #1403, opened 2025-11-24
When associating with multiple targets of the same module, the ABI jars are only included from the last target when using strict abi jars


rbeazleyspot:
TL;DR this line needs to append: https://github.com/bazelbuild/rules_kotlin/blob/master/kotlin/internal/jvm/associates.bzl#L79

`abi_jar_set = _sets.addAll(abi_jar_set, jar_bundle.abi_jars_set)`

Its possible for a target (h) to associate with another target (i) (for example some test helpers :h provide utility to setup the implementation :i in certain states)

then a target (t) adjacent to h associates with h and also target i (in order to run tests against i)
 
in this instance both targets, h and t, will adopt the kotlin module name of i

PR #1404, opened 2025-11-24
Include all abi jars from the associates array

fgoncalves:
This PR attempts to solve the issue reported by @rbeazleyspot in https://github.com/bazelbuild/rules_kotlin/issues/1403 which he basically solved. I'm just making the PR to help out.

I hope I understand the issue enough to explain it. It seems that when preparing the ABI jars only the last jar set is kept. However, we do have some scenarios where we associate 1 target with multiple other targets. Ross gives an example in the github issue posted. I often encounter this issue during tests. 

Let's say we're testing implementation `impl`.  When setting up fixtures for `impl` I often create a new target `helpers` that needs to be associated with `impl` in order to access its internal types. In order to not expose these types from `helpers` things have to be marked internal too. Then once we write the test target, this will have to associate itself with `helpers` and `impl`.

### DIFF
--- a/kotlin/internal/jvm/associates.bzl
+++ b/kotlin/internal/jvm/associates.bzl
@@ -76,7 +76,7 @@ def _get_associates(ctx, toolchains, associates):
         for a in associates:
             jar_bundle = _collect_associates(ctx = ctx, toolchains = toolchains, associate = a)
             jars.append(jar_bundle.jars)
-            abi_jar_set = jar_bundle.abi_jars_set
+            abi_jar_set = _sets.add_all(abi_jar_set, jar_bundle.abi_jars_set)
             module_names.append(a[_KtJvmInfo].module_name)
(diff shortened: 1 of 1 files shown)


restingbull:
This will need starlark tests. See [rules tests](https://github.com/bazelbuild/rules_kotlin/blob/master/src/test/starlark/compile/rule_tests.bzl) for where to get started.

rbeazleyspot:
> This will need starlark tests. See [rules tests](https://github.com/bazelbuild/rules_kotlin/blob/master/src/test/starlark/compile/rule_tests.bzl) for where to get started.

you could probably enhance the setup of the strict_abi related tests here: https://github.com/bazelbuild/rules_kotlin/blob/master/src/test/starlark/internal/jvm/jvm_deps_tests.bzl#L13-L25

not ideal but you could add another associate target to the array, just make sure it fails before applying the fix :D

PR #1405, opened 2025-11-25
Migrate from KSP1 (compiler plugin) to KSP2 (standalone tool)

agluszak:
KSP2 is a standalone program that runs separately from the Kotlin compiler, unlike KSP1 which was a compiler plugin.

- KSP now runs as a pre-compilation step producing generated sources
- Generated sources are packaged into srcjars and compiled with main sources
- Removed KSP1-specific code from KotlinToolchain and InternalCompilerPlugins
- Update examples/ksp to use bzlmod (MODULE.bazel) instead of WORKSPACE
- Expand KSP test coverage with Dagger-based coffee example demonstrating mixed Kotlin/Java code generation scenarios

### DIFF
--- a/kotlin/BUILD
+++ b/kotlin/BUILD
@@ -60,9 +60,7 @@ genrule(
     ]],
     outs = ["kotlin.md"],
     cmd = """
-    for md in $(SRCS); do
-      cat "$$md" >> $@
-    done
+    cat $(SRCS) >> $@
     """,
     executable = True,
     visibility = [


kotlin/internal/jvm/compile.bzl
```diff
@@ -428,39 +482,227 @@ def _run_kapt_builder_actions(
...
         A struct containing KSP outputs
     """
     ksp_generated_java_srcjar = ctx.actions.declare_file(ctx.label.name + "-ksp-kt-gensrc.jar")
     ksp_generated_classes_jar = ctx.actions.declare_file(ctx.label.name + "-ksp-kt-genclasses.jar")
 
-    _run_kt_builder_action(
-        ctx = ctx,
-        rule_kind = rule_kind,
+    # Declare output directories
+    ksp_kotlin_output_dir = ctx.actions.declare_directory(ctx.label.name + "-ksp-kotlin")
```
restingbull: Most of this action should be done inside a worker, without the use of tree artifacts. The current implementation will be fairly slow locally (as it will need to compress and extract artifacts both via the zipper and via the tree artifact) and incredibly slow via remote execution.

Right now, if a source file changes it will retrigger 5 actions. By moving it to a worker that handles the staging of the files and any necessary decompressing, it will cut down to one.

restingbull (changes requested):
- [x] Move the staging of the ksp action to a worker
- [x] Add worker and starlark tests

Avoid tree artifact unless absolutely necessary -- which it really isn't needed for kotlin as the jar/klib already serves the purpose of combining outputs into a single artifact.

src/main/kotlin/io/bazel/kotlin/builder/tasks/KotlinBuilder.kt
```diff
@@ -318,4 +347,266 @@ class KotlinBuilder
...
+    @Suppress("UNCHECKED_CAST")
+    private fun executeKsp2Task(
+      taskContext: WorkerContext.TaskContext,
+      argMap: ArgMap,
+    ): Int {
+      val workingDir = taskContext.directory
+      val moduleName = argMap.mandatorySingle(KotlinBuilderFlags.KSP2_MODULE_NAME)
+
+      // Create temporary directories for KSP2 processing
+      val kspWorkDir = workingDir.resolve("_ksp2/$moduleName")
```
restingbull: `workingDir.resolve("_ksp2").resolve(moduleName)`

It's usually better to let the system handle paths.

src/main/kotlin/io/bazel/kotlin/builder/tasks/KotlinBuilder.kt
```diff
@@ -318,4 +347,266 @@ class KotlinBuilder
...
+            }
+          }
+        }
+
+        // If no sources, add a placeholder source root
+        if (sourceRoots.isEmpty()) {
+          sourceRoots.add(stagedSourcesDir.toString())
+        }
+
+        // Load KSP2 via reflection
```
restingbull: It's a usually saner to use reflection once, ala [BuildToolsAPICompiler](https://github.com/bazelbuild/rules_kotlin/blob/master/src/main/kotlin/io/bazel/kotlin/compiler/BuildToolsAPICompiler.kt) than to use reflect for every class.

Reflection has overhead on the worker runtime, so we try to use it as little as possible.

src/main/kotlin/io/bazel/kotlin/builder/tasks/KotlinBuilder.kt
```diff
@@ -90,14 +98,41 @@ class KotlinBuilder
...
+      check(args.isNotEmpty()) { "expected at least a single arg got: ${args.joinToString(" ")}" }
+      val lines =
+        FLAGFILE_RE.matchEntire(args[0])?.groups?.get(1)?.let {
+          Files.readAllLines(FileSystems.getDefault().getPath(it.value), StandardCharsets.UTF_8)
+        } ?: args
+
+      val argMap = ArgMaps.from(lines)
+
+      // Check if this is a KSP2 mode request
+      if (argMap.optionalSingle(KotlinBuilderFlags.KSP2_MODE) == "true") {
```
restingbull: given that ksp is an independent action, it would make more sense to use it as a separate command -- rather than reusing the Build code. Follow the [MergeJdeps](https://github.com/bazelbuild/rules_kotlin/blob/master/src/main/kotlin/io/bazel/kotlin/builder/cmd/MergeJdeps.kt) pattern.

Frankly, the more we take out of the Build command, the better. It's pretty convoluted.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/Ksp2Invoker.kt
```diff
@@ -0,0 +1,169 @@
...
+import java.util.ServiceLoader
+
+/**
+ * Wrapper for KSP2 invocation via reflection.
+ *
+ * This class loads all necessary KSP2 classes and methods once during construction,
+ * avoiding repeated reflection overhead on each invocation. This follows the same
+ * pattern as KotlincInvoker and BuildToolsAPICompiler.
+ */
+class Ksp2Invoker(
```
restingbull: This should compile against the ksp concrete classes.

Have a look at the [BuildToolsAPICompiler]https://github.com/bazelbuild/rules_kotlin/blob/master/src/main/kotlin/io/bazel/kotlin/compiler/BuildToolsAPICompiler.kt()

restingbull: Otherwise, it will only fail at runtime. We'd rather know about inconsistencies during development/packaging the release.

restingbull:
lgtm % something weird with the stardoc doc generation? Not sure why it's dropping the ** from kwargs.

agluszak:
> lgtm % something weird with the stardoc doc generation? Not sure why it's dropping the ** from kwargs.

yeah, the weirdest thing is that the test passes for me locally ;_;

restingbull (approved):

PR #1406, opened 2025-11-25
Generate kotlinc options from kotlin-compiler-arguments-description

### LINKED ISSUE #1444: Support the new -jvm-default option along with existing -Xjvm-default
Kotlin 2.2.0 [introduced](https://kotlinlang.org/docs/whatsnew22.html#changes-to-default-method-generation-for-interface-functions) a new `-jvm-default` option as a stable replacement for `-Xjvm-default`. It has slightly different default behavior and completely replaces `-Xjvm-default` in Kotlin 2.3.0, so it's worth adding, at least in the scope of Kotlin 2.3.0 support.

agluszak:
Implement automatic generation of kotlinc options using the official kotlin-compiler-arguments-description artifact from JetBrains.

  - Auto-generated options: WriteKotlincCapabilities generates both capabilities.bzl (flag existence/stability metadata) and generated_opts.bzl (full typed Starlark options) from the compiler metadata artifact.
  - Simplified maintenance: Manual _KOPTS dict replaced with GENERATED_KOPTS import, with _MANUAL_KOPTS for special cases only.

  Changes:

  - Add kotlin-compiler-arguments-description dependency (version must match kotlin compiler version)
  - Rewrite WriteKotlincCapabilities to use the new artifact
  - Generate generated_opts_X.Y.bzl with attr.bool for booleans, attr.string for strings, attr.string_list for arrays
  - Simplify opts.kotlinc.bzl to merge generated + manual options

Closes #1444


src/main/starlark/core/options/opts.kotlinc.bzl
```diff
@@ -51,7 +58,22 @@ def _map_jdk_release_to_flag(version):
         return None
     return ["-Xjdk-release=%s" % version]
 
-_KOPTS_ALL = {
+# Manual overrides for options that need custom handling.
+# These take precedence over GENERATED_KOPTS.
+_MANUAL_KOPTS = {
+    # Escape hatch for arbitrary compiler args - not in kotlinc
+    "extra_kotlinc_args": struct(
```
restingbull: Pull this out of the PR.

I dislike exposing this on the public API without stronger guards.

src/main/starlark/core/repositories/kotlin/capabilities_2.2.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,493 @@
...
+    "-java-parameters": struct(
+        flag = "-java-parameters",
+        doc = "Generate metadata for Java 1.8 reflection on method parameters.",
+        default = False,
+        introduced = "v1_1_0",
+        stabilized = "v1_1_0",
+    ),
+    "-jvm-default": struct(
+        flag = "-jvm-default",
+        doc = """Emit JVM default methods for interface declarations with bodies. The default is 'enable'.
```
restingbull: Example of a flag that should have an enumerated list of possible values.

src/main/starlark/core/repositories/kotlin/generated_opts_2.2.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,740 @@
...
+            doc = """Compile against the specified JDK API version, similarly to javac's '-release'. This requires JDK 9 or newer.
+The supported versions depend on the JDK used; for JDK 17+, the supported versions are 1.8 and 9–24.
+This also sets the value of '-jvm-target' to be equal to the selected JDK version.""",
+            default = "",
+        ),
+        type = attr.string,
+        value_to_flag = None,
+        map_value_to_flag = _map_string_flag("-Xjdk-release"),
+    ),
+    "xjspecify_annotations": struct(
```
restingbull: Should be `x_jspecify_annotations` to match the existing attributes.

src/main/starlark/core/repositories/kotlin/generated_opts_2.2.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,740 @@
...
+        flag = "-Xenhanced-coroutines-debugging",
+        args = dict(
+            doc = """Generate additional linenumber instruction for compiler-generated code
+inside suspend functions and lambdas to distinguish them from user code by debugger.""" + "\nValues: unset = use kotlinc default, 'true' = enable, 'false' = disable.",
+            values = ["true", "false"],
+        ),
+        type = attr.string,
+        value_to_flag = {"true": ["-Xenhanced-coroutines-debugging"]},
+    ),
+    "xfriend_paths": struct(
```
restingbull: This is an example of a flag that should not be exposed as an option.

`friends` is handled by `associates`, as bazel needs to stage the files correctly.

Not all options can be exposed for this reason -- some of them belong in the rules, as they really cannot be used correctly.

src/main/starlark/core/repositories/compiler.bzl
```diff
@@ -39,15 +39,27 @@ def _kotlin_capabilities_impl(repository_ctx):
...
         "capabilities.bzl",
         template,
         executable = False,
     )
+    generated_opts_template = _get_template_by_version(
+        attr.compiler_version,
+        [repository_ctx.path(ct) for ct in attr._generated_opts_templates],
+        "generated_opts",
+    )
+    if generated_opts_template:
```
restingbull, line 51: this is not backwards compatible. Loading version sof kotlin without a generated options will fail, as the opts.kotlinc.bzl will always try to load the generated capabilities.

restingbull (changes requested):
Aside from a few small backward compatibility issues, the approach looks ok.

Need to be solved:
* Enumerated flags will need a set of valid values -- because you'd rather have the build fail in analysis that compiler invocation. Some invocations can live very deep in a build graph, which will cause maintenance headaches.
* Flags must live on an allow list. The `friends` flag is the best example, as it pretty much cannot be used correctly, and overlaps with existing rule functionality.

Finally, please remove `extra_kotlinc_args`. That will inevitably be misused in large code repositories for all the reasons that the flags are treated as attributes.

docs/kotlin.md
```diff
@@ -407,40 +419,75 @@ Define kotlin compiler options.
 | :------------- | :------------- | :------------- | :------------- | :------------- |
 | <a id="kt_kotlinc_options-name"></a>name |  A unique name for this target.   | <a href="https://bazel.build/concepts/labels#target-names">Name</a> | required |  |
 | <a id="kt_kotlinc_options-include_stdlibs"></a>include_stdlibs |  Don't automatically include the Kotlin standard libraries into the classpath (stdlib and reflect).   | String | optional |  `"all"`  |
-| <a id="kt_kotlinc_options-java_parameters"></a>java_parameters |  Generate metadata for Java 1.8+ reflection on method parameters.   | Boolean | optional |  `False`  |
+| <a id="kt_kotlinc_options-java_parameters"></a>java_parameters |  Generate metadata for Java 1.8 reflection on method parameters. Values: unset = use kotlinc default, 'true' = enable, 'false' = disable.   | String | optional |  `""`  |
```
restingbull: Should be a boolean, not a string.

agluszak: I mean, sure, we can leave this default-ness issue for another PR, but that's something we wanted to solve for our internal use at JB, cc @shalupov

docs/kotlin.md
```diff
@@ -407,40 +419,75 @@ Define kotlin compiler options.
...
-| <a id="kt_kotlinc_options-x_skip_prerelease_check"></a>x_skip_prerelease_check |  Suppress errors thrown when using pre-release classes.   | Boolean | optional |  `False`  |
-| <a id="kt_kotlinc_options-x_suppress_version_warnings"></a>x_suppress_version_warnings |  Suppress warnings about outdated, inconsistent, or experimental language or API versions.   | Boolean | optional |  `False`  |
+| <a id="kt_kotlinc_options-x_sanitize_parentheses"></a>x_sanitize_parentheses |  Transform '(' and ')' in method names to some other character sequence. This mode can BREAK BINARY COMPATIBILITY and should only be used as a workaround for problems with parentheses in identifiers on certain platforms.   | Boolean | optional |  `False`  |
+| <a id="kt_kotlinc_options-x_script_resolver_environment"></a>x_script_resolver_environment |  Set the script resolver environment in key-value pairs (the value can be quoted and escaped).   | List of strings | optional |  `[]`  |
```
restingbull: This is not a sensical option for a bazel invocation of kotlinc.

docs/kotlin.md
```diff
@@ -407,40 +419,75 @@ Define kotlin compiler options.
...
-| <a id="kt_kotlinc_options-x_no_param_assertions"></a>x_no_param_assertions |  Don't generate not-null assertions on parameters of methods accessible from Java   | Boolean | optional |  `False`  |
-| <a id="kt_kotlinc_options-x_no_receiver_assertions"></a>x_no_receiver_assertions |  Don't generate not-null assertion for extension receiver arguments of platform types   | Boolean | optional |  `False`  |
-| <a id="kt_kotlinc_options-x_no_source_debug_extension"></a>x_no_source_debug_extension |  Do not generate @kotlin.jvm.internal.SourceDebugExtension annotation on a class with the copy of SMAP   | Boolean | optional |  `False`  |
+| <a id="kt_kotlinc_options-x_link_via_signatures"></a>x_link_via_signatures |  Link JVM IR symbols via signatures instead of descriptors. This mode is slower, but it can be useful for troubleshooting problems with the JVM IR backend. This option is deprecated and will be deleted in future versions. It has no effect when -language-version is 2.0 or higher.   | Boolean | optional |  `False`  |
+| <a id="kt_kotlinc_options-x_module_path"></a>x_module_path |  Paths to Java 9+ modules.   | String | optional |  `""`  |
```
restingbull: Any option that requires a path is pretty much not usable without direct rule support. So they should not be exposed.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,668 @@
...
+        flag = "-Xadd-modules",
+        args = dict(
+            doc = "Root modules to resolve in addition to the initial modules, or all modules on the module path if <module> is ALL-MODULE-PATH.",
+            default = [],
+        ),
+        type = attr.string_list,
+        value_to_flag = None,
+        map_value_to_flag = _map_string_list_flag("-Xadd-modules"),
+    ),
+    "x_allow_no_source_files": struct(
```
restingbull: This is doesn't make sense in the context of the rule invocation, as the rules will fail this case before the flag. Exposing it is misleading to the user.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,668 @@
...
+    ),
+    "x_compile_builtins_as_part_of_stdlib": struct(
+        flag = "-Xcompile-builtins-as-part-of-stdlib",
+        args = dict(
+            doc = "Enable behaviour needed to compile builtins as part of JVM stdlib",
+        ),
+        type = attr.bool,
+        value_to_flag = {True: ["-Xcompile-builtins-as-part-of-stdlib"]},
+    ),
+    "x_compile_java": struct(
```
restingbull: Another option that shouldn't be exposed. Java compilation happens via the java builder, so making this visible will add to the confusion.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,668 @@
...
+        flag = "-Xdebug",
+        args = dict(
+            doc = """Enable debug mode for compilation.
+    Currently this includes spilling all variables in a suspending context regardless of whether they are alive.
+    If API Level >= 2.2 -- no-op.""",
+        ),
+        type = attr.bool,
+        value_to_flag = {True: ["-Xdebug"]},
+    ),
+    "x_default_script_extension": struct(
```
restingbull: Do any of the script flags make sense to be visible?

I'm not clear if you can compile scripts via the rule sets. There aren't any tests, at least.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,668 @@
...
+    ),
+    "x_jvm_expose_boxed": struct(
+        flag = "-Xjvm-expose-boxed",
+        args = dict(
+            doc = "Expose inline classes and functions, accepting and returning them, to Java.",
+        ),
+        type = attr.bool,
+        value_to_flag = {True: ["-Xjvm-expose-boxed"]},
+    ),
+    "x_klib": struct(
```
restingbull: Any option that needs a path is not really viable to use in a rules invocation without explicit support.

The cross platform libraries would need to be included in the action inputs, and reasoning about the path is not easy.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,668 @@
...
+    ),
+    "x_no_receiver_assertions": struct(
+        flag = "-Xno-receiver-assertions",
+        args = dict(
+            doc = "Don't generate not-null assertions for extension receiver arguments of platform types.",
+        ),
+        type = attr.bool,
+        value_to_flag = {True: ["-Xno-receiver-assertions"]},
+    ),
+    "x_no_reset_jar_timestamps": struct(
```
restingbull: The rules explicitly zero out the timestamps, so this is a confusing option.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,668 @@
...
+    * strict
+    * warn (report a warning)""",
+            default = [],
+        ),
+        type = attr.string_list,
+        value_to_flag = None,
+        map_value_to_flag = _map_string_list_flag("-Xnullability-annotations"),
+    ),
+    "x_output_builtins_metadata": struct(
+        flag = "-Xoutput-builtins-metadata",
```
restingbull: I'm not sure this one makes sense, as I do not think they would be added to the outputs.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,668 @@
...
+        args = dict(
+            doc = "Output builtins metadata as .kotlin_builtins files",
+        ),
+        type = attr.bool,
+        value_to_flag = {True: ["-Xoutput-builtins-metadata"]},
+    ),
+    "x_profile": struct(
```
restingbull: option requires a path as input and produces outputs that will be lost. Should not be exposed without explicit rule support.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,668 @@
...
+    ),
+    "x_use_inline_scopes_numbers": struct(
+        flag = "-Xuse-inline-scopes-numbers",
+        args = dict(
+            doc = "Use inline scopes numbers for inline marker variables.",
+        ),
+        type = attr.bool,
+        value_to_flag = {True: ["-Xuse-inline-scopes-numbers"]},
+    ),
+    "x_use_javac": struct(
```
restingbull: Java sources are compiled via a different pipeline, so this shouldn't be visible.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,668 @@
...
+        flag = "-Xuse-javac",
+        args = dict(
+            doc = "Use javac for Java source and class file analysis.",
+        ),
+        type = attr.bool,
+        value_to_flag = {True: ["-Xuse-javac"]},
+    ),
+    "x_use_k2_kapt": struct(
```
restingbull: kapt is explicitly supported via the rules, so this flag should not be visible.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,668 @@
...
+    -jvm-default=no-compatibility    Generate default methods for non-abstract interface declarations. Do not generate 'DefaultImpls' classes.
+    -jvm-default=disable             Do not generate JVM default methods. This is the default behavior up to language version 2.1.""",
+            values = ["", "enable", "no-compatibility", "disable"],
+        ),
+        type = attr.string,
+        value_to_flag = None,
+        map_value_to_flag = _map_string_flag("-jvm-default"),
+    ),
+    "jvm_target": struct(
+        flag = "-jvm-target",
```
restingbull: This is [set by the task](https://github.com/bazelbuild/rules_kotlin/blob/f2533476987a65046f83605b530c87527a2ae933/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt#L92), so it shouldn't be visible.

src/main/kotlin/io/bazel/kotlin/generate/WriteKotlincCapabilities.kt
```diff
@@ -98,6 +126,30 @@ object WriteKotlincCapabilities {
...
+  /**
+   * Known enumerated values for flags that have restricted choices.
+   * These values are extracted from kotlinc documentation and used to
+   * validate at analysis time rather than compilation time.
+   */
+  private val enumeratedFlagValues = mapOf(
+    "-jvm-default" to listOf("enable", "no-compatibility", "disable"),
+    "-Xassertions" to listOf("always-enable", "always-disable", "jvm", "legacy"),
+    "-Xlambdas" to listOf("class", "indy"),
+    "-Xsam-conversions" to listOf("class", "indy"),
```
restingbull: The reflection approach [could pull information from the valueDescription.](https://github.com/JetBrains/kotlin/blob/676a356468adfd5fb615cdce1db7c7319fe874b3/compiler/cli/cli-common/gen/org/jetbrains/kotlin/cli/common/arguments/K2JVMCompilerArguments.kt#L534)

Does the json expose something similar?

src/main/starlark/core/options/opts.kotlinc.bzl
```diff
@@ -98,50 +94,8 @@ _KOPTS_ALL = {
...
-        args = dict(
-            default = False,
-            doc = "Enable kotlin.Result as a return type",
-        ),
-        type = attr.bool,
-        value_to_flag = {
-            True: ["-Xallow-result-return-type"],
-        },
-    ),
-    "x_annotation_default_target": struct(
```
restingbull, line 133: We seem to have lost the common compiler arguments, [like this one](https://github.com/JetBrains/kotlin/blob/676a356468adfd5fb615cdce1db7c7319fe874b3/compiler/arguments/src/org/jetbrains/kotlin/arguments/description/CommonCompilerArguments.kt#L1048-L1070).

agluszak:
> @agluszak if we are making incompatible release anyway (Boolean -> String)

that's reverted

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,1337 @@
...
+    ),
+    "x_allow_unstable_dependencies": struct(
+        flag = "-Xallow-unstable-dependencies",
+        args = dict(
+            doc = "Do not report errors on classes in dependencies that were compiled by an unstable version of the Kotlin compiler.",
+        ),
+        type = attr.bool,
+        value_to_flag = {True: ["-Xallow-unstable-dependencies"]},
+    ),
+    "x_annotation_default_target": struct(
```
restingbull: This should be an [enum](https://github.com/bazelbuild/rules_kotlin/blob/b3793d8e9d68e8c616ce6a5648efaa184debd232/src/main/starlark/core/options/opts.kotlinc.bzl#L144-L149)

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,1337 @@
...
+    ),
+    "x_check_phase_conditions": struct(
+        flag = "-Xcheck-phase-conditions",
+        args = dict(
+            doc = "Check pre- and postconditions of IR lowering phases.",
+        ),
+        type = attr.bool,
+        value_to_flag = {True: ["-Xcheck-phase-conditions"]},
+    ),
+    "x_common_sources": struct(
```
restingbull: This adds source files to the compilation -- it's not a reasonable option for the compiler.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,1337 @@
...
+    "x_explicit_backing_fields": struct(
+        flag = "-Xexplicit-backing-fields",
+        args = dict(
+            doc = "Enable experimental language support for explicit backing fields.",
+        ),
+        type = attr.bool,
+        value_to_flag = {True: ["-Xexplicit-backing-fields"]},
+    ),
+    "x_fragment_dependency": struct(
+        flag = "-Xfragment-dependency",
```
restingbull: Another path flag.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,1337 @@
...
+    This argument is required for any HMPP module except the platform leaf module: it takes dependencies from -cp/-libraries.
+    The argument should be used only if the new compilation scheme is enabled with -Xseparate-kmp-compilation
+    """,
+            default = [],
+        ),
+        type = attr.string_list,
+        value_to_flag = None,
+        map_value_to_flag = _map_string_list_flag("-Xfragment-dependency"),
+    ),
+    "x_fragment_friend_dependency": struct(
```
restingbull: Path.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,1337 @@
...
+    The argument should be used only if the new compilation scheme is enabled with -Xseparate-kmp-compilation
+    """,
+            default = [],
+        ),
+        type = attr.string_list,
+        value_to_flag = None,
+        map_value_to_flag = _map_string_list_flag("-Xfragment-friend-dependency"),
+    ),
+    "x_fragment_refines": struct(
+        flag = "-Xfragment-refines",
```
restingbull: This is a wierd one. I'm not sure this make sense to be exposed as an option -- it seems more like a feature that should be defined in the rules.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,1337 @@
...
+        args = dict(
+            doc = "Declare that <fromModuleName> refines <onModuleName> with the dependsOn/refines relation.",
+            default = [],
+        ),
+        type = attr.string_list,
+        value_to_flag = None,
+        map_value_to_flag = _map_string_list_flag("-Xfragment-refines"),
+    ),
+    "x_fragment_sources": struct(
```
restingbull: Path flag.

src/main/starlark/core/repositories/kotlin/generated_opts_2.3.bzl.com_github_jetbrains_kotlin.bazel
```diff
@@ -0,0 +1,1337 @@
...
+        flag = "-Xfragment-sources",
+        args = dict(
+            doc = "Add sources to a specific fragment of a multiplatform compilation.",
+            default = [],
+        ),
+        type = attr.string_list,
+        value_to_flag = None,
+        map_value_to_flag = _map_string_list_flag("-Xfragment-sources"),
+    ),
+    "x_fragments": struct(
```
restingbull: See other `-Xfragment` flags. Pretty sure this are difficult, if not impossible to use correct when compiling via the rules.

restingbull (commented):
Given how often flags that do not make sense within the rules are sneaking into the generated options, it seems like having an allow list would make a lot more sense than a disallow list.

PR #1411, opened 2025-11-30
Fix Windows bzlmod builder classpath issue

agluszak:
## Summary
- Fixes Windows bzlmod builder classpath issue where Rlocation fails for canonical repo names starting with `+`
- Adds Kotlin stdlib to builder's `runtime_deps` instead of using `--main_advice_classpath`
- Removes the problematic wrapper script flag that caused manifest lookup failures

## Problem
On Windows with bzlmod, the Java launcher's Rlocation mechanism doesn't correctly handle canonical repository names (starting with `+`). When `--wrapper_script_flag=--main_advice_classpath=<path>` is processed, the launcher incorrectly prepends `_main/` to paths, causing manifest lookup failures:

```
LAUNCHER ERROR: Rlocation failed on _main/+rules_kotlin_extensions+com_github_jetbrains_kotlin_git/lib/annotations-13.0.jar, path doesn't exist in MANIFEST file
```

## Solution
Instead of relying on the buggy `--main_advice_classpath` mechanism, add the Kotlin stdlib directly to the builder's `runtime_deps`. This ensures the stdlib jars are on the classpath at startup without using the Windows launcher's path resolution.

## Test plan
- [x] Verified rules_kotlin builds successfully on Windows
- [x] Verified intellij-bazel builds successfully with the fixed rules_kotlin


agluszak:
Btw, can we add a windows job to the CI matrix?

restingbull:
> Btw, can we add a windows job to the CI matrix?

Sure -- it's broken but you can give it shot to get it passing.

src/main/kotlin/io/bazel/kotlin/builder/cmd/BUILD.bazel
```diff
@@ -70,6 +71,7 @@ kt_bootstrap_library(
 kt_bootstrap_binary(
     name = "merge_jdeps",
     data = _KOTLIN_STDLIBS,
```
restingbull: It doesn't need the std libs in data with the change to runtime libs, does it?

agluszak:
> Sure -- it's broken but you can give it shot to get it passing.

What's broken specifically?

src/main/kotlin/io/bazel/kotlin/builder/cmd/BUILD.bazel
agluszak: Apparently they are required to be in both places :/

agluszak: https://buildkite.com/bazel/rules-kotlin-kotlin/builds/5559/steps/canvas

PR #1421, opened 2025-12-10
chore: Add a target to update docs automatically

agluszak:
Up-to-date test is created by the macro automatically

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -75,7 +75,7 @@ tasks:
     build_targets:
       - //kotlin:stardoc
     test_targets:
-      - //docs:are_docs_up_to_date_test
+      - //docs/...
   ktlint:
     name: KtLint
     <<: *ubuntu2404


docs/BUILD.bazel
```diff
@@ -1,15 +1,8 @@
-load("@rules_shell//shell:sh_test.bzl", "sh_test")
+load("@aspect_bazel_lib//lib:write_source_files.bzl", "write_source_files")
```
restingbull, line 1: While docs would be generated, they wouldn't be part of the commit, would they?

Unless I'm missing something?

agluszak: >  they wouldn't be part of the commit

not sure what you mean, but `bazel run //docs:write_docs` will copy the generated file to docs/kotlin.md

agluszak: https://github.com/bazel-contrib/bazel-lib/blob/2.x/docs/write_source_files.md

restingbull: Right -- but making it part of the CI action would not add the new files to the PR. It would still need to be invoked and added to git before the CI runs.

The point of the docs up to date test is to ensure the documentation is updated with changes. This change removes that test, which increases the chance of the docs being out of date.

agluszak: No, it doesn't remove the test, the test is still there, but it's created by the macro, the name is `//docs:write_docs_test`

PR #1422, opened 2025-12-10
.bazelrc: do not add IntelliJ-internal packages to --deleted_pacakges

aehlig:


restingbull:
> > This is a generated file.
> 
> If you are talking about the files in `.bazelbsp`, yes, they are generated by IntelliJ in order to be used during its synchronssation build. Therefore it is a problem that they are marked as deleted.
> 
> If you are talking about `.bazelrc`, then what data is it generated from?
> 
> > What is the issue you are trying to solve?
> 
> https://youtrack.jetbrains.com/issue/BAZEL-2753/

You might try tweaking [.bazelignore](https://github.com/bazelbuild/rules_kotlin/blob/master/.bazelignore) and regenerating the deleted packages and see if that fixes it.

aehlig:
> See `bazel run @rules_bazel_integration_test//tools:update_deleted_packages` -- per the comment above the changes.

Running that command on a clean checkout, I get precisely the diff proposed in this PR.

agluszak:
>  whoever last updated `.bazelrc` as not using a clean checkout but one where InthelliJ with an older version of the bazel plugin was running before the update

Okay, this explains what happened and why I didn't remember adding .bazelbsp to .bazelrc manually :)

Issue #1397, opened 2025-11-11
local development does not work (java.lang.NoClassDefFoundError: dagger/internal/Factory)


adrianimboden:
Hi all

I found https://github.com/bazelbuild/rules_kotlin/issues/1235, which is closed. It references the same error and as a response, it mentions to follow the local development manual (https://github.com/bazelbuild/rules_kotlin/blob/master/README.md#with-bzlmod).

I did that, but it still fails with this:
```
ERROR: .../BUILD:3:11: JdepsMerge //gaCommon/src/main/java/com/getabstract/common/util/mail:mail { jdeps: 1 } failed: Worker process did not return a WorkResponse:

---8<---8<--- Start of log, file at .../bazel-workers/worker-15-JdepsMerge.log ---8<---8<---
Exception in thread "main" java.lang.NoClassDefFoundError: dagger/internal/Factory
        at java.base/java.lang.ClassLoader.defineClass1(Native Method)
        at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1027)
        at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
        at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:862)
        at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:760)
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:681)
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:639)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
        at io.bazel.kotlin.builder.DaggerJdepsMergerComponent$JdepsMergerComponentImpl.work(DaggerJdepsMergerComponent.java:50)
        at io.bazel.kotlin.builder.cmd.MergeJdeps.main$lambda$0(MergeJdeps.kt:29)
        at io.bazel.worker.Worker$Companion.from(Worker.kt:32)
        at io.bazel.kotlin.builder.cmd.MergeJdeps.main(MergeJdeps.kt:28)
Caused by: java.lang.ClassNotFoundException: dagger.internal.Factory
[...]

PR #1423, opened 2025-12-10
maintenance: Remove dagger from the builder

agluszak:


agluszak:
Let my reverse the question: why is it there in the first place? It doesn't seem to make anything simpler/easier, only adds boilerplate. I thought maybe it served some purpose in the past and now it's legacy? We removed it in our internal fork

agluszak:
I mean, my main argument is simplicity and a negative diff after removing it, but if you like dagger because of code style preferences - I don't want to start a tabs vs spaces holy war :)

agluszak:
Also, https://github.com/bazelbuild/rules_kotlin/issues/1397

PR #1424, opened 2025-12-10
maintenance: Remove non-BTAPI compilation

agluszak:
BTAPI is already used in Maven and Gradle plugins


restingbull (changes requested):
This needs to enable by default for a release or so before we can make the transition.

Maven and Gradle are separate code paths -- we commonly find truly enjoyable bugs that don't show up elsewhere.

PR #1425, opened 2025-12-12
chore: Bump bazel version to 8.5.0

agluszak:


MODULE.bazel
```diff
@@ -135,11 +135,11 @@ bazel_dep(name = "rules_bazel_integration_test", version = "0.34.0", dev_depende
 
 bazel_binaries = use_extension("@rules_bazel_integration_test//:extensions.bzl", "bazel_binaries", dev_dependency = True)
 bazel_binaries.download(version_file = "//:.bazelversion")
-bazel_binaries.download(version = "8.2.1")
+bazel_binaries.download(version = "8.5.0")
```
restingbull, line 138: This duplicates the `bazelversion`. It should be an older version, rather than the same version in use.

Bencodes: At this point we should probably update this to 9.x or bump our internal version to 9.x and leave the manual download to the latest 8.x stable release.

PR #1436, opened 2026-01-04
[maintenance] Remove cruft

agluszak:


agluszak:
> @agluszak something's up with the load formatting, they all ended up above the file licensing.

fixed

PR #1443, opened 2026-01-08
[maintenance] Update rules_android

agluszak:


examples/android/WORKSPACE
```diff
@@ -37,6 +37,36 @@ http_archive(
...
+    strip_prefix = versions.BAZEL_FEATURES.strip_prefix_template.format(version = versions.BAZEL_FEATURES.version),
+    urls = [url.format(version = versions.BAZEL_FEATURES.version) for url in versions.BAZEL_FEATURES.url_templates],
+)
+
+load("@bazel_features//:deps.bzl", "bazel_features_deps")
+
+bazel_features_deps()
+
+http_archive(
+    name = "rules_cc",
```
Bencodes: Was adding ruels_cc a requirement here?

agluszak: yes, because of https://github.com/bazelbuild/bazel/issues/27582 (explicitly loading rules_cc here downgrades it)

But it looks like updating bazel to 8.5.0 should fix it

tl;dr yet another instance of WORKSPACE dependency hell

agluszak: Yeah, so unfortunately rules_cc has to be there in WORKSPACE, see the latest CI run :(

```
ERROR: Failed to load Starlark extension '@@cc_compatibility_proxy//:symbols.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @@cc_compatibility_proxy
This could either mean you have to add the '@@cc_compatibility_proxy' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: cannot load build configuration key because of this cycle
```

Bencodes (approved):

PR #1445, opened 2026-01-08
[maintenance] Add Kotlin 2.3 compiler options definitions

### LINKED ISSUE #1444: Support the new -jvm-default option along with existing -Xjvm-default

agluszak:
Closes #1444


src/main/starlark/core/options/opts.kotlinc.bzl
```diff
@@ -75,6 +75,27 @@ _KOPTS_ALL = {
             True: ["-java-parameters"],
         },
     ),
+    "jvm_default": struct(
```
Bencodes, line 78: https://github.com/agluszak/rules_kotlin/blob/36ec69eba8fb9e3e5c18a3529268b6938192bbed/src/main/starlark/core/options/opts.kotlinc.bzl#L295

Doesn't this already exist?

restingbull:
> But there's #1406 still

with open comments.

PR #1448, opened 2026-01-20
Fix JaCoCo coverage issue with rules_java 9.3.0+ 

### LINKED ISSUE #1447: [coverage] Incompatible with rules_java 9.3.0
Repro https://github.com/dmivankov/bug_reports/tree/java93_kotlin_coverage
Short version is `src/main/kotlin/kotlin_worker.jar` is compiled against jacoco 0.8.11 (uses `org/jacoco/agent/rt/internal_4742761/Offline.class`) but `rules_java 9.3.0` upgrades jacoco to 0.8.14 (uses `org/jacoco/agent/rt/internal_29a6edd/Offline.class`).

Running `bazel coverage` with `rules_java 9.3.0` project on kotlin targets results in
```
Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.NoClassDefFoundError: org/jacoco/agent/rt/internal_4742761/Offline [in thread "main"]
```

Potential fixes
- breaking-change upgrade `rules_java` for `rules_kotlin` releases assets
- build `kotlin_worker` from source
- something to support both 0.8.11 and 0.8.14, at least for some time

agluszak:
Reorder runtime_deps to put @bazel_tools//tools/jdk:JacocoCoverage before the worker. This ensures the newer JaCoCo classes are loaded first, fixing the NoClassDefFoundError caused by version mismatch between rules_kotlin (JaCoCo 0.8.11) and rules_java 9.3.0+ (JaCoCo 0.8.14).

Also adds a coverage example to verify the fix.

Fixes #1447

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/test/BazelIntegrationTestRunner.kt
+++ b/src/main/kotlin/io/bazel/kotlin/test/BazelIntegrationTestRunner.kt
@@ -152,8 +152,8 @@ object BazelIntegrationTestRunner {
         // Run test script if it exists
         val testScript = workspace.resolve("test.sh")
         if (testScript.exists()) {
-          val bash = System.getenv("BIT_BASH_BINARY")
-            ?: throw AssertionError("BIT_BASH_BINARY environment variable not set")
+          val bashPathFile = BazelRunFiles.resolveVerifiedFromProperty(fs, "io.bazel.kotlin.test.bash_path")
+          val bash = Files.readString(bashPathFile).trim()
           println("Running test script [${testScript.fileName}]...")
           ProcessBuilder()
             .command(bash, testScript.toString())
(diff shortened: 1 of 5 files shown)


examples/coverage/test_coverage.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
Bencodes: Nothing actually invokes this script to verify that coverage actually works, is that intentional?

src/main/kotlin/io/bazel/kotlin/test/BazelIntegrationTestRunner.kt
```diff
@@ -148,6 +148,43 @@ object BazelIntegrationTestRunner {
             ).onFailThrow()
           }
         }
+
```
restingbull, line 151: Why not just add coverage as a command to the integration test?

PR #1455, opened 2026-02-13
Fix transitive dependency propagation for sourceless targets

olbapmar:
Fix transitive dependency propagation for sourceless targets (e.g. android_library with only exports/deps and no srcs) in `_reshade_embedded_kotlinc_jars`
                                                                                                                                                                                                                                         
## Why

When a target has no source files, jars is empty, so the reshading comprehension produces nothing. The final java_common.merge then merges an empty list, silently dropping all deps. This breaks downstream consumers that rely on the full processor classpath being propagated transitively through these wrapper targets.


olbapmar:
> Add a starlark test for this?

Done :)

olbapmar:
@Bencodes Just merged from `master` to pick up the fix for windows and unblock CI

PR #1456, opened 2026-02-14
fix: Conditionally set repo rules as reproducible

mortenmj:
This allows the repository rules to use the remote repo content cache, which allows users who opt into that feature to avoid downloading files created by these repo rules when using remote execution. Output below shows external files fetched when building our internal kotlin libraries with and without this patch.

Before:
```shell
external # du -h -d1 | sort -h | grep rules_kotlin
4.0K    ./rules_kotlin++rules_kotlin_extensions+kotlin_build_tools_impl
4.0K    ./rules_kotlin++rules_kotlin_extensions+kotlinx_serialization_core_jvm
4.0K    ./rules_kotlin++rules_kotlin_extensions+kotlinx_serialization_json
4.0K    ./rules_kotlin++rules_kotlin_extensions+kotlinx_serialization_json_jvm
56K     ./rules_kotlin++rules_kotlin_extensions+com_github_jetbrains_kotlin
76M     ./rules_kotlin++rules_kotlin_extensions+com_github_google_ksp
83M     ./rules_kotlin++rules_kotlin_extensions+com_github_jetbrains_kotlin_git
```

After:

```shell
external # du -h -d1 | sort -h | grep rules_kotlin
4.0K    ./rules_kotlin++rules_kotlin_extensions+kotlin_build_tools_impl
4.0K    ./rules_kotlin++rules_kotlin_extensions+kotlinx_serialization_core_jvm
4.0K    ./rules_kotlin++rules_kotlin_extensions+kotlinx_serialization_json
4.0K    ./rules_kotlin++rules_kotlin_extensions+kotlinx_serialization_json_jvm
56K     ./rules_kotlin++rules_kotlin_extensions+com_github_jetbrains_kotlin
```

### DIFF
--- a/src/main/starlark/core/repositories/compiler.bzl
+++ b/src/main/starlark/core/repositories/compiler.bzl
@@ -62,7 +62,7 @@ def _kotlin_capabilities_impl(repository_ctx):
         return None
 
     return repository_ctx.repo_metadata(
-        reproducible = True
+        reproducible = True,
     )
 
 def _coerce_int(string_value):


src/main/starlark/core/repositories/compiler.bzl
```diff
@@ -19,6 +19,14 @@ def _kotlin_compiler_impl(repository_ctx):
         executable = False,
     )
 
+    # Bazel <8.3.0 lacks repository_ctx.repo_metadata
+    if not hasattr(repository_ctx, "repo_metadata"):
```
Bencodes, line 23: I believe there is a helper function floating around in one of the libraries for this.

mortenmj: @Bencodes we could use `bazel_features.external_deps.repo_metadata_has_reproducible`, which uses the current Bazel version to make the same determination. Is that the function you mean? I don't think there's any advantage to using it. If there's some other helper I haven't been able to find, I'd appreciate if you could point me to it.

mortenmj: @Bencodes ping

PR #1457, opened 2026-02-15
[maintenance] Remove unused bazel_deps from MODULE.release.bazel

agluszak:


MODULE.release.bazel
```diff
@@ -5,36 +5,12 @@ module(
...
-
 bazel_dep(name = "platforms", version = "0.0.11")
 bazel_dep(name = "bazel_skylib", version = "1.7.1")
 bazel_dep(name = "rules_java", version = "8.9.0")
 bazel_dep(name = "rules_android", version = "0.7.1")
 bazel_dep(name = "bazel_features", version = "1.39.0")
 bazel_dep(name = "bazel_lib", version = "3.1.0")
-bazel_dep(name = "protobuf", version = proto_module_version, repo_name = "com_google_protobuf")
-bazel_dep(name = "rules_cc", version = "0.0.16")
-bazel_dep(name = "bazel_worker_api", version = "0.0.10")
```
Bencodes, line 22: I assume these aren't needed because the entire worker is being compiled directly into the builder jar?

Bencodes: I'm now wondering if we should be including these as runtime dependencies on the builder so that we aren't explicitly coupling ourselves to worker versions. 

Moving away from pre-compiled releases would also address this.

agluszak: > explicitly coupling ourselves to worker versions

I'm not sure if there really are _worker_ versions. The proto itself that defines communication with worker [hasn't changed in two years](https://github.com/bazelbuild/bazel-worker-api/blob/main/proto/worker_protocol.proto). Version of the `bazel_worker_api` bazel_dep doesn't really matter. 

What we _are_ coupling ourselves to is the protobuf compiler version.

agluszak: I guess we could also switch away from protobuf-based communication to json-based... https://bazel.build/remote/creating#making-worker

Bencodes: The `bazel_worker` json impl. is misleading. Behind the scenes it's serializing protos to json, even if you use json based worker communication you are still coupled to protos.

agluszak: in either case, I think for now we can remove these bazel_deps from the release MODULE.bazel

agluszak: @Bencodes bump

PR #1465, opened 2026-02-25
Use BTAPI

agluszak:
What BTAPI is: https://kotlinlang.org/docs/build-tools-api.html

 ["[Build Tools API] should be the preferred way to work with Kotlin compiler when integrating Kotlin builds into different build systems."](https://github.com/JetBrains/kotlin/blob/master/compiler/build-tools/kotlin-build-tools-api/README.md)

Pulled out of https://github.com/bazelbuild/rules_kotlin/pull/1449

Includes migration of Kotlin/KSP/compiler artifacts to rules_jvm_external (most of the diff comes from `kotlin_rules_maven_install.json`), BTAPI runtime/toolchain plumbing, plugin payload handling. Explicitly excludes incremental compilation implementation, ic_tests, and compiler-argument-parsing generation changes ("capabilities").

BTAPI unlocks:
- kotlinc-driven sub-target incremental compilation
- possibility to use any kotlin compiler version within BTAPI compatibility range
- simplification of the internal plumbing
- general futureproofing

I removed a lot of code that either became dead with the removal of BazelK2JVMCompiler.kt or had been dead earlier.

Custom compiler version is now set by providing a btapi_impl.jar to the toolchain. See [examples/btapi_compat](https://github.com/bazelbuild/rules_kotlin/pull/1465/changes#diff-a934ff7bbba16b7672a538e99a76ee63651ebeacf263785011532ecf4d66bae7).

The reason why I made the switch to rules_jvm_external for dependency management is that btapi pulls in a lot of dependencies that need to have matching versions and resolving them manually is painful. With rules_jvm_external you can say which compiler version you want and have maven resolve the deps for you.


src/main/kotlin/io/bazel/kotlin/builder/tasks/KotlinBuilder.kt
```diff
@@ -82,7 +81,17 @@ class KotlinBuilder(
       STRICT_KOTLIN_DEPS("--strict_kotlin_deps"),
       REDUCED_CLASSPATH_MODE("--reduced_classpath_mode"),
       INSTRUMENT_COVERAGE("--instrument_coverage"),
-      BUILD_TOOLS_API("--build_tools_api"),
+      BUILD_TOOLS_IMPL("--build_tools_impl"),
```
agluszak: it should be enough to pass this with all transitive deps

src/main/kotlin/io/bazel/kotlin/builder/tasks/KotlinBuilder.kt
```diff
@@ -82,7 +81,17 @@ class KotlinBuilder(
...
+      KOTLIN_COMPILER_EMBEDDABLE("--kotlin_compiler_embeddable"),
+      KOTLIN_DAEMON_CLIENT("--kotlin_daemon_client"),
+      KOTLIN_STDLIB("--kotlin_stdlib"),
+      KOTLIN_REFLECT("--kotlin_reflect"),
```
agluszak: should be the same version

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/BtapiPluginArguments.kt
```diff
@@ -0,0 +1,99 @@
...
+
+import org.jetbrains.kotlin.buildtools.api.arguments.CompilerPlugin
+import org.jetbrains.kotlin.buildtools.api.arguments.CompilerPluginPartialOrderRelation
+import kotlin.io.path.absolutePathString
+
+/**
+ * Converts typed BTAPI [CompilerPlugin] declarations into raw argument strings.
+ *
+ * Matches kotlin-build-tools-impl plugin argument formatting so the resulting
+ * arguments can be applied with [org.jetbrains.kotlin.buildtools.api.arguments.CommonToolArguments.applyArgumentStrings].
```
agluszak, line 27: this entire class will be removed when 2.4.0 is released as btapi will have the compatibility shims that are missing now

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/BtapiCompilerCache.kt
```diff
@@ -0,0 +1,62 @@
...
+ * Caches BtapiCompiler instances keyed by toolchain specification.
+ *
+ * A worker process may serve requests from multiple Kotlin toolchains, so compilers
+ * must be selected and cached per request rather than fixed at worker startup.
+ */
+@OptIn(ExperimentalBuildToolsApi::class)
+class BtapiCompilerCache : AutoCloseable {
+  private val compilers = ConcurrentHashMap<ToolchainSpec, BtapiCompiler>()
+
+  fun get(spec: ToolchainSpec): BtapiCompiler = compilers.computeIfAbsent(spec, ::createCompiler)
```
ywett02: nit: what about making it `operator`?

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/BtapiCompiler.kt
restingbull, line 1: BtapiCompiler seems entirely separate from the K2JVM version. How about it moves to a separate package and we can do a deprecate cycle on the K2JVM?

It would make the new api easier to understand.

kotlin/internal/jvm/jvm.bzl
```diff
@@ -530,17 +531,6 @@ DEPRECATED - please use `jar` and `srcjar` attributes.""",
 _kt_compiler_deps_aspect = aspect(
     implementation = _kt_compiler_deps_aspect_impl,
     attr_aspects = ["deps", "runtime_deps", "exports"],
-    attrs = {
-        "_jarjar": attr.label(
```
restingbull, line 534: You can still have plugins that compile against the non-namespaced version, which would still require jarjar

Pull this change out of the Btapi implementation -- let's just ship the btapi, with unittests, and then do the rules integration.

kotlin/internal/jvm/compile.bzl
```diff
@@ -213,7 +213,7 @@ def _new_plugins_from(targets):
         all_plugins[plugin.id] = plugin
 
     if plugins_without_phase:
-        fail("has plugin without a phase defined: %s" % cfgs_without_plugin)
```
restingbull: whoops. exhibit of why tests are important. :(

kotlin/internal/toolchains.bzl
```diff
@@ -188,6 +212,30 @@ _kt_toolchain = rule(
             `kt_abi_plugin_incompatible`""",
             default = False,
         ),
+        "internal_jdeps_gen": attr.label(
```
restingbull: This could be done before btapi. 

Also, should they be public?

kotlin/internal/toolchains.bzl
```diff
@@ -367,18 +418,19 @@ def define_kt_toolchain(
...
         visibility = ["//visibility:public"],
         jacocorunner = jacocorunner,
-        jvm_stdlibs = jvm_stdlibs if jvm_stdlibs != None else [
-            Label("//kotlin/compiler:annotations"),
-            Label("//kotlin/compiler:kotlin-stdlib"),
-        ],
-        jvm_runtime = jvm_runtime if jvm_runtime != None else [
-            Label("//kotlin/compiler:kotlin-stdlib"),
-        ],
+        build_tools_impl = build_tools_impl if build_tools_impl != None else Label("//kotlin/compiler:kotlin-build-tools-impl"),
```
restingbull: You do not need to add label here -- string will do fine. It's really on useful if you need to pin a target to a repository, and that should happen in the macro if necessary.

kotlin/internal/toolchains.bzl
```diff
@@ -121,17 +143,19 @@ _kt_toolchain = rule(
...
+            default = Label("//kotlin/compiler:kotlin-build-tools-impl"),
+        ),
         "debug": attr.string_list(
             doc = """Debugging tags passed to the builder. Two tags are supported. `timings` will cause the builder to
             print timing information. `trace` will cause the builder to print tracing messages. These tags can be
             enabled via the defines `kt_timings=1` and `kt_trace=1`. These can also be enabled on a per target bases by
             using `tags` attribute defined directly on the rules.""",
             allow_empty = True,
         ),
-        "experimental_build_tools_api": attr.bool(
```
restingbull, line 151: This needs a deprecation like the flag. Otherwise it's only half deprecated.

Makefile
restingbull, line 1: ... I'm honestly not sure why we still have this file.

src/main/kotlin/io/bazel/kotlin/builder/tasks/KotlinBuilder.kt
restingbull, line 1: It looks like most the new flags are "in addition to" -- this allows use to keep both pipelines intact while folks do testing and we can get a fairly smooth deprecation cycle. I believe it would fair simple to split the compilation in `Build.kt` and call KotlinJvmTaskExecutor or KotlinBtapiJvmTaskExecutor depending on the toolchain values.

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/BtapiCompiler.kt
agluszak: Done: https://github.com/bazel-contrib/rules_kotlin/pull/1534/changes

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/CompilationTask.kt
```diff
@@ -31,179 +30,28 @@ import io.bazel.kotlin.builder.utils.jars.SourceJarExtractor
...
-
-    if (options.isNotEmpty()) {
-      base64Encode("-P", *options) { enc ->
-        "plugin:${plugins.kapt.id}:apoptions=$enc"
-      }
-    }
-  }
-}
-
+@ExperimentalBuildToolsApi
```
ALikhachev: Should not it be `@OptIn` instead at the actual use place?

src/main/kotlin/io/bazel/kotlin/builder/tasks/BUILD.bazel
```diff
@@ -21,17 +21,18 @@ kt_bootstrap_library(
...
     deps = [
+        "//kotlin/compiler:kotlinx-coroutines-core-jvm",
         "//src/main/kotlin/io/bazel/kotlin/builder/toolchain",
         "//src/main/kotlin/io/bazel/kotlin/builder/utils",
         "//src/main/kotlin/io/bazel/kotlin/builder/utils/jars",
         "//src/main/kotlin/io/bazel/worker",
         "//src/main/protobuf:deps_java_proto",
         "//src/main/protobuf:kotlin_model_java_proto",
+        "@rules_kotlin_maven//:org_jetbrains_kotlin_kotlin_build_tools_api",
+        "@rules_kotlin_maven//:org_jetbrains_kotlin_kotlin_build_tools_compat",
```
ALikhachev: I'm not sure you need the compat artifact in the deps. It's only needed at runtime, but not at compile time

Makefile
agluszak: https://github.com/bazel-contrib/rules_kotlin/pull/1535

kotlin/internal/toolchains.bzl
agluszak: > Also, should they be public?

I think so. If you use a different btapi-impl jar, you should also use plugins matching that version

kotlin/internal/toolchains.bzl
agluszak: https://github.com/bazel-contrib/rules_kotlin/pull/1534 doesn't touch it

src/main/kotlin/io/bazel/kotlin/builder/tasks/KotlinBuilder.kt
agluszak: Done in https://github.com/bazel-contrib/rules_kotlin/pull/1534

PR #1466, opened 2026-03-03
[kmp] Outline of a roadmap

restingbull:


plans/kmp.md
cgruber, line 1: I've posted https://github.com/bazel-contrib/rules_kotlin/issues/1682 which has a slightly more articulated plan. I'm thinking let's iterate on the plan in that issue (and related sub-issues) until we agree on our forward plan, and then we can rationalize this document as a plan artifact.

Issue #1467, opened 2026-03-03
Associate internal symbols not visible to java sources when associate target has public only ABI jars


rbeazleyspot:
When a target associates with another target that is opted into public only abi jars (`experimental_use_abi_jars = True, experimental_treat_internal_as_private_in_abi_jars = True, experimental_remove_private_classes_in_abi_jars = True`) any java source compiled as part of the target cannot see any internal declarations, an example would be java generated using KAPT.

I have pushed a branch [here](https://github.com/rbeazleyspot/rules_kotlin/commit/acc8511ffdba1fdf0d0beb822a9bc1b755e16891) that restructures the Dagger example so there is a separate "wiring" target containing the dagger component and modules. It associates with the `coffee_base` target containing a mix of internal and public types.

This wont build as the java compile step cannot see the internal symbols even though its an associate.

It seams the fix is reasonably straight forward as you can see [here](https://github.com/rbeazleyspot/rules_kotlin/pull/1/changes) where there is a specific list of deps for the compilation task containing full class jars. Rather than just changing dep_infos as these are used for normal java info output and we dont want to undo the strict public abi jar advantages (the filtering logic of abi jars is also cleaner).

Would this be a legitimate change?

PR #1468, opened 2026-03-03
Provide associate jars when compile java sources in _run_kt_java_builder_actions

fredm-spotify:
This attempts to fix an issue when Kapt generates Java code that requires access to internal classes that are stripped from ABI jars. @rbeazleyspot has opened an issue https://github.com/bazelbuild/rules_kotlin/issues/1467 with much more detail and another approach at fixing it. This is a simpler way to go about it and we wanted your advice on these.

When `experimental_remove_private_classes_in_abi_jars` is enabled `associate[JavaInfo].compile_jars` doesn't contain internal symbols. However, Kapt might generate code that requires access to these. An example is the Dagger generated code. 

If a target `B` associates with a target `A` and `B` is generating Dagger code, the call to `java_common.compile` in `_run_kt_java_builder_actions` doesn't seem to have access to the internal symbols from the associates array. Hence, compilation of the Java code fails.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -976,12 +976,24 @@ def _run_kt_java_builder_actions(
         # annotation processors in `deps` also.
         if len(srcs.kt) > 0:
             javac_opts.append("-proc:none")
+
+        # When experimental_remove_private_classes_in_abi_jars is enabled, associate targets'
+        # JavaInfo compile_jars point to ABI jars with internal classes stripped. The Kotlin
+        # compiler uses compile_deps.compile_jars (which swaps ABI jars for full class jars),
+        # but java_common.compile derives its classpath from JavaInfo.compile_jars in deps.
+        # We need to add the associate full jars explicitly so javac can see internal classes
+        # referenced by generated code (e.g., Dagger components).
+        associate_java_infos = [
+            JavaInfo(output_jar = jar, compile_jar = jar, neverlink = True)
+            for jar in compile_deps.associate_jars.to_list()
+        ]
+
         java_info = java_common.compile(
             ctx,
             source_files = srcs.java,
             source_jars = generated_kapt_src_jars + srcs.src_jars + generated_ksp_src_jars,
             output = ctx.actions.declare_file(ctx.label.name + "-java.jar"),
-            deps = compile_deps.deps + kt_stubs_for_java + [p[JavaInfo] for p in ctx.attr.plugins if JavaInfo in p],
+            deps = compile_deps.deps + kt_stubs_for_java + [p[JavaInfo] for p in ctx.attr.plugins if JavaInfo in p] + associate_java_infos,
             java_toolchain = toolchains.java,
             plugins = _plugin_mappers.targets_to_annotation_processors_java_plugin_info(ctx.attr.plugins),
             javac_opts = javac_opts,
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -56,6 +56,8 @@ use_repo(
     "kotlinx_serialization_json_jvm",
 )
 
+register_toolchains("//src/test/starlark/compile:abi_stripping_test_toolchain")
+
 register_toolchains("//src/main/starlark/core/compile/cli")
 
 register_toolchains("//kotlin/internal:default_toolchain")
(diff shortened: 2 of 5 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -976,12 +976,24 @@ def _run_kt_java_builder_actions(
...
+            for jar in compile_deps.associate_jars.to_list()
+        ]
+
         java_info = java_common.compile(
             ctx,
             source_files = srcs.java,
             source_jars = generated_kapt_src_jars + srcs.src_jars + generated_ksp_src_jars,
             output = ctx.actions.declare_file(ctx.label.name + "-java.jar"),
-            deps = compile_deps.deps + kt_stubs_for_java + [p[JavaInfo] for p in ctx.attr.plugins if JavaInfo in p],
+            deps = compile_deps.deps + kt_stubs_for_java + [p[JavaInfo] for p in ctx.attr.plugins if JavaInfo in p] + associate_java_infos,
```
restingbull, line 996: Hm. If I'm not wrong, this undoes the advantages of the ABI jars for associates.

fredm-spotify: oh no! My thinking is that since it's associated I should be able to access the internal symbols as a dependency to the compile task.

And I thought that whatever goes in `deps` doesn't end up in the final jar which would keep the internal symbols not-exposed.

rbeazleyspot: i guess we could flag this change behind the pair of `experimental_treat_internal_as_private_in_abi_jars, experimental_remove_private_classes_in_abi_jars` flags, but this concept (a modules java code should have access to its associates internals) is whats for discussion in https://github.com/bazelbuild/rules_kotlin/issues/1467

restingbull: Inputs are part of the cache key, so changes would invalidate on any associate change. It's not a whole tree invalidation (just immediate downstream actions), but it will have an impact.

I suspect we'd really want to extend abi here -- a change to a private field shouldn't invalidate. Hm.

fredm-spotify: Ok, one thing I'm not sure I follow. Let me come up with a specific example. I've hit this issue when I was compiling dagger in a bazel target called `wiring` which was associated with the target that had the implementation - `impl`.

The way I understand this, is that the dagger generated code placed in `wiring` might reference internal symbols from `impl`. shouldn't, in theory, a change to the internal code from `impl` actually invalidate the build from `wiring`?

rbeazleyspot: and extended abi output is interesting, an associates-abi....

before `experimental_treat_internal_as_private_in_abi_jars, experimental_remove_private_classes_in_abi_jars ` then this exact phenomena happened (almost) .....  a change to an internal symbol invalidated downstream actions..... if this is a mahoosive kapt generated application component that thats expensive  (as is our case)

 maybe we could flag this behind the pair of properties i mentioned and evolve the idea ?

rbeazleyspot: actually, thinking more, making the normal non-strict public abi jar would be all we need here as the other output

rbeazleyspot: ive got something sort of working where another action is added to generate the associates-abi.jar, let me come back to you on monday

fredm-spotify: sounds good! thank you

rbeazleyspot: hmmm, so i have tried this out... assuming i could use the skip code gen plugin to avoid a full recompilation for the associates-abi.jar. Looks like that plugin is only good for k1 and i couldnt see any low risk options for having something similar for k2. So I could either do a full compilation and discard the output class jar in order to get an associates-abi.jar or maybe there is some jar processing that could be done to get something that would work, but that dosnt sound too great either

rbeazleyspot: the other thing im trying is to actually use the abigen plugin twice, one with the flags to produce a public only abi jar and the other to make the "associates abi jar" (containing the internal symbols). Given you cant register the plugin twice im trying a thin wrapper that registers it under a different ID in order to configure it differently

rbeazleyspot: WDYT of https://github.com/bazelbuild/rules_kotlin/pull/1479 as a starter for ten @restingbull 

you might want a cup of tea and a biscuit to take you through it

fredm-spotify: @restingbull I'm closing this PR. I understand now that the jar I'm providing also include private symbols that once changed shouldn't invalidate the cache.

I've also gave a go at @rbeazleyspot's solution and it works within the Spotify project. I think we would prefer to go with his approach. Thanks for all the help!

PR #1470, opened 2026-03-04
[maintenance] Use rules_jvm_external for getting dependencies

agluszak:
Stop relying on the downloaded kotlinc distribution - switching to rules_jvm_external will make it easier to use BTAPI. Not having to resolve dependencies manually will also let us avoid problems like KSP requiring a specific version of kotlin coroutines.


MODULE.bazel
```diff
@@ -92,43 +127,29 @@ maven.install(
...
     fail_if_repin_required = True,
     fetch_sources = True,
-    lock_file = "//:kotlin_rules_maven_install.json",
+    lock_file = "//:rules_kotlin_maven_dev_install.json",
     repositories = [
         "https://maven-central.storage.googleapis.com/repos/central/data/",
         "https://maven.google.com",
         "https://repo1.maven.org/maven2",
     ],
+    resolver = "maven",
```
Bencodes, line 145: Why did you switch to this resolver specifically?

agluszak: it doesn't do any resolution at runtime and we use it internally at JB

kotlin/config.bzl
```diff
@@ -0,0 +1,62 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@rules_jvm_external//:defs.bzl", "maven_install")
+load("@rules_jvm_external//:repositories.bzl", "rules_jvm_external_deps")
+load("@rules_jvm_external//:setup.bzl", "rules_jvm_external_setup")
+
+def kt_configure():
+    """Setup dependencies."""
+    rules_jvm_external_deps()
```
Bencodes: This looks incorrect and meant for workspace setup which was removed in the last PR.

PR #1478, opened 2026-03-09
feat: allow passing KSP options via kt_ksp_plugin

monosoul:
Adds `options` arg to `kt_ksp_plugin`, so that end users can pass options to KSP processors from the build files and use those in codegen.

### DIFF
--- a/src/test/starlark/ksp/BUILD.bazel
+++ b/src/test/starlark/ksp/BUILD.bazel
@@ -87,7 +87,9 @@ kt_jvm_library(
     ],
 )
 
-# Target under test - library with two plugins that have the same option key
+# Target under test - library with two plugins that have the same option key.
+# Tagged manual so it is not analyzed by wildcard builds; the analysis test
+# below pulls it in via target_under_test.
 kt_jvm_library(
     name = "ksp_test_lib_with_conflicting_options",
     srcs = ["TestModel.kt"],
@@ -95,6 +97,7 @@ kt_jvm_library(
         ":moshi_plugin_with_option_a",
         ":moshi_plugin_with_option_b",
     ],
+    tags = ["manual"],
     deps = [
         "@kotlin_rules_maven//:com_squareup_moshi_moshi",
         "@kotlin_rules_maven//:com_squareup_moshi_moshi_kotlin",


kotlin/internal/jvm/plugins.bzl
```diff
@@ -60,6 +60,13 @@ def _targets_to_ksp_annotation_processors(targets):
 def _targets_to_annotation_processors_java_plugin_info(targets):
     return [t[JavaPluginInfo] for t in targets if JavaPluginInfo in t]
 
+def _targets_to_ksp_options(targets):
+    options = {}
+    for t in targets:
+        if _KspPluginInfo in t:
+            options.update(t[_KspPluginInfo].options)
```
restingbull: This will silently overwrite options that conflict with other plugins.

src/test/starlark/ksp/ksp_test.bzl
```diff
@@ -103,6 +104,87 @@ def _ksp_single_action_test_impl(ctx):
...
+    actions = analysistest.target_actions(env)
+    ksp2_actions = [a for a in actions if a.mnemonic == "KotlinKsp2"]
+
+    asserts.true(
+        env,
+        len(ksp2_actions) > 0,
+        "Should have at least one KotlinKsp2 action when plugin has options",
+    )
+
+    if ksp2_actions:
```
Bencodes: Is this conditional actually needed since the assertion above this already checks for the existence of a KSP action?

kotlin/internal/jvm/plugins.bzl
monosoul: That's a good point. Tbh, probably it makes more sense to have those options at the plugin consumer targets rather than `kt_ksp_plugin`. Otherwise we'll have to either pick the behavior for the end users (i.e. either merge the options or throw an error on conflict), or provide a config option for the merge strategy. If we go for the latter, probably such option would have to live at plugin consumer targets (`kt_jvm_library`, `kt_jvm_test`). And at that point I think it makes more sense to simply move the options there.

monosoul: What do you say about this: c90f0fbd6f68e96aada64ede411c2320108ca192 ?

src/test/starlark/ksp/ksp_test.bzl
monosoul: Good catch, yeah, that was unnecessary

kotlin/internal/jvm/plugins.bzl
monosoul: On the other hand, kinda weird to have plugin-specific options on `kt_jvm_library` 🤔  If that doesn't work, then I'd prefer merging the options, as it was, just with explicit mention of that behavior in the docs.

agluszak: Yeah, it looks kinda weird. Would there be any practical use case for it?

monosoul: Yeah, in my case I need to pass a list of service names to be used in codegen. The same service names are passed to some other bazel rules, so I'd like to have a single place where I declare those.

restingbull: Let's go with fail for now. Principal of least surprise and all.

monosoul: @restingbull b7772e19e647e4c333be42e5c036ff3a11d8382c

PR #1498, opened 2026-03-14
Include native library paths in kt_ targets

### LINKED ISSUE #1088: kt_jvm_test does not include native library paths in java.library.path
`java_test` collects the paths to native libraries from [`JavaInfo.transitive_native_libraries`](https://bazel.build/versions/6.0.0/rules/lib/JavaInfo#transitive_native_libraries) and specifies them in the `java.library.path` property. `kt_jvm_test` does not do this.

The current workaround if loading native libraries is to use a combination of `java_test` and `kt_jvm_library` instead. e.g.

```bazel
kt_jvm_library(
    name = "my_test",
    srcs = ["MyTest.kt"],
)

java_test(
    name = "MyTest",
    test_class = "package.for.MyTest",
    runtime_deps = [":my_test"],
)
```

Bencodes:
Fixes #1088.

Kotlin tests and binaries can now load JNI dependencies declared in `deps`, `runtime_deps`, or `exports`, including dependencies forwarded through Java libraries, Kotlin imports, and source-free Kotlin libraries. Previously a direct native runtime dependency failed analysis with `NoneType, want JavaInfo`; dependencies reaching the launcher failed with `UnsatisfiedLinkError`.

Propagate native dependencies using `JavaInfo.native_libraries`, select runtime shared objects, and include those objects in runfiles. An aspect follows Java/Kotlin dependency edges to recover DLL outputs that Bazel 8 omits from Windows CcInfo. Unix launchers resolve the native paths through their runfiles lookup. On Windows, a small native helper resolves DLL runfiles before invoking the real JVM; Bazel's existing launcher still handles classpaths, coverage, and JVM arguments. User-supplied `java.library.path` flags retain precedence.

The JNI regression suite performs actual native calls. The new `examples/native_libraries` module covers an external native dependency, use of the Kotlin binary as a build tool, another working directory, argument quoting, and a copied Windows executable with only a runfiles manifest. Native-helper tests cover both platform separators, paths containing spaces, deduplication, missing files, launch-data parsing, and Windows CRT quoting.

Validation:
- macOS: 44 focused JNI, launcher, JVM analysis, and documentation tests passed with Bazel 8.4.2; the earlier focused suite also passed with Bazel 9.2.0.
- The packaged native example passed on all three configured Bazel versions, including its runtime tests, build-tool action, and coverage run.
- Buildkite #6730: Windows, macOS, Linux, remote execution, documentation, formatting, and Kotlin lint checks passed. All Windows integration shards passed, including the new JNI example on Bazel 7, 8, and 9.
- All 18 steps in [Buildkite #6730](https://buildkite.com/bazel/rules-kotlin-kotlin/builds/6730) passed, including every Linux and Windows integration shard and the existing Android, WORKSPACE, and worker examples.


examples/nested_module_resources/BUILD.bazel
```diff
@@ -75,3 +76,17 @@ kt_jvm_test(
         "@nested//:printer",
     ],
 )
+
+cc_binary(
```
Bencodes: May not be the best place to include this test in the long run, but added it to the module resources sample since these tests provide some assertions around how non-JVM things get linked up into the kt_ targets at runtime.

restingbull (approved):

PR #1530, opened 2026-03-19
Per-class tracking to achieve higher level of compilation avoidance

oliviernotteghem:
This change adds support for per-class compilation avoidance, enabling fine-grained dependency tracking during Kotlin compilation. When enabled, the jdeps output records exactly
   which classes from each dependency JAR are used, along with SHA-256 hashes of their bytecode.

  Motivation

  Traditional compilation avoidance works at the JAR level—if any file in a dependency JAR changes, all dependents must recompile. Per-class tracking improves this by allowing
  build systems to skip recompilation when a dependency JAR changes but the specific classes actually used remain unchanged (same hashes).

  Configuration

  Two new toolchain attributes:

  define_kt_toolchain(
      name = "my_toolchain",
      experimental_track_class_usage = "on",    # Track class usage with hashes
      experimental_track_resource_usage = "on", # Track Android R class references
  )

  Changes

  Proto definitions:
  - deps.proto: Added UsedClass message with fullyQualifiedName, internalPath, and hash fields; added usedResources field to Dependencies
  - kotlin_model.proto: Added track_class_usage and track_resource_usage fields

  JDeps plugin:
  - Added configuration keys and CLI options for the new tracking flags
  - BaseJdepsGenExtension: Added getHashFromJarEntry() to compute SHA-256 hashes; updated doWriteJdeps() to write UsedClass entries when tracking is enabled
  - JdepsGenExtension (K1): Added resource tracking for Android R class references
  - K2 extensions: Added usedResources support

  Builder:
  - Added TRACK_CLASS_USAGE and TRACK_RESOURCE_USAGE flags to KotlinBuilder
  - Pass tracking flags to jdeps plugin in KotlinJvmTaskExecutor

  Starlark rules:
  - toolchains.bzl: Added experimental_track_class_usage and experimental_track_resource_usage attributes
  - compile.bzl: Pass tracking flags to builder

  Output Format

  When enabled, jdeps includes per-class entries:

  dependency {
    path: "bazel-out/.../libparser.jar" kind: EXPLICIT usedClasses { fullyQualifiedName: "com.lib.Parser" internalPath: "com/lib/Parser.class"
      hash: <32 bytes SHA-256>
    } }

  Testing

  Added KotlinBuilderJvmClassTrackingTest with tests verifying:
  - Tracking is disabled by default (no UsedClass entries)
  - Tracking records classes with hashes when enabled
  - Multiple classes are tracked correctly
  - Supertype dependencies are included
  - Hashes are deterministic across compilations

### DIFF
--- a/kotlin/internal/toolchains.bzl
+++ b/kotlin/internal/toolchains.bzl
@@ -90,6 +90,8 @@ def _kotlin_toolchain_impl(ctx):
         experimental_strict_kotlin_deps = ctx.attr.experimental_strict_kotlin_deps,
         experimental_report_unused_deps = ctx.attr.experimental_report_unused_deps,
         experimental_reduce_classpath_mode = ctx.attr.experimental_reduce_classpath_mode,
+        experimental_track_class_usage = ctx.attr.experimental_track_class_usage,
+        experimental_track_resource_usage = ctx.attr.experimental_track_resource_usage,
         experimental_build_tools_api = ctx.attr.experimental_build_tools_api,
         javac_options = ctx.attr.javac_options[JavacOptions] if ctx.attr.javac_options else None,
         kotlinc_options = ctx.attr.kotlinc_options[KotlincOptions] if ctx.attr.kotlinc_options else None,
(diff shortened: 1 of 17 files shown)


src/main/protobuf/deps.proto
```diff
@@ -48,6 +48,22 @@ message Dependency {
 
   // Source file locations: compilers can pinpoint the uses of a dependency.
   repeated SourceLocation location = 3;
+
+  // List of used classes from this dependency (for per-class compilation avoidance).
+  repeated UsedClass usedClasses = 4;
```
agluszak: this is a divergence vs [Bazel jdeps](https://github.com/bazelbuild/bazel/blob/92ec6cb38da44a5ffe1214c449e25ed6a9ad392c/src/main/protobuf/deps.proto#L4)

kotlin/internal/toolchains.bzl
```diff
@@ -150,6 +152,27 @@ _kt_toolchain = rule(
                 "KOTLINBUILDER_REDUCED",
             ],
         ),
+        "experimental_track_class_usage": attr.string(
+            doc = """Enable per-class compilation avoidance. Records which specific classes
```
agluszak: turning this on doesn't really enable (as in: cause) any compilation avoidance, right?

oliviernotteghem: @agluszak : you're absolutely right. To enable consumption of the class usage data and enable more granular compilation avoidance you need also this change on Bazel core https://github.com/bazelbuild/bazel/pull/16457

src/main/protobuf/deps.proto
oliviernotteghem: See my comment below (where the same proto changes are done in the mentioned PR).

jbarr21:
@restingbull can you please review this PR? if you are willing to accept this change, i can address your review feedback and rebase on latest main branch.

jbarr21:
any interest here?

PR #1547, opened 2026-04-02
Add path-mapping multiplex sandboxing support to Kotlin worker

mauriciogg:
Add path-mapping and multiplex sandboxing support to Kotlin worker

Adds `experimental_path_mapping` `experimental_multiplex_sandboxing` attribute to the Kotlin toolchain, which sets supports-multiplex-sandboxing in execution requirements and added support in the worker to used the sandbox dir from the work request.. 

Path mapping requires `multiplex_sandboxing`


restingbull:
Please add an example of this enabled (we use those for integration tests.)

Bencodes:
> Please add an example of this enabled (we use those for integration tests.)

I think we should also enable multiplex workers by default at this point and remove the flag. And then let people disable this using the Bazel flags instead of toolchain level flags.

mauriciogg:
> Please add an example of this enabled (we use those for integration tests.)

Done.

mauriciogg:
> > Please add an example of this enabled (we use those for integration tests.)
> 
> I think we should also enable multiplex workers by default at this point and remove the flag. And then let people disable this using the Bazel flags instead of toolchain level flags.

Applied the changes from your PR https://github.com/bazel-contrib/rules_kotlin/pull/1549. So now multiplex_sandboxing and path_mapping are opt-in. I can merge this once your PR is merged

Bencodes:
@mauriciogg can you take a look at fixing the errors here?

mauriciogg:
> @mauriciogg can you take a look at fixing the errors here?

Done.

Bencodes (approved):

PR #1586, opened 2026-04-29
BTAPI-based incremental compilation

agluszak:


src/main/kotlin/io/bazel/kotlin/builder/tasks/KotlinBuilder.kt
```diff
@@ -100,18 +105,20 @@ class KotlinBuilder(
     taskContext: WorkerContext.TaskContext,
     args: List<String>,
   ): Int {
-    val (argMap, compileContext) = buildContext(taskContext, args)
+    var compileContext: CompilationTaskContext? = null
```
agluszak, line 108: I guess it can be restored to how it was before

PR #1601, opened 2026-05-26
Update KSP to 2.3.9

Bencodes:
Bumping rules_kotlin to the latest version.


Bencodes:
This release still appears to have memory leaks behind the scenes from those dispatchers not being disposed.

restingbull (approved):

restingbull:
> This release still appears to have memory leaks behind the scenes from those dispatchers not being disposed.

I guess we should kill the worker every x invocations?

PR #1612, opened 2026-06-10
Do not percent-encode '+' in classpath-jar manifest Class-Path entries

mgalindo-sc:
Add '+' to the allowed character class so it passes through literally.

The java stub template packs over-long classpaths into a classpath jar whose manifest Class-Path entries percent-encode any character outside `[-_.~/a-zA-Z0-9].` Bzlmod canonical repository names put '+' into every external-repo classpath entry, so those entries now contain "%2b" sequences. '+' is a valid URI path character and needs no encoding, and the "%2b" breaks consumers that pass classpath URLs through printf-style formatting

### DIFF
--- a/third_party/java_stub_template.txt
+++ b/third_party/java_stub_template.txt
@@ -318,14 +318,14 @@ function create_classpath_jar() {
   for path in ${CLASSPATH}; do
     # Loop through the characters of the path and convert characters that are
     # not alphanumeric nor -_.~/ to their 2-digit hexadecimal representation
-    if [[ ! $path =~ ^[-_.~/a-zA-Z0-9]*$ ]]; then
+    if [[ ! $path =~ ^[-_.~/a-zA-Z0-9+]*$ ]]; then
       local i c buff
       local converted_path=""
 
       for ((i=0; i<${#path}; i++)); do
         c=${path:$i:1}
         case ${c} in
-              [-_.~/a-zA-Z0-9] ) buff=${c} ;;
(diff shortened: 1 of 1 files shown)


Bencodes:
Since this template is just copied from rules_java should this just get upstreamed?

mgalindo-sc:
> Add '+' to the allowed character class so it passes through literally.
> 
> The java stub template packs over-long classpaths into a classpath jar whose manifest Class-Path entries percent-encode any character outside `[-_.~/a-zA-Z0-9].` Bzlmod canonical repository names put '+' into every external-repo classpath entry, so those entries now contain "%2b" sequences. '+' is a valid URI path character and needs no encoding, and the "%2b" breaks consumers that pass classpath URLs through printf-style formatting

Change in rules java: https://github.com/bazelbuild/rules_java/pull/358. Should this repo point to `@bazel_tools//tools/java:java_stub_template` ?

Bencodes:
I believe that Bazel is looking to remove it from bazel_tools and instead have it hosted in rules_java and/or people just copy it.

Bencodes (approved):

mgalindo-sc:
> bazel_tools//tools/java

Yeah looks like the one under `bazel_tools` is just a reference to the one in `rules_java` but visibility there is restricted to `bazel_tools` only https://github.com/bazelbuild/rules_java/blob/master/java/bazel/rules/BUILD.bazel#L65-L67.

Bencodes:
This made it into the upstream rules_java https://github.com/bazelbuild/rules_java/pull/358

PR #1629, opened 2026-06-15
Bcr publish workflow

Bencodes:
Attempting to get BCR distributions working again. The bazel-contrib has an org level key for `BCR_PUBLISH_TOKEN` that rules_kotlin has access to.


restingbull (approved):

restingbull:
Thanks for keeping up on this.

PR #1639, opened 2026-06-19
Drop explicit worker strategy in .bazelrc

Bencodes:


.bazelrc
```diff
@@ -9,7 +9,6 @@ common --incompatible_autoload_externally=
 common --java_runtime_version=remotejdk_17
 common --tool_java_runtime_version=remotejdk_17
 
-build --strategy=KotlinCompile=worker
```
Bencodes, line 12: Workers are enabled by default here

restingbull (approved):

PR #1640, opened 2026-06-19
Cache URLClassLoader in KotlinKsp2 worker actions

Bencodes:
Potential work around for this https://github.com/google/ksp/issues/2959 otherwise KotlinKsp2 actions just eventually OOM when building in a large project.

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/Ksp2Task.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/Ksp2Task.kt
@@ -30,6 +30,7 @@ import java.nio.charset.StandardCharsets
 import java.nio.file.FileSystems
 import java.nio.file.Files
 import java.nio.file.Path
+import java.util.concurrent.ConcurrentHashMap
 import java.util.jar.JarEntry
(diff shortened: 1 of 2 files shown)


restingbull (approved):

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/Ksp2Task.kt
```diff
@@ -51,6 +52,11 @@ class Ksp2Task : Work {
   companion object {
     private val FLAGFILE_RE = Pattern.compile("""^--flagfile=((.*)-(\d+).params)$""").toRegex()
 
+    // Work around https://github.com/google/ksp/issues/2959: KSP2 creates a large
+    // processor classloader, and persistent multiplex workers otherwise retain one
+    // per work request until the worker JVM exits. Cache one loader per processor classpath.
+    private val kspClassLoaderCache = ConcurrentHashMap<String, URLClassLoader>()
```
Bencodes, line 58: Might need to re-think this at some point if builds are being done with a wide range of KotlinKsp2 plugins that hit this.

PR #1649, opened 2026-06-30
Inherit module_name from associates for export-only targets

eugenezh:
For targets without sources derive module name from the declared associates/exports, if any. This way the derived module name stays consistent across all associated parts.


restingbull (approved):

Bencodes:
@eugenezh can you rebase this branch so that the status checks pass?

PR #1650, opened 2026-06-30
Ensure "target" and "source" compiler options are honored by KAPT.

eugenezh:
Depending on javac version, these flags might have several spellings, either with one dash "-" or two dashes "--". If several spellings for a flag are available, one of them is considered "primary". KAPT passes flags to the compiler directly in an argument map, bypassing the CLI arguments parser, which parses and validates passed arguments. Compiler reads flags from the argument map using only "primary" flag name as a key. If KAPT passes only "-target" flag name in the argument map, but compiler uses "--target" name as a "primary" one, the option is effectively ignored. The fix is to pass "source" and "target" flags using all their possible spellings. This ensures the compiler will honor these options, no matter which flag name spelling is considered "primary" in this compiler version.


Bencodes (approved):

restingbull (approved):

Bencodes:
@eugenezh can you rebase this branch so that the status checks pass?

PR #1663, opened 2026-07-14
Fix broken compilation with BTAPI and add tests

eugenezh:
- Add missing dependency with `org.jetbrains.kotlin.daemon.common` package required for BTAPI infrastructure
- Explicitly specify sources and destination directory to `BuildToolsAPICompiler`, because this data is not parsed automatically from the arguments list.
- Added `KotlinBuilderJvmBtaTest` and `useBuildToolsApi()` flag to test framework to enable compilation with BTAPI in tests. In `KotlinAbstractTestBuilder` ensure that CompilationTaskInfo instance returned by CompilationTaskContext is the same as the JvmCompilationTask uses.


Bencodes (approved):

src/main/kotlin/io/bazel/kotlin/compiler/BuildToolsAPICompiler.kt
```diff
@@ -34,19 +35,41 @@ class BuildToolsAPICompiler {
 
     val kotlinToolchains = KotlinToolchains.loadImplementation(this.javaClass.classLoader!!)
 
-    // Create compilation operation with empty sources and dummy destination
-    // (the actual sources/destination will be set via applyArgumentStrings)
+    // BTAPI handles .kt/.java sources and destination arguments separately
+    val sources = mutableListOf<Path>()
+    var destination = Paths.get(".")
```
Bencodes: This wouldn't be needed if we were to instead update the `compile` function here to accept at least the sources when it gets called from `compileKotlin` in CompilationTask. It already has the sources defined there via `inputs.javaSourcesList` and `inputs.kotlinSourcesList`.

eugenezh: Yes, good point. We can also pass the destination directory in a dedicated parameter to avoid its extraction from the args as well

eugenezh:
Now passing sources and destination explicitly

Bencodes (approved):

PR #1666, opened 2026-07-19
Avoid loading Kotlin compiler metadata for unrelated toolchains

dzbarsky:
Registered `toolchain()` targets are loaded while Bazel resolves any toolchain type, so the Kotlin default toolchain currently loads Kotlin compiler metadata even for builds that do not use Kotlin. Move the Kotlin toolchain type and default `toolchain()` declaration into a load-free package, retain the existing toolchain-type label as a public alias, include the package in the release archive, and mark the source-only CLI bootstrap toolchain as a dev dependency. Add an integration example that makes compiler metadata fail immediately if an unrelated toolchain resolves it, and include the example in the repository's integration-test package list.

Validation:

- `bazel --output_base=/private/tmp/rules-kotlin-lazy-default-toolchain-output-base build --jobs=4 --noshow_progress --color=no //:unrelated` from `examples/lazy_default_toolchain` passes.
- `buildifier -mode=check` passes for the changed Starlark, BUILD, and MODULE files; `git diff upstream/master...HEAD --check` passes.
- Verified the release archive contains `kotlin/internal/default_toolchain/BUILD.bazel`. The local integration build was interrupted during Kotlin compilation, and the remote integration build is blocked by the unrelated `remote_java_tools_darwin_arm64/.../ijar: Exec format error`.


restingbull:
What is the use case being solved here? e.g. what does loading "Kotlin compiler metadata" result in?

dzbarsky:
> What is the use case being solved here? e.g. what does loading "Kotlin compiler metadata" result in?

It refers to the eager .bzl load. But more to the point, right now every bazel repo in the world eagerly downloads the kotlin compiler jars due to these eager loads, even when not using kotlin. This PR fixes it.

restingbull:
> > What is the use case being solved here? e.g. what does loading "Kotlin compiler metadata" result in?
> 
> It refers to the eager .bzl load. But more to the point, right now every bazel repo in the world eagerly downloads the kotlin compiler jars due to these eager loads, even when not using kotlin. This PR fixes it.

> kotlin compiler jars due to these eager loads

That should not be the case. (Might be related to the move of the jars to /kotlin/compiler`, have to look -- there is a distinct separation)

That aside,  how does moving the toolchain and type to a separate package change the behavior?

dzbarsky:
> > > What is the use case being solved here? e.g. what does loading "Kotlin compiler metadata" result in?
> > 
> > 
> > It refers to the eager .bzl load. But more to the point, right now every bazel repo in the world eagerly downloads the kotlin compiler jars due to these eager loads, even when not using kotlin. This PR fixes it.
> 
> > kotlin compiler jars due to these eager loads
> 
> That should not be the case. (Might be related to the move of the jars to /kotlin/compiler`, have to look -- there is a distinct separation)
> 
> That aside, how does moving the toolchain and type to a separate package change the behavior?

Because that's how bazel works :) It eagerly loads any repo referenced `.bzl`s that are `load`'ed, but the labels referred to by a `toolchain` rule are not eager, they are lazy until the toolchain is actually selected for execution.

You're welcome to play around with https://github.com/hermeticbuild/hermetic-llvm, you'll notice there is no Kotlin in sight but there are jars downloaded. If you want to pursue another solution that also avoids these downloads, I'm cool with that! I just want this fixed.

restingbull:
> Because that's how bazel works :) It eagerly loads any repo referenced .bzls that are load'ed, but the labels referred to by a toolchain rule are not eager, they are lazy until the toolchain is actually selected for execution.

That implies that all .bzl files in a packaged are loaded by simply referencing the package. Surprising behavior.

I'm not being obstinate, just need to understand the why -- so far, this is long on how and short on why.

dzbarsky:
> > Because that's how bazel works :) It eagerly loads any repo referenced .bzls that are load'ed, but the labels referred to by a toolchain rule are not eager, they are lazy until the toolchain is actually selected for execution.
> 
> That implies that all .bzl files in a packaged are loaded by simply referencing the package. Surprising behavior.

Yes! That's exactly what happens. It's a bit unintuitive, but it's actually required for correctness. Consider the following:

`foo.bzl`:
```
def foo(name):
    filegroup(
        name = "bar", # SUPRISE!
        srcs = [],
    )
```

`bar.bzl`:
```
def bar(name):
    filegroup(
        name = "foo", # SUPRISE!
        srcs = [],
    )
```

`BUILD.bazel`:
```
load(":foo.bzl", "foo")
load(":bar.bzl", "bar")

foo(name = "foo")
bar(name = "bar")
```

When you do `bazel build //:foo`, how could Bazel do the right thing without loading all `.bzl` file referenced from the package in order to evaluate them? Essentially bazel has no idea if a symbol is a rule or a macro (which can define whatever targets it wants!). And then consider also the existence of `native.existing_rules()` and finalizer macros, etc. Sadly BUILD.bazel is not nearly as imperative as it would seem at first sight!

> I'm not being obstinate, just need to understand the why -- so far, this is long on how and short on why.

Hopefully the above explains why? [Here's a blog post explaining this, see the section around BUILD fetches](https://aspect.build/blog/avoid-eager-fetches).

restingbull:
Ah, I'd mentally dropped that bit. Nothing like being rusty.

Doesn't explain the GitHub download, though -- the references to that do not have any load statements. It's entirely aliases. Which argues that alias is special. Easy enough to test.

dzbarsky:
> Ah, I'd mentally dropped that bit. Nothing like being rusty.
> 
> Doesn't explain the GitHub download, though -- the references to that do not have any load statements. It's entirely aliases. Which argues that alias is special. Easy enough to test.

Oh I think I interpreted the `@com_github_jetbrains_kotlin` repo as the compiler itself, but looks like those JARs are only under `@com_github_jetbrains_kotlin_git` which is lazy, aliases are fine. That's confusing naming IMO but it's my bad, I should have double-checked more carefully. So I think only the `@com_github_jetbrains_kotlin` repo was eagerly evaluated. Perhaps it should be renamed `@kotlin_capabilities` or similar to make it clearer (but maybe that's annoying with WORKSPACE, and anyway should be a separate change, will defer that to you!)

restingbull:
> > Ah, I'd mentally dropped that bit. Nothing like being rusty.
> > Doesn't explain the GitHub download, though -- the references to that do not have any load statements. It's entirely aliases. Which argues that alias is special. Easy enough to test.
> 
> Oh I think I interpreted the `@com_github_jetbrains_kotlin` repo as the compiler itself, but looks like those JARs are only under `@com_github_jetbrains_kotlin_git` which is lazy, aliases are fine. That's confusing naming IMO but it's my bad, I should have double-checked more carefully. So I think only the `@com_github_jetbrains_kotlin` repo was eagerly evaluated. Perhaps it should be renamed `@kotlin_capabilities` or similar to make it clearer (but maybe that's annoying with WORKSPACE, and anyway should be a separate change, will defer that to you!)

I am back to trying to figure out why the jars are being downloaded, though.

Partially related, I'm getting an error from `hermetic-llvm`:
```
ERROR:  Evaluation of query "allpaths(//..., "@@rules_kotlin++rules_kotlin_extensions+com_github_jetbrains_kotlin_git//...")" failed: preloading transitive closure failed: no such package '@@[unknown repo 'linux_kernel_headers_riscv.3.0.101' requested from @@+kernel_headers+kernel_headers (did you mean 'linux_kernel_headers_riscv.7.0.10'?)]//': The repository '@@[unknown repo 'linux_kernel_headers_riscv.3.0.101' requested from @@+kernel_headers+kernel_headers (did you mean 'linux_kernel_headers_riscv.7.0.10'?)]' could not be resolved: No repository visible as '@linux_kernel_headers_riscv.3.0.101' from repository '@@+kernel_headers+kernel_headers'
```

restingbull:
> Registered toolchain() targets are loaded while Bazel resolves any toolchain type, so the Kotlin default toolchain currently loads Kotlin compiler metadata even for builds that do not use Kotlin.

I'm having trouble reproduce this in [isolation](https://github.com/bazel-contrib/rules_kotlin/pull/1667) -- it doesn't not appear the toolchains outside of the directly referenced ones are loaded. Do you have a reproduction of the behavior? 

(This is important -- if it isn't toolchain eagerly loading, the cause might be different than expected.)

ivangolub-oai (approved):

restingbull (changes requested):
Please provide a simple reproduction case of unwanted jar downloads by including `bazel_dep(name = "rules_kotlin", version = "2.4.0")` in the module or transitive closure.

PR #1667, opened 2026-07-21
[repo fetch] Validate toolchain is always loaded.

restingbull:


restingbull:
TLDR; the toolchain is not fetched.

PR #1672, opened 2026-07-23
Fingerprint KSP2 classloader cache by jar content, not just path

nikolai-manzhos:
Bazel keeps the same output path across incremental rebuilds of a KSP processor, so the path-only cache key can't tell a recompiled jar from an unchanged one. This causes the worker to reuse a stale classloader after a processor is edited and rebuilt.

Replace `computeIfAbsent` with `compute` and pair each cached classloader with a SHA-256 content fingerprint (path + size + bytes). When the fingerprint changes for a cached classpath, the stale classloader is closed and replaced. This keeps at most one entry per distinct classpath, so it can't regress the metaspace OOM the cache was originally added to fix.

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/Ksp2Task.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/Ksp2Task.kt
@@ -30,6 +30,8 @@ import java.nio.charset.StandardCharsets
 import java.nio.file.FileSystems
 import java.nio.file.Files
 import java.nio.file.Path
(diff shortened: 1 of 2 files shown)


Bencodes:
@nikolai-manzhos out of curiosity, did you run into any edge cases that justified this change?

nikolai-manzhos:
> @nikolai-manzhos out of curiosity, did you run into any edge cases that justified this change?

yes, while working on a ksp processor changes weren't applied until the worker was shutdown and re-started.

Bencodes (approved):

PR #1674, opened 2026-07-23
Drop support for custom pre-built release archives

Bencodes:
Consumers now compile the rules_kotlin tools from source. The release uses the same `MODULE.bazel` and build definitions as a development checkout, replacing the precompiled archive and its separate release-only BUILD/MODULE files.

### Changes

- Package the required sources with `rules_pkg`; omit compiled JARs, repository tests, examples, and development tooling.
- Keep consumer dependencies public and development dependencies private to root builds. Move test-only Maven artifacts and helpers out of the production dependency graph and repin both Maven installs.
- Test examples against the actual release archive, including `bazel query '@rules_kotlin//...'`, and have BCR presubmit build the entire published package.
- Update release automation and documentation for source releases. New releases require Bzlmod; legacy WORKSPACE setup is removed.
- Preserve the Build Tools API runtime and embeddable compiler plugins introduced by #1724, including their sources in the archive. Document the distinction between the single CLI compiler override and multiple named Build Tools API runtimes.

Rebased onto `origin/master` at `593bfa2c` (#1724), resolving conflicts in the removed release-only files, source BUILD files, repository initialization, and README. Updated the new runtime repository tests to load `initialize.bzl`.

### Release archive size

Measured the precompiled release built from `origin/master` (`593bfa2c`) against this PR's source release:

| | Before | After |
| --- | ---: | ---: |
| Compressed archive | 50,244,759 bytes (50.24 MB) | 158,260 bytes (158.26 kB) |
| Uncompressed file contents | 56,243,780 bytes | 703,156 bytes |
| Bundled JARs | 11 | 0 |

The compressed download is **99.685% smaller** (about **317× smaller**). Consumers compile the required tools instead of downloading bundled precompiled rules_kotlin tools.

### Validation after rebasing

- Passed 40 focused tests covering archive contents, Build Tools API release records and runtime selection, plugin dialects and reshading, worker runtime flags, builder execution, KAPT, and KSP.
- Passed the trivial and KSP example integration suites on Bazel 9.2.0. Both exercise the legacy and Build Tools API backends, example tests, coverage, and the whole-package query against the actual archive.
- Queried the entire final archive from the trivial example using `archive_override` in `MODULE.bazel`: 188 targets, no errors.
- Fully built `@rules_kotlin//...` from that external consumer on Bazel 9.2.0: 180 targets, including execution rather than only analysis.
- Rebuilt both archives for the size comparison and passed formatting/diff checks. The full expensive example/Bazel-version matrix was not run.


MODULE.bazel
```diff
@@ -60,73 +70,89 @@ register_toolchains("//src/main/starlark/core/compile/cli")
...
     name = "kotlin_rules_maven",
     artifacts = [
-        "com.google.code.findbugs:jsr305:3.0.2",
-        "com.google.errorprone:error_prone_annotations:2.23.0",
-        "junit:junit:4.13-beta-3",
-        "com.google.guava:guava:33.0.0-jre",
-        "com.google.truth:truth:0.45",
-        "com.google.auto.service:auto-service:1.1.1",
-        "com.google.auto.service:auto-service-annotations:1.1.1",
+        "com.eed3si9n.jarjar:jarjar:1.17.0",
```
Bencodes, line 81: There is a Bazel module for this that we can lean into

MODULE.bazel
```diff
@@ -60,73 +70,89 @@ register_toolchains("//src/main/starlark/core/compile/cli")
...
         "com.google.auto.value:auto-value:1.11.0",
```
Bencodes, line 82: If we drop Dagger we'd be mostly vanilla Kotlin here

MODULE.bazel
```diff
@@ -21,17 +21,27 @@ bazel_dep(name = "rules_android", version = "0.7.3")
...
     "remote_android_tools_extensions",
+    dev_dependency = True,
 )
 use_repo(remote_android_extensions, "android_tools")
 
-android_sdk_repository_extension = use_extension("@rules_android//rules/android_sdk_repository:rule.bzl", "android_sdk_repository_extension")
+android_sdk_repository_extension = use_extension(
+    "@rules_android//rules/android_sdk_repository:rule.bzl",
+    "android_sdk_repository_extension",
+    dev_dependency = True,
```
restingbull, line 31: Breaks querying `@rules_kotlin`.

src/main/starlark/core/compile/cli/BUILD.bazel
restingbull, line 1: This ships the cli rules as part of the release. Confusing to say the least.

BUILD
```diff
@@ -16,10 +14,55 @@ load("//kotlin:lint.bzl", "ktlint_config")
...
 # limitations under the License.
-load("//src/main/starlark/release:packager.bzl", "release_archive")
-
 package(default_applicable_licenses = [":license"])
 
+exports_files(["MODULE.bazel"])
+
+# Integration tests reference this repository with paths relative to their fixture workspace.
+# Keep the repository sources as declared test inputs so those paths resolve within runfiles.
+filegroup(
```
restingbull, line 23: Um, this is pretty horrible patch for integration tests. At the least the release archive creates a clean target instead of exporting... well, everything?

CONTRIBUTING.md
```diff
@@ -28,38 +28,17 @@ Otherwise, the Buildkite CI will yell at you about formatting/linting violations
...
-
-To cope with API churn, the release archive can configure different rule attributes for depending on the chosen kotlin version.
- Each major release of kotlin (1.4, 1.5) has a specific sub-repository under [src/main/starlark](src/main/starlark). The naming convention for these
-  is `rkt_<major>_<minor>` ([r]elease [k]o[t]lin).
-
-The version is selected by the [kotlin_repositories](src/main/starlark/repositories/initialize.release.bzl) rule during initialization. 
-New versions of kotlin that change the API should be added to [versions.bzl](src/main/starlark/repositories/versions.bzl), under `CORE` following the 
-existing naming convention.
-
-Multiple versions of kotlin are not currently handled.(_help wanted_)
```
restingbull, line 61: Still unhandled

MODULE.bazel
```diff
@@ -60,73 +70,89 @@ register_toolchains("//src/main/starlark/core/compile/cli")
...
-        "com.google.auto.service:auto-service-annotations:1.1.1",
+        "com.eed3si9n.jarjar:jarjar:1.17.0",
         "com.google.auto.value:auto-value:1.11.0",
         "com.google.auto.value:auto-value-annotations:1.11.0",
+        "com.google.code.findbugs:jsr305:3.0.2",
         "com.google.dagger:dagger:2.57.2",
         "com.google.dagger:dagger-compiler:2.57.2",
-        "com.google.dagger:dagger-producers:2.57.2",
+        "com.google.errorprone:error_prone_annotations:2.23.0",
+        "com.google.guava:guava:33.0.0-jre",
```
restingbull, line 88: Guava is a non-trivial dep, and can cause all sorted of havoc when two versions are present in a repo. There is no enforcement to separate multiple maven  dependency versions... which leaves a potentially nasty rake in the grass.

MODULE.bazel
```diff
@@ -60,73 +70,89 @@ register_toolchains("//src/main/starlark/core/compile/cli")
...
+    lock_file = "//:kotlin_rules_maven_install.json",
+    repositories = [
+        "https://maven-central.storage.googleapis.com/repos/central/data/",
+        "https://maven.google.com",
+        "https://repo1.maven.org/maven2",
+    ],
+)
+use_repo(maven, "kotlin_rules_maven", "unpinned_kotlin_rules_maven")
+
+dev_maven = use_extension(
```
restingbull, line 102: ... and now `bazel build @rules_kotlin//...` breaks as well.

Something that would become a standard target for CI to seed cache. Instead, there ends up being a deep link to the build target, extending the public api.

src/main/kotlin/io/bazel/kotlin/test/BazelIntegrationTestRunner.kt
```diff
@@ -27,30 +21,7 @@ object BazelIntegrationTestRunner {
...
-          entry.isFile -> Files.write(
-            destination.apply { parent.createDirectories() },
-            stream.readBytes(),
-          )
-
-          else -> throw NotImplementedError(entry.toString())
-        }
-      }
-    }
+    val rulesKotlin = workspace.resolve("../..").normalize().toAbsolutePath()
```
restingbull, line 24: A tar archive is much cleaner than adding stack of directories. It's almost RBE antagonistic.

src/main/kotlin/io/bazel/kotlin/test/BazelIntegrationTestRunner.kt
```diff
@@ -117,13 +88,18 @@ object BazelIntegrationTestRunner {
...
           "//...",
         ).onFailThrow()
-        bazel.run(
-          workspace,
-          *systemFlags,
-          "query",
-          *commandFlags,
-          "@rules_kotlin//...",
-        ).onFailThrow()
+        listOf(
```
restingbull, line 91: This reruns the unit and integration tests for every example and every bazel version.

src/main/kotlin/io/bazel/kotlin/test/BazelIntegrationTestRunner.kt
Bencodes: I agree

PR #1680, opened 2026-07-28
Support for conditional classpath reduction

Bencodes:
Adds a globally configurable repository allow-list for `experimental_prune_transitive_deps`.

When transitive pruning is enabled, rules_kotlin normally compiles against direct dependencies only. This change allows those repositories to retain a subset of their transitive jars without disabling pruning for the entire Kotlin target. This is mostly useful if you want to retain subsets of the dependency graph like maven dependencies that are less likely to cause cache invalidation, while stripping out more if the higher-churn jars that live directly in your repository.

Example `.bazelrc` configuration:

```bazelrc
build --@rules_kotlin//kotlin/settings:experimental_prune_transitive_deps=True
build --@rules_kotlin//kotlin/settings:experimental_prune_transitive_deps_keep_transitive_repositories=maven,rules_jvm_external++maven+maven
```

A real world example is when you enable classpath reduction, you are forced to declare every single dependency that you depend on both directly and transitively. If a single type is referenced at compile time or during annotation processing, that dependency needs to be included, even dependencies that 3rd party maven dependencies declare in their pom files. This change helps smooth that out by assuming that things within a certain namespace, like maven dependencies, should just be included and not stripped out saving developers from having to verbosely list out everything down to the entire transitive classpath at the maven level.

### DIFF
--- a/kotlin/internal/toolchains.bzl
+++ b/kotlin/internal/toolchains.bzl
@@ -101,6 +101,7 @@ def _kotlin_toolchain_impl(ctx):
         empty_jdeps = ctx.file._empty_jdeps,
         jacocorunner = ctx.attr.jacocorunner,
         experimental_prune_transitive_deps = ctx.attr._experimental_prune_transitive_deps[BuildSettingInfo].value,
+        experimental_prune_transitive_deps_keep_transitive_repositories = ctx.attr._experimental_prune_transitive_deps_keep_transitive_repositories[BuildSettingInfo].value,
         experimental_strict_associate_dependencies = ctx.attr._experimental_strict_associate_dependencies[BuildSettingInfo].value,
         experimental_ksp2_psi_resolution = ctx.attr._experimental_ksp2_psi_resolution[BuildSettingInfo].value,
     )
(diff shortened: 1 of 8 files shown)


kotlin/internal/toolchains.bzl
```diff
@@ -101,6 +101,7 @@ def _kotlin_toolchain_impl(ctx):
         empty_jdeps = ctx.file._empty_jdeps,
         jacocorunner = ctx.attr.jacocorunner,
         experimental_prune_transitive_deps = ctx.attr._experimental_prune_transitive_deps[BuildSettingInfo].value,
+        experimental_prune_transitive_deps_keep_transitive_repositories = ctx.attr._experimental_prune_transitive_deps_keep_transitive_repositories[BuildSettingInfo].value,
```
restingbull, line 104: Is the end state to have this always on the toolchain, or per target? I can see arguments for either or both.

Bencodes: Potentially yes since these flags are quite dependent on the user and how their project is set up.

restingbull (approved):

kotlin/internal/toolchains.bzl
Bencodes: @restingbull I added to the PR description to help illustrate where this is most-useful. We primarily have used this one internally to reduce the annoyances of classpath reduction on the developer experience and build file maintenance. This is less needed if we have a good solution for generating "correct" build files.

PR #1691, opened 2026-08-01
Don't pass the transitive compile-jar closure to JdepsMerge

snuderl:
## What this changes

`_run_merge_jdeps_action` declares the **full transitive compile-jar closure** of every dependency as
inputs to the `JdepsMerge` action whenever `experimental_report_unused_deps != "off"`. The merger only
opens the jars that the merged jdeps actually names, and those can only be jars that were on the Kotlin
compile classpath, so the closure is over-declared. This PR adds an optional `classpath_jars` argument and
passes `compile_deps.compile_jars` for Kotlin-only targets; targets that also run javac keep the old input
set, because javac compiles against the unpruned transitive classpath and its jdeps can name jars outside
the Kotlin classpath.

The win only materialises when `experimental_prune_transitive_deps` is on — otherwise the compile classpath
*is* the transitive closure and the input set is unchanged.

### Why it is safe

- `BaseJdepsGenExtension.doWriteJdeps` emits exactly two kinds of paths: the entries of
  `--direct_dependencies` (as `UNUSED`) and the jars classes were loaded from, both mapped through
  `--full_classpath`. `_run_kt_builder_action` sets `--direct_dependencies` from the direct deps' compile
  jars and `--classpath` from `compile_deps.compile_jars`; the former is a subset of the latter.
- `JdepsMerger.merge` reads `Target-Label` from the manifest of exactly the paths present in the merged
  jdeps. So every jar it opens is in `compile_deps.compile_jars`.
- If that reasoning were ever wrong, the action fails with a missing-input error — it cannot silently
  mis-report unused deps.

### Tests

`src/test/starlark/rules/jdeps_merge_inputs_tests.bzl` asserts the `JdepsMerge` input set with
`experimental_prune_transitive_deps` on, against a `main -> dependency_a -> transitive_dep` graph:

- `kotlin_only` — with unused-dep reporting on, inputs contain `dependency_a`'s compile jar but not the
  transitive one (fails without this change).
- `with_java_sources` — the same target with an added `.java` source still gets the transitive closure.
- `report_unused_deps_off` — no dep jars are declared at all.

Reporting is enabled by pointing `--extra_toolchains` at a toolchain declared in the test package, since
`experimental_report_unused_deps` is a toolchain attribute rather than a flag.

I also verified end to end in a scratch package with `experimental_report_unused_deps = "error"` and
`experimental_prune_transitive_deps = True` that an unused direct dep on a Kotlin-only `kt_jvm_library`
still fails the build with `** Please remove the following dependencies:`, that a target whose deps are all
used builds clean, and that both hold for a mixed Kotlin/Java target.

## Measurements

From a ~1,500 `kt_library` target Kotlin monorepo with pruning enabled.

Static input counts, `bazel aquery 'mnemonic("JdepsMerge", deps(<target>))'`:

| Subgraph | JdepsMerge actions | inputs before | inputs after | mean/action |
| --- | --- | --- | --- | --- |
| graphqlapi | 631 | 90,165 | 12,659 (−86%) | 142.9 → 20.1 |
| socialapi | 263 | 13,125 | 3,746 (−71%) | 49.9 → 14.2 |

Invalidation fan-in (for each first-party jar, how many JdepsMerge actions it invalidates when its ABI
changes): graphqlapi mean 58.2 → 8.5, socialapi 12.4 → 4.1. These targets do not recompile on such a
change (pruned classpath), only their merge action was being invalidated.

Cold remote-execution A/B on one subgraph (95 JdepsMerge executions in both arms): worker-fetched bytes
144.3 MB → 53.6 MB (−63%), files 454 → 289; per-action exec time unchanged (~0.25 s). The win is fetched
bytes, merkle-tree/digest work, cache-lookup payload and invalidation — not per-action CPU.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -379,7 +379,7 @@ def _build_resourcejar_action(ctx, toolchains, extra_resources = {}):
     )
     return resources_jar_output
 
-def _run_merge_jdeps_action(ctx, toolchains, jdeps, outputs, deps, associate_jars = depset(), classpath_jars = None):
+def _run_merge_jdeps_action(ctx, toolchains, jdeps, outputs, deps, classpath_jars, associate_jars = depset()):
     """Creates a Jdeps merger action invocation."""
     args = ctx.actions.args()
     args.set_param_file_format("multiline")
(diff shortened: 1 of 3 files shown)


restingbull (approved):

Bencodes:
@snuderl can you double check to see if this PR still works correctly for you internally? (I rebased it)

Bencodes (approved):

PR #1698, opened 2026-08-04
Split test deps into kotlin_rules_maven_test

Bencodes:
Splitting our test-specific dependencies out of the main maven repo where they can be better isolated by dev_dependency. There is some work coming down the pipeline relating to the Build Tool API changes that will benefit from the separation here.

There might be some more dependencies that we can pull out of the production maven repo to further slim that down, but we can do that incrementally.


MODULE.bazel
```diff
@@ -75,39 +75,28 @@ rbe.config(
...
+]
+
 maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
 maven.install(
     name = "kotlin_rules_maven",
     artifacts = [
         "com.google.code.findbugs:jsr305:3.0.2",
         "com.google.errorprone:error_prone_annotations:2.23.0",
-        "junit:junit:4.13-beta-3",
         "com.google.guava:guava:33.0.0-jre",
```
Bencodes, line 90: Guava is used by the Bazel worker APIs that we depend on

restingbull (approved):

PR #1702, opened 2026-08-07
Extend experimental_prune_transitive_deps semantics to the javac half…

eugenezh:
… of mixed targets

- When experimental_prune_transitive_deps is enabled, compile target's Java sources with the same classpath the Kotlin sources are compiled against.
- Tests added.

### DIFF
--- a/kotlin/internal/jvm/jvm_deps.bzl
+++ b/kotlin/internal/jvm/jvm_deps.bzl
@@ -35,9 +35,11 @@ def _jvm_deps(ctx, toolchains, associate_deps, deps = [], deps_java_infos = [], 
         [toolchains.kt.jvm_stdlibs]
     )
 
+    prune_transitive_deps = (toolchains.kt.experimental_prune_transitive_deps and
+                             not "kt_experimental_prune_transitive_deps_incompatible" in ctx.attr.tags)
+
     # Reduced classpath, exclude transitive deps from compilation
-    if (toolchains.kt.experimental_prune_transitive_deps and
-        not "kt_experimental_prune_transitive_deps_incompatible" in ctx.attr.tags):
+    if prune_transitive_deps:
         transitive_jars = []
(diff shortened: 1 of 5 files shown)


restingbull (approved):

kotlin/internal/jvm/jvm_deps.bzl
```diff
@@ -73,9 +75,27 @@ def _jvm_deps(ctx, toolchains, associate_deps, deps = [], deps_java_infos = [],
...
+            JavaInfo(output_jar = jar, compile_jar = jar, deps = [], neverlink = True)
+            for jar in compile_depset_list_filtered
+        ]
+    else:
+        pruned_deps_for_java = None
+
     return struct(
         module_name = associates.module_name,
         deps = dep_infos,
+        pruned_deps_for_java = pruned_deps_for_java,
```
Bencodes: You might as well set this to https://github.com/bazel-contrib/rules_kotlin/pull/1702/changes#diff-ca072bf9da461834e9b015042cb38ba7d06b7977889bc1956b4da376b9df2512R1075-R1077 and return the java deps instead of having to do the conditional checks down stream.

eugenezh: yes, makes sense. I've pushed the update: now all "if prune_transitive_deps" checks are encapsulated in one place

Bencodes (approved):

kotlin/internal/jvm/jvm_deps.bzl
```diff
@@ -35,9 +35,11 @@ def _jvm_deps(ctx, toolchains, associate_deps, deps = [], deps_java_infos = [],
         [toolchains.kt.jvm_stdlibs]
     )
 
+    prune_transitive_deps = (toolchains.kt.experimental_prune_transitive_deps and
```
Bencodes, line 38: Probably fine for now, but we _could_ add a different opt-out flag here so that this doesn't break peoples java builds randomly when they upgrade.

PR #1704, opened 2026-08-07
Wire the experimental_build_tools_api build setting into the toolchain

eugenezh:
- The //kotlin/settings:experimental_build_tools_api bool_flag was a no-op. Now the flag toggles compilation with Build Tools API for all kotlin toolchains.
- The toolchain now reads the setting via a BuildSettingInfo label attr; the effective toggle is the per-toolchain attr OR-ed with the build setting.
- Changed default value to 'False' making Build Tools API an explicit opt-in.
- README updated and tests added


restingbull (approved):
Technically, this moves us back from the BTAPI progress. On the other hand, it makes it clear there could be some migration issues fro folks.

@Bencodes -- we should start encouraging people to move after this lands.

PR #1707, opened 2026-08-10
Introduce 'no_proc' option to explicitly disable java annotation proc…

eugenezh:
…essing

This option allows to explicitly turn off annotation processing for kt_jvm_* targets containing only .java sources.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -1061,7 +1061,8 @@ def _run_kt_java_builder_actions(
     # the final ABI jar. Otherwise just use the KT ABI jar as final ABI jar.
     ksp_generated_java_src_jars = generated_ksp_src_jars and is_ksp_processor_generating_java(ctx.attr.plugins)
     if srcs.java or generated_kapt_src_jars or srcs.src_jars or ksp_generated_java_src_jars:
-        javac_opts = javac_options_to_flags(ctx.attr.javac_opts[JavacOptions] if ctx.attr.javac_opts else toolchains.kt.javac_options)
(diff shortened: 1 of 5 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -1056,7 +1056,7 @@ def _run_kt_java_builder_actions(
 
         # Kotlin takes care of annotation processing. Note that JavaBuilder "discovers"
         # annotation processors in `deps` also.
-        if len(srcs.kt) > 0:
+        if len(srcs.kt) > 0 and "-proc:none" not in javac_opts:
```
Bencodes: Explicitly checking the `no_proc` attr value here might make more sense than checking lists for arbitrary values.

Bencodes (approved):

kotlin/internal/jvm/compile.bzl
eugenezh: Agree, but if we remove the "len(srcs.kt) > 0" clause, this will change the behavior for existing configurations. If we are fine with this, having explicit option for the behavior is, of course better.

restingbull: Check the attribute directly is sort of annoying:

`not getattr((ctx.attr.javac_opts[JavacOptions] if ctx.attr.javac_opts else toolchains.kt.javac_options), "no_proc", False)`

eugenezh: we can extract the expression from line above:

`javac_options = ctx.attr.javac_opts[JavacOptions] if ctx.attr.javac_opts else toolchains.kt.javac_options`

and then check: 

```
if len(srcs.kt) > 0 and not javac_options.no_proc:
    javac_opts.append("-proc:none")
```

restingbull (approved):

Bencodes (approved):

PR #1708, opened 2026-08-10
Expose additional kotlinc language and diagnostics options

eugenezh:
- Adds per-target options to kt_kotlinc_options that could previously only be controlled toolchain-wide or not at all: api_version, language_version, progressive, x_allow_kotlin_package, x_allow_unstable_dependencies, x_render_internal_diagnostic_names, x_report_all_warnings, x_when_guards, and x_xlanguage.
- For api_version and language_version it is also possible to override the toolchain-wide values on per-target basis.
- Tests added


src/main/starlark/core/options/opts.kotlinc.bzl
```diff
@@ -127,6 +150,27 @@ Options:
         value_to_flag = None,
         map_value_to_flag = _map_jvm_target_to_flag,
     ),
+    "language_version": struct(
```
Bencodes, line 153: Doesn't this create complications around targets supporting multiple language versions since it's coupled to the opts now.

eugenezh: The old behavior "same language version for all" is still possible to achieve if the attribute is defined on the toolchain level. This change adds the flexibility to override it for certain targets with specific requirements. One example is our monorepo at JetBrains, where we have a number of targets that must be compiled differently.

restingbull: language version is always <= to toolchain version? Otherwise, not sure how the compilation would work.

eugenezh: The real ceiling is the compiler release: the compiler is, of course, aware of its own language versions and some versions below. For example, a 2.4 compiler accepts language version values 2.0–2.4. And it does not mean toolchain's language_version is always set to the highest possible one. For example, the project may use a newer compiler with lower language version set for all targets as a default with only some targets having their language versions above the toolchain's one. At the same time some targets may use the lower language versions for  the code compatibility reasons. 

Mixed-version dependency graphs within one build are safe because dependency metadata is read by the compiler itself: a target compiled with language version 2.1 can reference target compiled with language_version 2.4.

restingbull: @Bencodes -- this is equivalent to the java `-source` compatibility flag, not the compiler version.

restingbull (approved):

PR #1713, opened 2026-08-18
[lint] Fix most buildifier lint errors

restingbull:
* Not renaming providers due to inflight pre and backward compatibility
* Skipping sorting dicts (blast radius bigger than expected)


kotlin/compiler/compiler.bzl
```diff
@@ -52,11 +53,14 @@ def _import_labels(labels, rule_kind, **rule_args):
             args["srcjar"] = "@%s//:%s" % (_KT_COMPILER_REPO, sources)
         rule_kind(**args)
 
-def kt_configure_compiler():
+def kt_configure_compiler(name = "kt_configure_compiler"):
```
Bencodes, line 56: We can just disable this with a one-off ignore rather than adding an unused attribute.

kotlin/compiler/ksp.bzl
```diff
@@ -20,11 +21,14 @@ load(
 
 _KSP_COMPILER_PLUGIN_REPO_PREFIX = "@" + _KSP_COMPILER_PLUGIN_REPO + "//:"
 
-def kt_configure_ksp():
+def kt_configure_ksp(name = "kt_configure_ksp"):
```
Bencodes, line 24: Same here, unused seems more confusing than an explicit source level ignore.

Bencodes (approved):

PR #1716, opened 2026-08-21
Compile through the typed Build Tools API

eugenezh:
The worker now invokes the Kotlin compiler through typed Build Tools API, replacing the implementation that parsed CLI flags. The legacy way to invoke compiler remains fully functional and is still the default.

- introduce KotlinBtapiCompiler, the typed contract between the worker and the Build Tools API implementation
- encapsulate all Build-Tools-API-specific worker code in the new tasks/jvm/btapi package (BtapiInvoker, BtapiTaskExecutor)

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/cmd/Build.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/cmd/Build.kt
@@ -18,12 +18,35 @@
 package io.bazel.kotlin.builder.cmd
 
 import io.bazel.kotlin.builder.tasks.CompileKotlin
+import io.bazel.kotlin.builder.tasks.JvmTaskExecutor
 import io.bazel.kotlin.builder.tasks.KotlinBuilder
 import io.bazel.kotlin.builder.tasks.jvm.InternalCompilerPlugins
 import io.bazel.kotlin.builder.tasks.jvm.KotlinJvmTaskExecutor
+import io.bazel.kotlin.builder.tasks.jvm.btapi.BtapiInvoker
+import io.bazel.kotlin.builder.tasks.jvm.btapi.BtapiTaskExecutor
+import io.bazel.kotlin.builder.toolchain.CompilationTaskContext
 import io.bazel.kotlin.builder.toolchain.KotlinToolchain
+import io.bazel.kotlin.model.JvmCompilationTask
 import io.bazel.worker.Worker
 import kotlin.system.exitProcess
+
+private class DispatchingTaskExecutor(
+  private val legacyExecutor: KotlinJvmTaskExecutor,
+  btapiFactory: () -> BtapiTaskExecutor,
+) : JvmTaskExecutor {
+  private val btapiExecutor: BtapiTaskExecutor by lazy(btapiFactory)
+
+  override fun execute(
+    context: CompilationTaskContext,
+    task: JvmCompilationTask,
+  ) {
+    if (context.info.buildToolsApi) {
+      btapiExecutor.execute(context, task)
+    } else {
+      legacyExecutor.execute(context, task)
+    }
+  }
+}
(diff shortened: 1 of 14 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -749,7 +749,8 @@ def _run_kt_builder_action(
         executable = toolchains.kt.kotlinbuilder.files_to_run.executable,
         execution_requirements = _utils.add_dicts(
             toolchains.kt.execution_requirements,
-            {"worker-key-mnemonic": mnemonic},
+            # Compile through Build Tools API with a separate worker pool
+            {"worker-key-mnemonic": mnemonic + "Btapi" if toolchains.kt.experimental_build_tools_api else mnemonic},
```
Bencodes: I have mixed feelings about encoding the BTAPI metadata into the mnemonic since people might be modifying these within their own repos using CLI flags that target these mnemonics. We might unintentionally break or cause regressions here.

eugenezh: I agree, dynamically composing potentially referenceable data is not good. The change is not worth it: it exists only to prevent mixed legacy/BTA compiler invocation within the same worker. This shouldn't happen in real-world usage: either the project will use the legacy way, or it will switch to the bta way. I will revert this change.

Bencodes: cc: @restingbull, you might have some stronger opinions on what's correct here and what's going to work best with RBE performance.

restingbull: It's best to avoid changing mnemonics if at all possible. They are used for a huge number of things that aren't readily apparent (including machine pool allocations, etc.

src/main/kotlin/io/bazel/kotlin/builder/cmd/Build.kt
```diff
@@ -38,8 +61,11 @@ object Build {
             toolchain.kapt3Plugin,
             toolchain.jdepsGen,
           )
-        val compilerBuilder = KotlinToolchain.KotlincInvokerBuilder(toolchain)
-        val jvmTaskExecutor = KotlinJvmTaskExecutor(compilerBuilder, plugins)
+        val jvmTaskExecutor =
+          DispatchingTaskExecutor(
+            legacyExecutor = KotlinJvmTaskExecutor(toolchain, plugins),
+            btapiFactory = { BtapiTaskExecutor(BtapiInvoker(toolchain), plugins) },
```
Bencodes, line 67: These warnings are new as part of this PR:
```
(16:40:27) INFO: From CliKotlinc src/main/kotlin/io/bazel/kotlin/builder/cmd/build_lib.jar [for tool]:
src/main/kotlin/io/bazel/kotlin/builder/cmd/Build.kt:67:48: warning: cannot access 'io.bazel.kotlin.compiler.KotlinBtapiCompiler' which is a supertype of 'BtapiInvoker'. This may be forbidden soon. Check the module classpath for missing or conflicting dependencies.
            btapiFactory = { BtapiTaskExecutor(BtapiInvoker(toolchain), plugins) },
                                               ^^^^^^^^^^^^
```

https://buildkite.com/bazel/rules-kotlin-kotlin/builds/6615/list?jid=01a02530-6903-4e72-be1f-6019fe292a6d&tab=output#01a02530-6903-4e72-be1f-6019fe292a6d/L1213

Bencodes (approved):

src/main/kotlin/io/bazel/kotlin/builder/cmd/Build.kt
eugenezh: Thanks for pointing on this. This is a missing transitive dependency that should actually be a direct dependency. I'll add it.

PR #1718, opened 2026-08-21
Pass compiler plugins as a single typed JSON payload

eugenezh:
Replace the four per-phase string-list worker flags (stubs/compiler plugin classpath + options) with one self-describing --plugins_payload JSON argument: the typed plugin transport the Build Tools API compilation uses.

- introduce new Plugin proto messages (id, classpath, key/value options, phases) the payload is parsed into; the worker expands them into the existing per-phase task fields, so the -Xplugin/-P arguments handed to kotlinc are unchanged for both the legacy and the Build Tools API invocation
- plugin options still arrive in their legacy packed "key=value" form and are split at the first '='; cleaner key/value handling of KtCompilerPluginOption is done later in a separate commit
- the payload contract is tested on both ends against the same example (Starlark producer, Kotlin parser)

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -53,6 +53,10 @@ load(
     "//kotlin/internal/utils:utils.bzl",
     _utils = "utils",
 )
+load(
+    "//src/main/starlark/core/plugin:payload.bzl",
+    _plugin_payload = "plugin_payload",
+)
 
 # UTILITY ##############################################################################################################
 def find_java_toolchain(ctx, target):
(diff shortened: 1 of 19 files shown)


kotlin/internal/jvm/compile.bzl
```diff
@@ -689,31 +685,7 @@ def _run_kt_builder_action(
...
-        omit_if_empty = True,
-    )
-
-    args.add_all(
-        "--compiler_plugin_options",
-        plugins.compile_phase.options,
-        map_each = _format_compile_plugin_options,
-        omit_if_empty = True,
-    )
+    args.add("--plugins_payload", _plugin_payload.plugins_payload_json(plugins.plugins))
```
Bencodes: This _might_ actually break path stripping since it's manually encoding the paths into json blob. Might be worth duble checking that.

eugenezh: Yes, it will actually break path stripping. Thanks for pointing on this one. I'll update the change so that the payload_json computation is deferred (with args.add_all and map_each) until optional PathMapper is installed => the computed json will contain correctly stripped paths.

Bencodes (approved):

PR #1724, opened 2026-08-25
Toolchain btapi runtime

eugenezh:
Support different Kotlin compiler implementations/versions within the build
worker when compiling through the Build Tools API. The compiler and the
Build Tools API implementation are supplied via the toolchain.

- The toolchain gains classpath attributes for the Build Tools API runtime
  (the implementation layer, the compiler, and the runtime libraries) and
  for the internal compiler plugins matching the selected compiler dialect.
- When the Build Tools API toggle is on, a set of corresponding worker
  flags is emitted; legacy actions are unchanged.
- The worker records the runtime on the task model, and its presence is
  the signal that toggles the compilation through the Build Tools API;
  different compiler runtimes coexist in one worker process.
- The embeddable compiler family is selectable without leaving the
  repository, including embeddable-dialect variants of the in-repo
  compiler plugins.
- Analysis tests pin the runtime flags in both toggle positions; builder
  unit tests run the Build Tools API compilation end to end on the
  embeddable runtime.


src/main/starlark/core/repositories/versions.bzl
```diff
@@ -95,6 +95,32 @@ versions = struct(
         ],
         sha256 = "3953d283e7710c990672e403a87df393d9726a9bf3e172194ebb5c33e062fcb0",
     ),
+    # The embeddable compiler family: the Maven-published kotlinc build whose bundled third-party
+    # packages are shaded (e.g. org.jetbrains.kotlin.com.intellij). Compiler plugins published for
+    # Gradle/Maven consumption are compiled against this dialect. These jars are an alternative
+    # runtime for the Build Tools API compilation path, selectable through the btapi_*/internal_*
+    # toolchain attributes; the bundled CLI distribution remains the default.
+    KOTLIN_COMPILER_EMBEDDABLE = version(
+        version = "2.4.10",
```
Bencodes: Do these actually need to track the user-supplied Kotlin version?

eugenezh: correct. I have reworked these definitions: now the kotlinc_embeddable artifacts are user-overridable with the help of kotlinc_embeddable_version structure passed in compiler_embeddable_release arg, similar to how the 'compiler_release' argument works. 

versions.bzl contain the definition for the current release version

src/main/kotlin/shade.embeddable.jarjar
restingbull, line 1: Need to flip the plugin to shade non-embedded plugin: https://github.com/bazel-contrib/rules_kotlin/blob/master/kotlin/internal/jvm/impl.bzl#L465

src/main/kotlin/io/bazel/kotlin/builder/toolchain/KotlinToolchain.kt
```diff
@@ -143,6 +162,8 @@ class KotlinToolchain private constructor(
     @JvmStatic
     fun createToolchain(
       kotlinc: File,
+      kotlinStdlib: File,
```
restingbull: Is this the stdlib for compilation or the compiler?

eugenezh: this one is for the compiler itself, needed to make the btapi classloader complete. I can rename it to "compilerStdlib" to reduce the naming confusion

restingbull: Let's do that -- ideally, we'd stop shipping stdlib via the toolchain, but that is taking a minute.

src/main/kotlin/shade.embeddable.jarjar
eugenezh: I'll prepare the change. Should I amend this PR with it, or do a separate commit within this PR or have it as a separate PR?

restingbull (approved):

PR #1729, opened 2026-08-28
Add Kover code coverage support

jbarr21:
## Summary

This PR adds **Kover** (JetBrains' Kotlin code coverage tool) as an alternative to JaCoCo for code coverage in `rules_kotlin`.

### Why Kover?

| Feature | JaCoCo | Kover |
|---------|--------|-------|
| Instrumentation | Build-time bytecode modification | Runtime JVM agent |
| Kotlin support | Limited (misses inline functions) | Native Kotlin awareness |
| Recompilation needed | Yes | No |
| Mixed Java/Kotlin | Problematic | Works well |

### New Toolchain Attributes

```python
define_kt_toolchain(
    name = "kotlin_toolchain",
    experimental_kover_enabled = True,
    experimental_kover_agent = "@maven//:org_jetbrains_kotlinx_kover_jvm_agent",
    experimental_kover_exclude = ["com.example.generated.*"],
    experimental_kover_exclude_annotation = ["Generated"],
    experimental_kover_exclude_inherited_from = ["android.app.Activity"],
)
```

### Usage

```bash
bazel coverage //your/kotlin/test_target
```

Output files are created alongside test outputs:
- `*-kover_report.ic` - Binary coverage data
- `*-kover_metadata.txt` - Metadata for Kover CLI report generation

Generate reports with:
```bash
java -jar kover-cli.jar report @path/to/kover_metadata.txt --html report/
```

### Changes

- **kotlin/internal/jvm/kover.bzl** (NEW) - Core Kover integration logic
- **kotlin/internal/jvm/compile.bzl** - Disable JaCoCo instrumentation when Kover enabled
- **kotlin/internal/jvm/impl.bzl** - Kover support in `kt_jvm_test`
- **kotlin/internal/jvm/kt_android_local_test_impl.bzl** - Kover support for Android local tests
- **kotlin/internal/toolchains.bzl** - New toolchain configuration attributes

### Notes

- Kover agent is added to both `-Xbootclasspath/a:` and `-javaagent:` to support code using the bootstrap classloader
- Report generation is done outside Bazel (test actions are terminal)
- For mixed Java/Kotlin sourcesets, JaCoCo instrumentation must be disabled

## Test plan

- [x] Unit tests for `get_kover_jvm_flags` function
- [x] Toolchain builds successfully
- [ ] Manual testing with actual Kotlin test targets

## Usage

To enable Kover in your project:

      define_kt_toolchain(
          name = "kotlin_toolchain",
          experimental_kover_enabled = True,
          experimental_kover_agent = "@maven//:org_jetbrains_kotlinx_kover_jvm_agent",
          experimental_kover_exclude = ["com.example.generated.*"],
          experimental_kover_exclude_annotation = ["Generated"],
      )

      Then run: bazel coverage //your/kotlin/test_target

## Android Code Duplication
kt_android_local_test_impl.bzl duplicates significant logic from rules_android. This needs to be kept in sync manually. There's ongoing work with Google to add better extensibility points to rules_android.


kotlin/internal/jvm/kt_android_local_test_impl.bzl
```diff
@@ -189,13 +224,167 @@ def _process_jvm(ctx, resources_ctx, **_unused_sub_ctxs):
...
             additional_jvm_flags = jvm_flags,
         ),
-        runfiles = ctx.runfiles(files = runfiles),
+        runfiles = ctx.runfiles(
+            files = runfiles,
+            transitive_files = depset(transitive = transitive),
+        ),
+    )
+
+# TODO: follow up with Google to have rules_android provide better extensibility points
```
restingbull, line 233: File an issue on the TODO.

kotlin/internal/jvm/kt_android_local_test_impl.bzl
```diff
@@ -189,13 +224,167 @@ def _process_jvm(ctx, resources_ctx, **_unused_sub_ctxs):
...
+
+    # Use a heuristic based on the rule name and the "srcs" list
+    # to determine the primary Java class.
+    expected = "/" + ctx.label.name + ".java"
+    for f in ctx.attr.srcs:
+        path = f.label.package + "/" + f.label.name
+        if path.endswith(expected):
+            return _java.resolve_package(path[:-5])
+
+    # Last resort: Use the name and package name of the target.
```
restingbull: Let's error out here -- this is pretty long shot in many code bases.

Getting a class not found for the derived name is... not great.

kotlin/internal/toolchains.bzl
```diff
@@ -145,6 +150,23 @@ _kt_toolchain = rule(
...
         ),
+        "experimental_kover_agent": attr.label(
+            doc = """Kover agent Jar target used for code coverage, only used if experimental_kover_enabled is true.""",
+            allow_single_file = True,
+        ),
+        "experimental_kover_enabled": attr.bool(
+            doc = """Use kover for code coverage.""",
+            default = False,
+        ),
+        "experimental_kover_exclude": attr.string_list(
```
restingbull, line 161: Why would these live on the toolchain? Exclusions seem like a target, flag, or separate configuration rule.

src/main/kotlin/io/bazel/kotlin/test/BazelIntegrationTestRunner.kt
```diff
@@ -149,8 +149,20 @@ object BazelIntegrationTestRunner {
...
+              } else {
+                bazel.run(
+                  workspace,
+                  *systemFlags,
+                  "coverage",
+                  *commandFlags,
+                  *coverageTargets,
+                ).onFailThrow()
+              }
+              if (isWindows) {
```
restingbull, line 164: this clause can't happen.

kotlin/internal/toolchains.bzl
```diff
@@ -145,6 +150,23 @@ _kt_toolchain = rule(
             doc = "Enables experimental support for Build Tools API integration",
             default = False,
         ),
+        "experimental_kover_agent": attr.label(
```
restingbull, line 153: Should this be private? (e.g. _experimental_kover_agent with a default pointing at the default agent target.)

kotlin/internal/toolchains.bzl
```diff
@@ -145,6 +150,23 @@ _kt_toolchain = rule(
...
+            default = False,
+        ),
+        "experimental_kover_exclude": attr.string_list(
+            doc = """List of exclusions to use when generating kover reports.""",
+        ),
+        "experimental_kover_exclude_annotation": attr.string_list(
+            doc = """List of annotation exclusions to use when generating kover reports.""",
+        ),
+        "experimental_kover_exclude_inherited_from": attr.string_list(
+            doc = """List of annotation exclusions from inherited classes to use when generating kover reports.""",
```
restingbull, line 168: I'm having little trouble figuring out the use case for inherited annotation exclusions. I know it's a Kover feature, but it's a weird one. 

Do ya'll use it?

src/main/kotlin/io/bazel/kotlin/test/BazelIntegrationTestRunner.kt
jbarr21: how so? i didn't write this, but i see 2 if/else blocks after another. i can merge the 2 if/else blocks if you prefer

kotlin/internal/toolchains.bzl
jbarr21: we use it for such things as "kotlin.Enum", "java.lang.Enum", & "android.view.View". we do snapshot tests for UI instead of unit testing with lots mocks

kotlin/internal/toolchains.bzl
jbarr21: it looks like it was designed to be public. the way this was wired up for us is that the consuming project passes is in the resolved agent jar target to be used

kotlin/internal/toolchains.bzl
jbarr21: i don't have the context for why it was done this way, but i could see it being doing this way for more flexability. unlike lint & detekt that have their own defined config files, we are just passing additional options here

kotlin/internal/jvm/kt_android_local_test_impl.bzl
jbarr21: I don't have specific context on the ask, but i created the issue with a ref back to the PR

https://github.com/bazelbuild/rules_android/issues/558

kotlin/internal/jvm/kt_android_local_test_impl.bzl
jbarr21: 17b4bd6

kotlin/internal/jvm/kt_android_local_test_impl.bzl
restingbull: Usual form for TODO is `TODO[<issue url>]: `

src/main/kotlin/io/bazel/kotlin/test/BazelIntegrationTestRunner.kt
restingbull: line 153 checks if isWindows. This is the else clause of that check, which makes it dead code.

kotlin/internal/toolchains.bzl
restingbull: It's an extension of the public API, which is already pretty weird. Given that a large repository could have multiple unrelated classes to be excluded (different projects with specific dependencies), this ends up being kind of awkward to maintain while living on the toolchain.

kotlin/internal/toolchains.bzl
restingbull: In that case, let's drop the boolean and enable on the jar being present. Right now, there are two toggles for the same thing -- increases the space to make mistakes on configuration.

examples/kover/BUILD.bazel
```diff
@@ -0,0 +1,9 @@
+load("@rules_kotlin//kotlin:core.bzl", "define_kt_toolchain")
+
+package(default_visibility = ["//visibility:public"])
+
+define_kt_toolchain(
```
Bencodes, line 5: Please add `examples/kover` to the `--deleted_packages` lists in `.bazelrc`. Otherwise, root `//...` builds include these packages.

```
❯ bazel build //examples/kover:all //examples/kover/app:all
ERROR: no such package '@@[unknown repo 'maven' requested from @@]//': The repository '@@[unknown repo 'maven' requested from @@]' could not be resolved: No repository visible as '@maven' from main repository
ERROR: /Users/blee/.codex/worktrees/6a17/rules_kotlin/examples/kover/BUILD.bazel:5:20: no such package '@@[unknown repo 'maven' requested from @@]//': The repository '@@[unknown repo 'maven' requested from @@]' could not be resolved: No repository visible as '@maven' from main repository and referenced by '//examples/kover:kotlin_toolchain_impl'
ERROR: Analysis of target '//examples/kover:kotlin_toolchain_impl' failed; build aborted: Analysis failed
```

kotlin/internal/jvm/kt_android_local_test_impl.bzl
```diff
@@ -189,13 +224,164 @@ def _process_jvm(ctx, resources_ctx, **_unused_sub_ctxs):
...
+    ctx.actions.write(
+        output = classpath_file,
+        content = args,
+    )
+    return stub_file
+
+def _get_classpath(s):
+    return "${J3}" + s.short_path
+
+def _get_jvm_flags(ctx, main_class, robolectric_properties_path, additional_jvm_flags):
```
Bencodes, line 361: Largely copied from https://github.com/bazelbuild/rules_android/blob/26f90cfc233406b2804c3d27ef7cc652fccbcc73/rules/android_local_test/impl.bzl#L487-L508, need to make sure we keep these in sync somehow.

In an ideal world we'd just call into their rules instead of outright copying.

Bencodes: Did you mean to drop the JNI support here? https://github.com/bazelbuild/rules_android/blob/main/rules/android_local_test/impl.bzl#L488-L491

kotlin/internal/jvm/kover.bzl
```diff
@@ -0,0 +1,207 @@
...
+        A tuple of (kover_output_file, kover_args_file).
+    """
+
+    # declare code coverage raw data binary output file
+    binary_output_name = "%s-kover_report.ic" % name
+    kover_output_file = ctx.actions.declare_file(binary_output_name)
+
+    # Declare an initial output without relying on a shell command so this also
+    # works on Windows. The Kover agent replaces it when the test runs.
+    ctx.actions.write(kover_output_file, "")
```
Bencodes, line 127: This feels like a bit of a hack

Bencodes, line 127: This feels like a bit of a hack. The coverage outputs should probably also get written into the COVERAGE_DIR as well. If we really wanted to align with Bazel then we'd also convert the Kover format into lcov while doing that.

PR #1731, opened 2026-08-31
Compile the java half of a target against the full Kotlin output

eugenezh:
A file-private Kotlin class compiles to package-private bytecode, which may be referenced from java code in the same target. These file-private Kotlin classes are stripped from ABI jar, if "removePrivateClasses" is set. Compiling java half of the target against the ABI jar leads to compilation errors.

### DIFF
--- a/kotlin/internal/jvm/compile.bzl
+++ b/kotlin/internal/jvm/compile.bzl
@@ -1028,7 +1028,8 @@ def _run_kt_java_builder_actions(
         compile_jars.append(kt_compile_jar)
         output_jars.append(kt_runtime_jar)
         if not annotation_processors or not srcs.kt:
-            kt_stubs_for_java.append(JavaInfo(compile_jar = kt_compile_jar, output_jar = kt_runtime_jar, neverlink = True))
+            # Compile the java half of this target against the FULL Kotlin output, not the ABI;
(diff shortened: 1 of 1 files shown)


restingbull (approved):

kotlin/internal/jvm/compile.bzl
```diff
@@ -1028,7 +1028,8 @@ def _run_kt_java_builder_actions(
         compile_jars.append(kt_compile_jar)
         output_jars.append(kt_runtime_jar)
         if not annotation_processors or not srcs.kt:
-            kt_stubs_for_java.append(JavaInfo(compile_jar = kt_compile_jar, output_jar = kt_runtime_jar, neverlink = True))
+            # Compile the java half of this target against the FULL Kotlin output, not the ABI;
+            kt_stubs_for_java.append(JavaInfo(compile_jar = kt_runtime_jar, output_jar = kt_runtime_jar, neverlink = True))
```
Bencodes, line 1032: Worth noting that this could increase build times for folks that have mixed-source codebases. Primarily when a change that was previously isolated and not included in the compile time jar, will now cause the Java action to get invalidated and rebuild.

restingbull: It seems like a reasonable tradeoff for correctness. Before, private classes would be dropped from the compilations step -- now they are read.

The simple fix is to split the sources (without private classes) into separate targets. (Might be worth writing an annotation processor catch that and print the suggestion. Not sure.)

Should we call it out in the release notes?

PR #1736, opened 2026-09-12
Invalidate BTAPI compiler caches when runtime jars change

Bencodes:
BTAPI cached compiler invokers by jar paths, so replacing a custom runtime jar at the same path kept using the old compiler. Share KSP2's SHA-256 fingerprinting of jar paths, sizes, and contents, and replace BTAPI cache entries when that fingerprint changes.

BTAPI retains classpath order, while KSP2 keeps its existing order-insensitive cache behavior. Hashing streams jar contents instead of allocating a buffer for each entire jar.

Validation:
- Confirmed the new regression test fails with the old cache behavior.
- `bazel test //src/...`: all 166 tests passed, including runtime replacement with unchanged file size and timestamp, classpath ordering, and BTAPI compilation tests.
- `bazel run //docs:write_docs` and `bazel run //:buildifier.fix`: passed with no generated changes.

Follow-up to #1724.

### DIFF
--- a/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/Ksp2Task.kt
+++ b/src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/Ksp2Task.kt
@@ -20,6 +20,7 @@ package io.bazel.kotlin.builder.tasks.jvm
 import io.bazel.kotlin.builder.utils.ArgMap
 import io.bazel.kotlin.builder.utils.ArgMaps
 import io.bazel.kotlin.builder.utils.Flag
+import io.bazel.kotlin.builder.utils.fingerprintOf
 import io.bazel.worker.Status
 import io.bazel.worker.Work
 import io.bazel.worker.WorkerContext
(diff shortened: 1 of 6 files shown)


restingbull (approved):

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/Ksp2Task.kt
```diff
@@ -91,7 +90,7 @@ class Ksp2Task : Work {
 
     fun getKspClassLoader(processorClasspath: List<String>): URLClassLoader {
       val cacheKey = processorClasspath.sorted().joinToString(File.pathSeparator)
-      val fingerprint = fingerprintOf(processorClasspath)
+      val fingerprint = fingerprintOf(processorClasspath.sorted())
```
eugenezh: In any classpath the order matters. Should we stop sorting to make the fingerprint order-sensitive?

src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/btapi/BtapiTaskExecutor.kt
```diff
@@ -99,7 +105,7 @@ class BtapiTaskExecutor(
     val btapiRuntime =
       task.info.toolchainInfo.btapi
         .toRuntime()
-    val compiler = invokers.computeIfAbsent(btapiRuntime.apiImplClasspath, ::loadCompilerInvoker)
+    val compiler = getCompilerInvoker(btapiRuntime.apiImplClasspath)
```
eugenezh: maybe lazy-calculate the classpath fingerprint once per-btapiRuntime and store it in the Runtime structure to avoid the calculation on every invocation?

PR #1737, opened 2026-09-12
Support pluggable Android local test coverage pipelines

Bencodes:
Separate `kt_android_local_test` coverage runtime selection from Kotlin compilation, using the `CoverageProcessor` introduced in bazelbuild/rules_android#562. A downstream rule can replace that processor and reuse Kotlin resource handling, compilation, deploy-jar creation, launcher generation, and finalization.

The default processor preserves the Kotlin toolchain's JaCoCo runner. The JVM processor consumes its dependencies, entry classes, and JVM flags while retaining the Java security-manager flag. Expose `make_rule` with merged Kotlin attributes and toolchains so custom implementations do not need to copy the rule declaration.

The analysis-test example replaces only `CoverageProcessor` with an inert agent configuration. It covers default/custom pipelines with coverage enabled and disabled, including Kotlin compilation, launcher substitutions, runfiles, custom attributes, and user JVM flags. It introduces no Kover dependency and does not change bytecode instrumentation.

Temporarily pin rules_android to `9e7e6f868e663bbbd3ff07bb2781e12d0d4d3cbf` in the root module and Android-consuming examples. Overrides are root-only, so example modules need their own pins. Set the tool Java source level to 17 alongside these pins, as required by the newer Android tools. This is independent of rules_kotlin#1729; merging/releasing requires the upstream API, and the archive pins should become a released dependency once available.

Validation (macOS, Bazel 9.2.0):
- `bazel test //src/test/starlark/android_local_test:coverage_tests`: all four analysis tests pass.
- `bazel build //src/test/starlark/android_local_test:coverage_tests_default_subject`: passes, including Kotlin compilation and Android resource processing.
- Buildifier formatting/lint and `git diff --check` pass.

The agent is deliberately an analysis-only stand-in; no custom coverage engine is executed. The full examples/platform matrix remains for CI.

### DIFF
--- a/kotlin/internal/jvm/kt_android_local_test.bzl
+++ b/kotlin/internal/jvm/kt_android_local_test.bzl
@@ -55,11 +55,24 @@ _ATTRS = _utils.add_dicts(_BASE_ATTRS, _attrs.runnable_common_attr, {
     ),
 })
 
-kt_android_local_test = _make_rule(
-    implementation = _kt_android_local_test_impl,
-    attrs = _ATTRS,
-    additional_toolchains = [
-        _TOOLCHAIN_TYPE,
-        _JAVA_RUNTIME_TOOLCHAIN_TYPE,
-    ],
-)
+def make_rule(attrs = {}, implementation = _kt_android_local_test_impl, additional_toolchains = []):
+    """Creates a Kotlin Android local test rule with a custom processing pipeline.
+
+    Args:
+      attrs: Additional attributes or overrides for the Kotlin test attributes.
+      implementation: Rule implementation running the processing pipeline.
+      additional_toolchains: Toolchains required by custom processors.
+
+    Returns:
+      A test rule with the Kotlin compilation attributes and toolchains.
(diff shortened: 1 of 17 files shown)


kotlin/internal/jvm/kt_android_local_test.bzl
```diff
@@ -55,11 +55,24 @@ _ATTRS = _utils.add_dicts(_BASE_ATTRS, _attrs.runnable_common_attr, {
...
 
-kt_android_local_test = _make_rule(
-    implementation = _kt_android_local_test_impl,
-    attrs = _ATTRS,
-    additional_toolchains = [
-        _TOOLCHAIN_TYPE,
-        _JAVA_RUNTIME_TOOLCHAIN_TYPE,
-    ],
-)
+def make_rule(attrs = {}, implementation = _kt_android_local_test_impl, additional_toolchains = []):
```
restingbull, line 58: What's the benefit of the macro here?

PR #1743, opened 2026-09-14
Prune the legacy compiler classloader to the jars the legacy invocati…

eugenezh:
…on loads

- The Build Tools API invocation now assembles its own classloaders from the toolchain-supplied runtime spec, so none of them is loaded here anymore.


restingbull (approved):
Love this

PR #1753, opened 2026-09-16
Remove discarded dependency protobuf serialization

Bencodes:
Remove a protobuf serialization whose bytes were discarded immediately before the actual output serialization. Preserve the existing writer and dependency ordering. Add regression coverage for large Unicode dependency lists, duplicates, and stronger dependency kinds.

Validation on the original stack: all seven merger cases passed; all 168 source test targets and Android compilation passed.


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/JdepsMerger.kt
```diff
@@ -85,7 +85,6 @@ class JdepsMerger {
       rootBuilder.addAllDependency(dependencyMap.values)
 
       rootBuilder.success = true
-      rootBuilder.build().toByteArray()
```
restingbull, line 88: Wow. Some much for reviews.

restingbull (approved):

PR #1758, opened 2026-09-18
Emit KSP2 output jar entries in sorted order

knutwannheden:
`packageDirectoriesToJar` writes entries in `Files.walk` order. That is filesystem
iteration order, which is not guaranteed to be stable between runs, so two runs over
identical inputs can still produce byte-different jars.

#1657 fixed the other half of this by stamping every entry with a fixed epoch. The
ordering half is still open.

## Mechanism

`Files.walk` returns entries in whatever order the filesystem hands them back. That
order is stable often enough to be easy to miss -- it tends to reflect creation order
on a freshly written directory -- but it is not a guarantee, and it changes with
directory size, with the underlying filesystem, and after entries are removed and
recreated. KSP output directories are written fresh on every run, by processors that
make no promise about the order in which they emit files.

## Why it matters beyond the jar

A jar's bytes are its action's output digest. When that digest changes even though
nothing it depends on did, every consumer of the generated code is invalidated: the
compilation that reads it, and transitively everything built on top of that.

The result is a rebuild that should be a no-op instead re-executing a large part of the
graph. Neither the local action cache nor a remote cache can help, because as far as
the build is concerned the inputs really are different each time -- so this is felt
most sharply by exactly the setups that were supposed to benefit most from caching.
It also means these jars cannot be reproduced byte-for-byte, which matters for anyone
verifying build outputs.

## The fix

Collect the walk and sort it before writing. Natural `Path` order also places a parent
directory ahead of its children, which is the order entries were already being written
in, so nothing else in the method has to change.

`packageDirectoriesToJar` loses its `private` modifier so the tests can drive it
directly. The class has a no-arg constructor whose only work is initializing a cache,
so constructing one in a test is free.

## Tests

Three, covering the property rather than the implementation:

- two packagings of the same directory, separated by more than a wall-clock second,
  are byte-identical
- every entry carries the fixed epoch
- entries come out sorted

The last one fails if the sort is removed -- the walk order for that fixture is not
already sorted, so the test is actually load-bearing rather than incidentally green.


src/main/kotlin/io/bazel/kotlin/builder/tasks/jvm/Ksp2Task.kt
```diff
@@ -319,40 +326,44 @@ class Ksp2Task : Work {
...
-                if (parentPath !in addedEntries) {
-                  addedEntries.add(parentPath)
-                  jar.putNextEntry(jarEntry(parentPath))
-                  jar.closeEntry()
-                }
-              }
+        // Natural Path order also places a parent directory ahead of its children.
+        val paths =
+          Files.walk(dir).use { stream ->
+            stream.sorted().collect(Collectors.toList())
```
eugenezh, line 332: sorting Path objects may be OS-dependent: for example, Windows sorts case-insensitively, while Linux honors case, so "a" and "B" entries will have a different order on these systems. Need some platform-independent deterministic ordering

